Issue #33, opened 2026-07-22
[Hotfix][Critical] Rotate leaked secrets & move out of source control - TASK-001


MohKardan:
## Summary
Two live credentials are committed to this **public** repository and are present in git history. They must be rotated immediately and moved out of source control.

- **Task**: TASK-001 (`docs/process/SPRINT_PLAN.md`)
- **Audit findings**: C-1 (hardcoded API key), C-2 (Telegram bot token in committed config)
- **Priority**: Critical — highest priority in Sprint 1, blocks safe deployment
- **Estimate**: 1–2 days

## Why this is urgent
The repository is **public**. Both values below are readable by anyone and are indexed by automated secret scanners within minutes of being pushed. Assume they are **already compromised**.

Rotation is what actually fixes this. Scrubbing git history is secondary — it does not un-leak a value that has already been public.

## What is exposed

| Secret | Location | Impact |
|---|---|---|
| Telegram bot token | `config/appsettings.global.json` → `Services/TallaEgg.TelegramBot.Infrastructure/TelegramBotToken` | Full takeover of the bot: read messages, send messages to users as us |
| API key | `src/TallaEgg/TallaEgg.Core/APIKeyConstant.cs` → `TallaEggApiKey` | This is the **only** authentication mechanism across services — effectively an open API |

**Not affected**: the `ConnectionStrings` in the same file are all `Server=localhost;…;Trusted_Connection` — no passwords, no remote hosts. They do not need rotation (but should still leave source control).

## Root cause of the ongoing leak
`.gitignore` ignores `appsettings.*.json` (line 382) but then explicitly re-includes the file:

```
384: !config/appsettings.global.json
692: !config/appsettings.global.json
```

So the config is force-tracked. Until those negations are removed, any secret placed in that file leaks again.

Separately, `TallaEggApiKey` is a `const string`, so it is compiled into every consuming service — it cannot be rotated by config alone and requires a rebuild + redeploy of all services.

## Steps (in this order — order matters)

1. **Revoke the Telegram token first.** BotFather → `/revoke` → generate a new token. This instantly invalidates the leaked value and is the single highest-value action.
2. **Generate a new API key.** Move it from `const string` to configuration (`IOptions<T>` / environment variable), per the audit's recommendation.
3. **Stop the leak**: remove the `!config/appsettings.global.json` negations from `.gitignore`, `git rm --cached config/appsettings.global.json`, and commit a `config/appsettings.template.json` with placeholders instead.
[...]

Issue #90, opened 2026-08-07
Automated quote publishing: multi-source gold price feed with admin-configurable spread


MohKardan:
## Summary

An investor conversation surfaced two product suggestions: an automated quote (so the shop isn't
dependent on an admin manually typing `buyPrice-sellPrice` every time the market moves) and a
profit/loss view in the accounting section. This issue covers the first. The second is
deliberately **not** filed separately — it will be folded into #61 (margin position monitoring)
when that work starts, since both need the same per-user cost-basis calculation.

## Current manual flow (unchanged by this work)

An admin sends a message matching `^(\d+)-(\d+)$` (buy-sell), which calls `Quote.Publish(...)` →
`POST /api/quotes`. This publishes a new active quote for `MAUA/IRT` and deactivates the previous
one. See `Orders.Core/Quote.cs` and `BotHandlerAdmin.cs`.

This issue adds an **automatic caller** of that same path — the publish mechanism itself does not
change, and the manual command keeps working exactly as it does today (it always overrides,
since only one quote per symbol is ever active).

### 1. Price source: an ordered chain, not one provider

An `IGoldPriceProvider` interface (single method, returns the current 18k gold price or null/throws
on failure), with two initial implementations:

- `NerkhGoldPriceProvider` — nerkh.io. Free tier: 1 req/60s, **no API key required**.
- `BrsApiGoldPriceProvider` — brsapi.ir. Free tier: 1500 req/day, requires a free API key.

A small orchestrator tries providers in a configured order and uses the first one that succeeds,
logging which provider answered and which were skipped. Adding a third or fourth source later is
just a new class implementing the interface plus one line in DI registration — no changes to the
orchestrator or the background service that calls it.

### 2. Publisher: a new BackgroundService in Orders.Api

Same pattern already established by `MatchingEngineService` (also a `BackgroundService` in
`Orders.Api`). On each tick: ask the provider chain for a price, apply the configured spread,
call `Quote.Publish(...)`.

### 3. Spread: admin-configurable at runtime, not a config file value

Config-file values only get read at startup, so a value the admin needs to change without a
redeploy can't live in `appsettings.global.json`. This needs a small new table in the Orders
database — one migration:

```
AutoQuoteSettings: SpreadPercent, IsEnabled, UpdatedByUserId, UpdatedAt
```

A new admin bot command sets it — proposed shape `اسپرد 0.5`, open to a better name. `IsEnabled`
[...]

PR #170, opened 2026-08-30
[Chore] Add run-tallaegg agent skill

MohKardan:
## Description
Adds `.claude/skills/run-tallaegg/`: an agent-facing skill that documents and scripts how to build, launch, and actually drive the TallaEgg backend end to end — not just run its unit tests — without needing a live Telegram bot token or network access to `api.telegram.org`.

## Changes Made
- `.claude/skills/run-tallaegg/driver.ps1` — a PowerShell driver with four subcommands (`start`, `status`, `smoke`, `stop`) that builds the solution in Release, launches Users/Wallet/Orders as background processes, polls their Swagger pages for health, and stops them by resolving current port ownership plus any survivor from `launcher.pids`.
- `.claude/skills/run-tallaegg/SKILL.md` — points future agents at the driver, documents Prerequisites/Setup/Build/Run/Test, and records Gotchas and Troubleshooting hit while building this.
- `.gitignore`: added `run-logs/` for the driver's redirected stdout/stderr (the individual `.log` files were already covered by the existing `*.log` rule; the directory itself and `launcher.pids` were not).

## Why this shape
The repo already ships `TelegramBot/TallaEgg.TelegramBot.Simulator`, a project that replays real bot conversations (registration, admin approval, quote publishing, quote-fill trades, menu navigation) through the real `IBotHandler` and the real Users/Wallet/Orders APIs and database — faking only the Telegram transport. That's the interaction surface this skill points at via `driver.ps1 smoke`, rather than writing a new harness from scratch.

## What a `smoke` run changes on the database
- **Rows with `TelegramId < 0`** — every simulated user. Each run's `DataReset` phase wipes only that range before starting, so runs are self-cleaning and a real (positive-id) user's data is never touched.
- **The auto-quote flag for `MAUA/IRT`** — the Simulator turns it off in Phase 2 (a background publisher replacing the run's quotes breaks quote-fill trades) and never turns it back on. This is per-symbol Orders-DB state, not `TelegramId`-scoped, so `DataReset` does not restore it. `driver.ps1 smoke` reads the flag before the run and puts it back afterwards, including when the run fails part-way.

> An earlier revision of this PR claimed a run "only ever touches rows with `TelegramId < 0`" and so "can never touch a real user's data". The second half is true; the first half was not, and the auto-quote flag was the counterexample. Fixed in review — see below.

## Review fixes (2026-08-30)
Review found the driver reporting success in several cases where it had not done the work:
- **A failed build was never detected.** In Windows PowerShell 5.1 a native command's nonzero exit does not throw, even under `$ErrorActionPreference = 'Stop'`, so `start` fell through to `dotnet run --no-build`, served the *previous* build and printed "All services up" — the exact outcome this skill exists to prevent. Now checks `$LASTEXITCODE` and throws.
- **`smoke` could not fail.** The Simulator returns 0 whatever happens and only logs its error count; the driver now parses the summary line and throws on a nonzero count or a missing line.
- **Partial argument overrides escalated the run ~100x.** `smoke --seed 7` passed only `--seed`, so `SimulationOptions` fell back to its own compiled defaults (100 users / 120 quotes / 1000 trades). Arguments are now merged over the documented defaults; unknown ones are rejected, and `--users` below 2 is refused (user #0 is the admin/market maker and cannot be its own counterparty, so a 1-user run always ends `errors 1`).
- **`start` now refuses when the ports are already listening**, instead of launching processes that die on "address already in use" while the health poll answers 200 from the old ones.
- **Timeout diagnostics tailed the wrong stream.** All three APIs use Serilog `.WriteTo.Console()` with no `standardErrorFromLevel`, so errors land in `*.out.log` and the `*.err.log` files the driver printed were empty. Both are tailed now.
- **`stop` now also kills `launcher.pids` survivors**, so a service that died before Kestrel bound (unreachable SQL Server) is cleaned up, and counts only kills that actually succeeded.
- **The Setup recipe now uses `Copy-Item -NoClobber`.** `config/appsettings.global.json` is untracked (#33), so an agent following the doc verbatim on a configured box would have destroyed live credentials with no way to recover them.
- **The script's own usage block said `pwsh`**, which the skill's Gotchas already documented as unavailable here.

## Testing
Verified on this machine against a real local SQL Server Express instance:
- [x] `dotnet build TallaEgg.sln --configuration Release` — 0 warnings, 0 errors
- [x] `driver.ps1 start` → `status` → `smoke` → `stop` — ended `errors 0`
- [x] Auto-quote for MAUA/IRT set enabled before a `smoke` run and observed restored to enabled afterwards
- [x] A deliberately broken build made `start` throw `Build failed (exit 1)` instead of launching stale binaries
- [x] `smoke --users 1`, `smoke --bogus 3`, and `start` with the ports already bound each refused with a specific message
- [x] `smoke --seed 7` confirmed to keep 5/5/10 rather than falling back to 100/120/1000
- [x] Follow-up `Invoke-RestMethod http://localhost:5140/api/orders/MAUA/IRT/best-prices` confirmed the simulated data is queryable afterward
- [x] `dotnet test tests/TallaEgg.AllServices.Tests/TallaEgg.AllServices.Tests.csproj` — 501/501 passing (unchanged)
- [x] No secrets/tokens in diff (confirmed — no config values were copied into these files; `config/appsettings.global.json` was never touched)

## Deployment Notes
Docs/tooling only — no service code changed, no migrations, no config changes.

## Known follow-ups (not in this PR)
- The launch + health-poll logic overlaps `.github/workflows/manual-test-run.yml`, which already encodes the same project paths, probe, and retry budget. Extracting one script under `scripts/` that both call would remove the divergence.
- `AGENT.md` still says "There is no working script" for starting the stack. That stays accurate for the human path (this driver is agent-facing and deliberately does not launch the bot), but is worth revisiting alongside the deduplication above.

### DIFF
--- a/.claude/skills/run-tallaegg/SKILL.md
+++ b/.claude/skills/run-tallaegg/SKILL.md
@@ -1,0 +1,176 @@
+---
+name: run-tallaegg
+description: Build, run, and drive the TallaEgg backend (Users/Wallet/Orders APIs plus the Telegram bot's own handlers) end to end without a live Telegram connection. Use when asked to start TallaEgg, run the services, smoke-test the bot, generate sample trading data, or verify a change actually works against the running app rather than just its unit tests.
+---
+
+TallaEgg is a Telegram bot (Persian-language gold trading) backed by three ASP.NET Core
+microservices (Users.Api, Wallet.Api, Orders.Api) and SQL Server, one database per service.
+There is no GUI to screenshot and no live Telegram connection available outside a real bot
+token — the agent path here is `.claude/skills/run-tallaegg/driver.ps1`, a thin wrapper that
+builds the solution, launches the three APIs, and drives them through
+**`TelegramBot.TallaEgg.TelegramBot.Simulator`** — a project the repo already ships, which
+replays real bot conversations (registration, admin approval, quote publishing, quote-fill
+trades, menu navigation) through the real `IBotHandler` and the real APIs/database, faking
+only the Telegram transport itself. That is the interaction surface: not a browser, not a
+window, a conversation replayed through the actual handler code.
+
+All paths below are relative to the repo root (`TallaEgg.sln`'s directory), not to this skill
+directory. The driver is Windows PowerShell (`powershell.exe`, not `pwsh`) — this project's
+native dev environment is Windows with a local SQL Server Express instance, per `AGENT.md`.
+
+## Prerequisites
+
+- .NET 9 SDK (`dotnet --version` -> `9.x`)
+- A reachable SQL Server instance. On a normal dev box this is SQL Server Express
+  (`Server=localhost\SQLEXPRESS`, Windows integrated auth) — confirm it's running:
+
+  ```powershell
+  Get-Service | Where-Object { $_.Name -like '*SQL*' } | Select-Object Name, Status
+  # -> MSSQL$SQLEXPRESS  Running
+  ```
+
+  Any reachable SQL Server works; only the connection strings in the config below need to match.
+
+## Setup
+
+`config/appsettings.global.json` is git-ignored (it holds live credentials — never commit it)
+and every service reads its own section from it. If it doesn't exist yet, create it from the
+template and point the four connection strings at your SQL Server:
+
+```powershell
+Copy-Item config/appsettings.global.example.json config/appsettings.global.json
+# then edit ConnectionStrings:{UsersDb,WalletDb,OrdersDb,AffiliateDb} to a real server
+```
+
+Each API runs its own EF Core migrations at startup (`context.Database.MigrateAsync()`), so
+there is no separate migrate step — an empty/nonexistent database is created and schema'd the
+first time `driver.ps1 start` launches it.
+
+The Simulator's config section (`Services:TallaEgg.TelegramBot.Infrastructure`) needs a
+syntactically-present `TelegramBotToken`, but it is never dialed for this path — the Simulator
+only constructs a `TelegramBotClient` and never calls `StartReceiving`/`GetMe`. The example
+file's placeholder (`REPLACE_WITH_TEST_BOT_TOKEN_FROM_BOTFATHER`) is fine as-is; verified by
+running the smoke test with exactly that value in place.
+
+## Build
+
+```powershell
+dotnet build TallaEgg.sln --configuration Release
+# -> Build succeeded. 0 Warning(s) 0 Error(s)
+```
+
+(`driver.ps1 start`, below, does this for you — this is only if you want the build on its own.)
+
+## Run (agent path)
+
+Everything goes through `driver.ps1`:
+
+```powershell
+& .claude/skills/run-tallaegg/driver.ps1 start    # build + launch Users/Wallet/Orders, wait for health
+& .claude/skills/run-tallaegg/driver.ps1 status    # confirm what's listening
+& .claude/skills/run-tallaegg/driver.ps1 smoke     # the actual interaction — see below
+& .claude/skills/run-tallaegg/driver.ps1 stop      # stop whatever is listening on the 3 ports
+```
+
+`start` builds in Release, launches the three APIs as background processes (stdout/stderr
+redirected to `run-logs/<service>.{out,err}.log`), and polls each service's Swagger page
+(`/api-docs/index.html`) until all three answer 200 or 60 seconds pass. On timeout it prints
+the last 40 lines of each service's stderr log and throws.
+
+`smoke` is the real interaction. With no arguments it runs a small simulation
+(`--users 5 --quotes 5 --trades 10 --seed 1`): registers 5 fake Telegram users through the real
+`/start` + contact-share flow, promotes one to admin, has admin approve/reject the rest, funds
+wallets, publishes quotes through the real `18500000-18550000`-style text command, fills trades
+by quote acceptance, and clicks through the help/history/balance menu buttons — end to end,
+through the real handler and API code, against the real database. Pass your own args through
+after `smoke`:
+
+```powershell
+& .claude/skills/run-tallaegg/driver.ps1 smoke --users 20 --quotes 20 --trades 50 --seed 7
+```
+
+A clean run ends with a line like:
+
+```
+=== Done in 00:00:10.26. Registered 5 (4 approved), trades attempted 10, errors 0 ===
+```
+
+`errors 0` is the thing to check. Every simulated user has `TelegramId < 0`, and each run's
+first phase (`DataReset`) wipes only rows with `TelegramId < 0` before it starts, so re-running
+`smoke` repeatedly against the same database is safe and self-cleaning — it can never touch a
+real (positive-id) user's data.
+
+Verified end-to-end on this machine, twice (bash-launched and PowerShell-launched), against a
+real local SQL Server Express instance — both runs finished with `errors 0` and left queryable
+data behind, confirmed via:
+
+```powershell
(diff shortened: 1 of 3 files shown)


MohKardan:
## Review — not ready to merge

Reviewed per `docs/process/CODE_REVIEW_GUIDE.md` at **level 2** (tooling/docs, no service code, no money/auth/schema path). No business rule is touched: the diff is two new files plus one `.gitignore` line, so zero-commission, the uncapped market-maker position, cross-asset credit and `LockBalance` are all untouched. No Persian text, no personal identity, and no credential literals in the diff.

### Claims I re-ran

| Claim | Result |
|---|---|
| `dotnet build TallaEgg.sln -c Release` → 0 warnings, 0 errors | Confirmed |
| `dotnet test ...` → 501/501 | Confirmed on this branch (`239fec3`) |
| `.gitignore` rationale: `*.log` already covered the log files, `launcher.pids` did not | Confirmed via `git check-ignore -v` |
| No secrets in the diff | Confirmed (only the example file's `REPLACE_WITH_...` placeholder is referenced) |
| **"It only ever touches rows with `TelegramId < 0`"** | **False — see [blocking] 1** |
| **`smoke` default is `--users 5 --quotes 5 --trades 10`** | **True only with zero args — see [blocking] 4** |

---

### [blocking] 1 — The headline safety claim is false: `smoke` permanently disables auto-quote

`driver.ps1:18` and `SKILL.md:98-101` both say the run "only ever touches rows with `TelegramId < 0`" and "can never touch a real user's data". But `Simulation.cs:61` calls `SetAutoQuoteEnabledAsync(Symbol, isEnabled: false, ...)` in Phase 2, and that is the only `SetAutoQuoteEnabled` call in the Simulator — nothing re-enables it. It persists through `IAutoQuoteSettingsRepository` as a per-symbol row in the Orders DB: global state, not `TelegramId`-scoped, and `DataReset` (`WHERE TelegramId < 0`) does not restore it.

So one `driver.ps1 smoke` leaves `AutoQuotePublisherService` (#90) off for MAUA/IRT permanently on that database, and nothing in the skill tells the reader to turn it back on. This matters because it is precisely the claim that licenses running against a real local dev database. Either restore the flag after the run, or state plainly what global state the run changes and how to undo it. The same sentence is in the PR description and should change with it.

### [blocking] 2 — A failed build is never detected, so `start` silently runs stale binaries

`driver.ps1:62` runs `dotnet build ... | Out-Host` and never checks `$LASTEXITCODE`. In Windows PowerShell 5.1 a native command's nonzero exit does **not** throw, even under `$ErrorActionPreference = 'Stop'` — verified on this box:

```
PS> $ErrorActionPreference='Stop'; cmd /c "exit 3"; "no throw, LASTEXITCODE=$LASTEXITCODE"
no throw, LASTEXITCODE=3
```

So on a compile error the build output scrolls past, `start` proceeds to `dotnet run --no-build`, the previous build comes up, the readiness poll passes, and it prints "All services up". An agent then "verifies a change actually works against the running app" (`SKILL.md:3`) against code that never compiled — the exact failure this skill exists to prevent. An `if ($LASTEXITCODE -ne 0) { throw ... }` after the build closes it.

### [blocking] 3 — The Setup recipe destroys `config/appsettings.global.json` if it already exists

`SKILL.md:41` (repeated at 166-168) is an unguarded `Copy-Item config/appsettings.global.example.json config/appsettings.global.json`. The "if it doesn't exist yet" condition is prose; the command is not. An agent following the skill verbatim on a box that already has the file silently overwrites live connection strings, bot token and `OwnerTelegramIds` with the template — and per `CLAUDE.md` that file is deliberately untracked since #33, so there is no `git checkout` to recover it. `-NoClobber`, or an `if (-not (Test-Path ...))` wrapper, makes the guard executable.

### [blocking] 4 — A partial argument override silently escalates the run ~100x

`driver.ps1:114` is all-or-nothing: `if ($Rest -and $Rest.Count -gt 0) { $Rest } else { <defaults> }`. So `driver.ps1 smoke --seed 7` passes only `--seed 7`, and `SimulationOptions.FromArgs` falls back to its own compiled defaults — **100 users, 120 quotes, 1000 trades** (`SimulationOptions.cs:9-11`) — not the 5/5/10 documented at `SKILL.md:86-88`. Someone expecting the ~10-second run this file documents gets the multi-minute 1000-trade run against their dev database. Merging `$Rest` over the defaults rather than replacing them fixes it.

### [blocking] 5 — `smoke` cannot fail

`Program.cs` ends `await simulation.RunAsync(options); return;` — exit code 0 regardless of `_errors.Count` — and `driver.ps1:116` checks neither `$LASTEXITCODE` nor the summary line. A run where every trade fails prints `errors 1000` and `driver.ps1 smoke` still returns success. Since `SKILL.md:98` says "`errors 0` is the thing to check", the driver should be what checks it; otherwise any automated caller reads a fully failed run as green.

---

### [suggestion] Smaller things

1. **`driver.ps1:90` tails only `*.err.log`, which is empty.** All three APIs use Serilog `.WriteTo.Console()` with no `standardErrorFromLevel` (`Users.Api/Program.cs:125`, `Wallet.Api/Program.cs:63`, `Orders.Api/Program.cs:201`), so Information *and* Error go to stdout. The documented troubleshooting path (`SKILL.md:169-172`) shows the operator three empty files while the cause sits in `*.out.log`.
2. **`stop` cannot clean up the one failure `start` throws on.** If the APIs die before Kestrel binds (SQL unreachable), `start` throws while leaving the processes alive; `stop` then matches nothing on the ports and reports "Stopped 0 process(es)". `launcher.pids` is written at line 78 and read by nothing — reading it back in `stop` would close this.
3. **`SKILL.md:175` recommends a command the same document calls a failure.** `smoke --users 1 --quotes 1 --trades 1` promotes the only user to admin, leaves `users.Skip(1)` empty, so `customers` is empty and `GenerateTradesAsync` records "No approved users to trade with." The run ends `errors 1`.
4. **The script's own `.DESCRIPTION` (lines 9-13) says `pwsh driver.ps1 ...` five times**, which `SKILL.md:141-144` documents as broken here. Confirmed — `pwsh` is not installed on this box (Windows PowerShell 5.1 only).
5. **The health probe depends on `ASPNETCORE_ENVIRONMENT=Development`.** All three APIs map Swagger inside `if (app.Environment.IsDevelopment())`. On a box where the environment is Production (this repo's own `scripts/windows-services/install-services.ps1` sets that for installed services) the services start fine and `start` times out regardless. Worth a probe that does not depend on it, or at least a note.
6. **`$stopped++` at line 126 fires even when `Stop-Process` was suppressed** by `-ErrorAction SilentlyContinue`, so `stop` can report kills it did not make.
7. **This duplicates `.github/workflows/manual-test-run.yml`**, which already encodes the same three project paths, the same `/api-docs/index.html` probe, the same 30×2s budget and the same `run-logs/` convention — and the two have already drifted (the workflow starts all three together; the driver sleeps 3s after users only, so Orders still races Wallet despite the comment at lines 73-74). One script under `scripts/` that both call would avoid a third copy.
8. **`AGENT.md:31` still reads "There is no working script"** and tells developers to open three terminals. If this launcher works, that line should change; if it is deliberately agent-only, the PR should say so.

### [question] Scope

There is no linked issue. `PR_TEMPLATE.md` asks for a "Linked Issue/Task" and `STANDARDS.md:103` lists "PR created with descriptive title and linked issue" — worth recording which issue or decision this closes, since it also introduces `.claude/` as a new tracked convention in a public repo. Nothing in the diff is out-of-scope work, though: the `.gitignore` line is a genuine consequence of the driver writing `run-logs/`.

PR #171, opened 2026-08-30
[Hotfix] Stop logging routine polling blips as Error — closes #148

### LINKED ISSUE #148: Every transient Telegram polling blip is logged at Error with a stack trace, so the bot's error log is 100% noise
## The measurement

Bot error lines on 2026-08-27:

```
total  [ERR]/[FTL] lines : 1214
of which "Polling error" : 1214
```

**Every single one.** The one signal an operator would scan for a real fault is entirely one recoverable network blip, roughly one every 50 seconds, each with a full stack trace.

The underlying exception is not a fault:

```
Telegram.Bot.Exceptions.RequestException: Exception during making request
 ---> System.Net.Http.HttpRequestException: An error occurred while sending the request.
 ---> System.Net.Http.HttpIOException: The response ended prematurely. (ResponseEnded)
```

The connection to `api.telegram.org` drops, the poller reconnects, and the bot keeps working — trades placed through it during that window settled normally.

## Two things in the handler

`TelegramBotHostedService.HandlePollingErrorAsync`:

```csharp
_logger.LogError(exception, "Polling error ({Source})", source);

if (exception.Message.Contains("timeout", StringComparison.OrdinalIgnoreCase) ||
    exception.Message.Contains("timed out", StringComparison.OrdinalIgnoreCase))
{
    _logger.LogWarning("Timeout detected. Waiting 10 seconds before retrying...");
    return Task.Delay(TimeSpan.FromSeconds(10), cancellationToken);
}
```

**1. Everything is `LogError`, unconditionally**, before any judgment about what happened. A dropped connection and an invalid bot token produce the same severity.

**2. The backoff is decided by substring-matching an exception message.** `ResponseEnded` contains neither "timeout" nor "timed out", so this failure takes no backoff at all — the branch that exists to slow down retries never runs for the failure that actually happens here. Matching on `Message` is also fragile against library upgrades and locale.

## What would close it

- **Severity by exception type, not message text.** `HttpRequestException`, `HttpIOException` and `TaskCanceledException` are transport blips — `Warning`, and without a stack trace, which says nothing useful about a dropped socket. `ApiRequestException` (401 on a rotated token, 429) is a real fault and should stay `Error` immediately.
- **Escalate on persistence, not on occurrence.** Count consecutive failures; log `Error` once the bot has genuinely stopped receiving updates for some threshold, and reset the counter on the first success. "Down for two minutes" is worth waking someone for; "one dropped connection" is not.
- **Apply the backoff to every transient case**, not only to whichever ones happen to have "timeout" in the message.

## Why it matters beyond tidiness

Same argument as the compiler warnings cleared in #141: when 1214 error lines are routine, the 1215th — the one that matters — is invisible. This log is the only place an operator would see that the bot stopped working, and right now it cannot be read.

## Note

The dropped connections themselves are an environment issue — this bot reaches Telegram through a local proxy — and are not what this issue asks to fix. It asks that a recoverable blip stop being recorded as a failure.

MohKardan:
## Description
`HandlePollingErrorAsync` logged every Telegram polling error unconditionally at `Error` with a full stack trace, and only backed off when the exception message happened to contain "timeout"/"timed out". Nearly every real failure is `TelegramBotClient.SendRequest` wrapping a dropped connection into a plain `RequestException` ("Exception during making request"), which matches neither string — so it got zero backoff and made the bot's error log 100% noise (1214/1214 error lines on 2026-08-27 were this one recoverable blip).

## Linked Task / Finding
Closes #148

## Type
- [x] Hotfix

## Changes
- Split polling-error severity by exception **type** instead of message-substring matching:
  - `ApiRequestException` (Telegram rejected the request — rotated token, rate limit) → `Error`, immediately, no backoff. This is a real fault.
  - Plain `RequestException` (transport-level: dropped connection, timeout, malformed response) → `Warning`, no stack trace, and now **always** backs off (10s), not only when the message text happened to say "timeout".
  - Any other exception type → stays at `Error`, so nothing outside those two known-transient shapes is silenced.
- Added escalate-on-persistence: a run of transient failures now logs a single `Error` once the bot has genuinely been failing to poll for 2 consecutive minutes, and resets once a gap between failures implies polling recovered in between (so a later outage can escalate again).
- Added `TelegramBotHostedServicePollingErrorTests` (5 tests) covering the severity split, the backoff behavior, and the escalate/reset timing (driven via an explicit timestamp parameter so the 2-minute threshold is tested deterministically, no real sleeping).
- Added two small test doubles (`CapturingLogger<T>`, `FakeHostApplicationLifetime`) under `tests/TallaEgg.AllServices.Tests/Fakes/`.

## Checklist (Definition of Done — see docs/process/STANDARDS.md)
- [x] Builds without warnings; tests pass (`dotnet test`) — 506/506 passing (501 previously + 5 new)
- [x] No secrets/tokens/credentials in the diff
- [x] Comments in English, explain the "why"
- [x] Follows naming conventions (STANDARDS.md)
- [x] Tests added/updated (unit)
- [ ] Docs updated — not applicable, no behavior change to a documented API
- [ ] Peer reviewed — pending

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/TelegramBotHostedService.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/TelegramBotHostedService.cs
@@ -5,6 +5,7 @@ using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Telegram.Bot;
+using Telegram.Bot.Exceptions;
 using Telegram.Bot.Polling;
 using Telegram.Bot.Types;
 using Telegram.Bot.Types.Enums;
@@ -23,6 +24,18 @@ public class TelegramBotHostedService : BackgroundService
     private readonly IHostApplicationLifetime _applicationLifetime;
 
     private CancellationTokenSource? _receiverCts;
+
+    // A gap this much longer than the retry delay between two transient polling failures means
+    // updates were flowing again in between, so the next failure starts a fresh streak rather
+    // than continuing the old one (issue #148).
+    private static readonly TimeSpan PollingRetryDelay = TimeSpan.FromSeconds(10);
+    private static readonly TimeSpan PollingRecoveryGap = TimeSpan.FromSeconds(30);
+    private static readonly TimeSpan PollingDownAlertThreshold = TimeSpan.FromMinutes(2);
+
+    private DateTimeOffset? _pollingFailureStreakStartedAt;
+    private DateTimeOffset? _lastPollingFailureAt;
+    private int _consecutivePollingFailures;
+    private bool _pollingDownAlertLogged;
 
     public TelegramBotHostedService(
         ITelegramBotClient botClient,
@@ -220,16 +233,60 @@ public class TelegramBotHostedService : BackgroundService
 
     private Task HandlePollingErrorAsync(ITelegramBotClient botClient, Exception exception, HandleErrorSource source, CancellationToken cancellationToken)
     {
-        _logger.LogError(exception, "Polling error ({Source})", source);
-
-        if (exception.Message.Contains("timeout", StringComparison.OrdinalIgnoreCase) ||
-            exception.Message.Contains("timed out", StringComparison.OrdinalIgnoreCase))
+        // ApiRequestException means Telegram itself rejected the request (rotated token, rate
+        // limit, ...). That's a real fault, not a transport blip, so it stays loud immediately.
+        if (exception is ApiRequestException apiException)
         {
-            _logger.LogWarning("Timeout detected. Waiting 10 seconds before retrying...");
-            return Task.Delay(TimeSpan.FromSeconds(10), cancellationToken);
+            _logger.LogError(exception, "Polling error ({Source}): Telegram API rejected the request ({ErrorCode}).", source, apiException.ErrorCode);
+            return Task.CompletedTask;
         }
 
+        // Every other RequestException is TelegramBotClient.SendRequest wrapping a transport-level
+        // failure -- a dropped connection, a timed-out request, a malformed response. These are
+        // routine on a flaky connection and recover on their own; logging each one at Error with a
+        // full stack trace was making the log 100% noise, burying the one entry that matters
+        // (issue #148). Log at Warning, without the stack trace, unless the bot has genuinely
+        // stopped receiving updates for a while -- that's worth escalating to Error once.
+        if (exception is RequestException)
+        {
+            return HandleTransientPollingFailure(source, exception, DateTimeOffset.UtcNow, cancellationToken);
+        }
+
+        // Anything else is unexpected for this handler -- keep it loud rather than risk silencing
+        // a real bug behind the transient-failure path above.
+        _logger.LogError(exception, "Polling error ({Source})", source);
         return Task.CompletedTask;
+    }
+
+    // `now` is passed in rather than read here so the escalation timing can be driven
+    // deterministically from a test without sleeping real wall-clock minutes.
+    private Task HandleTransientPollingFailure(HandleErrorSource source, Exception exception, DateTimeOffset now, CancellationToken cancellationToken)
+    {
+        if (_lastPollingFailureAt is null || now - _lastPollingFailureAt.Value > PollingRecoveryGap)
+        {
+            _pollingFailureStreakStartedAt = now;
+            _consecutivePollingFailures = 0;
+            _pollingDownAlertLogged = false;
+        }
+
+        _lastPollingFailureAt = now;
+        _consecutivePollingFailures++;
+        var downFor = now - _pollingFailureStreakStartedAt!.Value;
+
+        if (downFor >= PollingDownAlertThreshold && !_pollingDownAlertLogged)
+        {
+            _pollingDownAlertLogged = true;
+            _logger.LogError(exception,
+                "Polling has been failing for {Duration} ({Count} consecutive failures, {Source}); the bot may not be receiving updates.",
+                downFor, _consecutivePollingFailures, source);
+        }
+        else
+        {
+            _logger.LogWarning("Polling error ({Source}): {ExceptionType} - {Message}. Retrying in {Delay}...",
+                source, exception.GetType().Name, exception.Message, PollingRetryDelay);
+        }
+
+        return Task.Delay(PollingRetryDelay, cancellationToken);
     }
 }
 
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/TelegramBotHostedService.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/TelegramBotHostedService.cs
@@ -25,17 +25,35 @@ public class TelegramBotHostedService : BackgroundService
 
     private CancellationTokenSource? _receiverCts;
 
-    // A gap this much longer than the retry delay between two transient polling failures means
-    // updates were flowing again in between, so the next failure starts a fresh streak rather
-    // than continuing the old one (issue #148).
-    private static readonly TimeSpan PollingRetryDelay = TimeSpan.FromSeconds(10);
-    private static readonly TimeSpan PollingRecoveryGap = TimeSpan.FromSeconds(30);
+    // Backoff after a transient polling failure. The first retries are fast so an isolated blip --
+    // which the old code retried immediately -- does not cost the bot ten seconds of deafness, and
+    // it settles at the last entry once the failure looks sustained (issue #148).
+    private static readonly TimeSpan[] PollingRetryDelays =
+    [
+        TimeSpan.FromSeconds(1),
+        TimeSpan.FromSeconds(2),
+        TimeSpan.FromSeconds(5),
+        TimeSpan.FromSeconds(10),
+    ];
+
+    private static readonly TimeSpan PollingRetryDelayMax = PollingRetryDelays[^1];
+
+    // How long the bot must have been failing to poll before that is worth one Error line.
     private static readonly TimeSpan PollingDownAlertThreshold = TimeSpan.FromMinutes(2);
 
     private DateTimeOffset? _pollingFailureStreakStartedAt;
     private DateTimeOffset? _lastPollingFailureAt;
+    private DateTimeOffset? _lastUpdateReceivedAt;
     private int _consecutivePollingFailures;
     private bool _pollingDownAlertLogged;
+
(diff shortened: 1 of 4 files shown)


MohKardan:
## Review — request changes

Level 3 was not required (no money/auth/schema), but the change is the bot's only operational
signal, so I pulled the branch and ran it. **Scope is clean** — 4 files, all on-topic, no drive-by
refactors, nothing touching fees/credit/LockBalance/tokens, all text in English.

**Every number in the description reproduces**, which is worth saying explicitly:

| Claim | Verified |
|---|---|
| 506/506 passing | `dotnet test` on this branch → `Passed: 506, Failed: 0` |
| 501 previously | worktree at `origin/main` → `Passed: 501` |
| 5 new tests | 5 `[Fact]` in the new file |
| Builds without warnings | `dotnet build TallaEgg.sln` → `0 Warning(s)` |

The severity split itself is right, and `ApiRequestException` is checked before `RequestException`
(it derives from it), so the ordering holds. The problems are in the escalation half.

---

### [blocking] 1. The escalate-on-persistence Error is arithmetically unreachable on this bot

`ProxyBotClient` — the registered `ITelegramBotClient` (`Program.cs:99`) — sets
`HttpClient.Timeout = 120s` (`ProxyBotClient.cs:19`, `:41`, `:49`). `DefaultUpdateReceiver` passes
`(int)botClient.Timeout.TotalSeconds` as the `getUpdates` long-poll timeout, so a hung connection
throws only after ~120s. Add the new 10s backoff and consecutive failures arrive **~130s apart**.

At line 265, any gap over `PollingRecoveryGap` (30s) resets the streak — so
`_pollingFailureStreakStartedAt == now` and `downFor == 0` on every single failure. Escalating
needs `downFor >= 120s` while no gap can ever exceed 30s. Impossible unless the connection fails
*instantly* (refused / DNS).

I drove `HandleTransientPollingFailure` directly, one hour of **continuous, uninterrupted** failure:

```
cadence= 31s over 1h -> Error=0  Warning=117
cadence= 50s over 1h -> Error=0  Warning=73
cadence= 71s over 1h -> Error=0  Warning=51
cadence=110s over 1h -> Error=0  Warning=33
```

The cadence #148 actually measured — 1214 failures/day, roughly one per 71s — is in the resetting
regime. So is every cadence above 30s.

This matters because severity was downgraded to `Warning` *on the promise* that persistence
escalates. With escalation unreachable, a bot that is 100% down now produces **zero** Error lines,
where the old code at least produced noisy ones. #148 asked that the error log become readable; as
it stands the change makes it empty instead. That is a net loss of the exact signal the issue was
opened to protect.

The premise stated in the comment at line 27 — *"A gap this much longer than the retry delay ...
means updates were flowing again in between"* — only holds if a failing request fails fast. It does
not when the failure **is** a timeout, and "timeout" is listed in this PR's own description as one
of the cases this branch handles.

The deeper cause: recovery is *inferred* from the gap between failures rather than *observed*. No
value of `PollingRecoveryGap` can separate "poll succeeded, then failed" from "poll hung 120s, then
failed" — the two look identical from the error path alone. `HandleUpdateAsync` already sees every
successful delivery; clearing the streak there (or tracking a last-success timestamp) would give a
real signal instead of a proxy for one.

### [blocking] 2. `HandleErrorSource.FatalError` is silently downgraded to `Warning`

Telegram.Bot documents the three sources as:

- `PollingError` — *"Polling of updates will continue"*
- `HandleUpdateError` — *"Polling of updates will continue"*
- `FatalError` — *"A fatal uncaught exception occured somewhere. **Polling of updates will stop**"*

The new branch dispatches on exception type only. A `RequestException` arriving as `FatalError`
means the receive loop has died permanently — and it now yields one `Warning` line with no stack
trace, followed by a 10s `Task.Delay` against a loop that will never run again. Before this PR it
was an `Error` with a stack trace. `source` is already a parameter of
`HandleTransientPollingFailure` and is currently used only for string formatting; `FatalError`
should never take the transient path.

### [blocking] 3. The escalation tests pin a cadence production does not produce

`TelegramBotHostedServicePollingErrorTests.cs:106` says 10s apart is *"a realistic failure
cadence"*. With a 120s client timeout the real gap is ~130s, and #148's own measurement is ~71s.
Both escalation tests pass only because they chose the one cadence that keeps the streak alive.

Related fragility: the jump from `t=100` to `t=130` in `ConsecutiveTransientFailures_...` is
*exactly* `PollingRecoveryGap`, and the check at line 265 is a strict `>`. Change the constant to
29s, or the comparison to `>=`, and the test fails with "no Error was logged" rather than pointing
at the boundary.

A test at a 130s cadence asserting escalation still fires would have caught finding 1.

---

### [suggestion] 4. Short recoveries permanently disarm the alert

`_pollingDownAlertLogged` is reset only inside the `> PollingRecoveryGap` branch. Fail for 2
minutes → Error fires, flag set. Recover for 25s, fail again → 25s is not > 30s, so no reset: the
flag stays `true` and `_pollingFailureStreakStartedAt` stays at the original time. A six-hour
outage after that logs nothing. And if it ever does fire, it reports `"Polling has been failing for
06:00:00"` over a window in which polling was mostly healthy.

### [suggestion] 5. The Warning line carries no diagnostic information

`RequestException.Message` is a fixed literal — `"Exception during making request"` — as #148's own
stack trace shows. The real cause (socket reset vs. TLS failure vs. DNS vs. 407 proxy auth) lives
only in `InnerException`, and the new call passes no exception object at all. All ~1200 daily lines
become byte-identical. Logging `exception.InnerException?.Message` in the template would keep the
noise reduction without discarding the cause.

### [suggestion] 6. The first blip now costs 10s of blind time where it used to cost none

The old code returned `Task.CompletedTask` for anything not matching "timeout", so an isolated
dropped connection retried on the next loop iteration. Now every `RequestException` — including the
first of a streak — waits 10s, and the receiver awaits it. At the observed rate that is ~3.4h/day
of added deafness, and a customer who taps a keyboard button during a blip waits 10s instead of
milliseconds. A fast first retry with escalating backoff (1s → 2s → 5s → 10s) would fix the noise
without paying availability for it.

### [question] 7. `ApiRequestException` with no backoff

A rotated token (401) or a rate limit (429) fails instantly and gets `Task.CompletedTask`, and
`DefaultUpdateReceiver` has no backoff of its own — so the loop re-issues `getUpdates` immediately,
producing an unthrottled Error storm. **This is pre-existing, not introduced here** — the old code
also fell through to `Task.CompletedTask` for these. But this PR is what makes it deliberate, and
the comment names "rate limit" as a case that intentionally gets no delay. Given the PR's goal is a
readable error log, is leaving that path unthrottled intended? A 429 in particular is Telegram
explicitly asking us to wait.

---

### [nit] 8. Minor

- The comment at lines 27–29 sits above `PollingRetryDelay` but describes `PollingRecoveryGap`. The
  constant carrying the load in finding 1 ends up undocumented.
- `FakeHostApplicationLifetime.StopApplicationCalled` is never read by any test — the XML doc
  advertises behaviour nothing exercises.
- `CapturingLogger<T>` duplicates the private `CapturingLogger` in
  `MarketModeStartupValidatorTests.cs:47`. Leaving that one alone is the right call for scope here —
  noting it only so the duplication is recorded.

---

**Summary:** findings 1–3 are one defect wearing three hats — the escalation cannot fire at this
bot's real poll cadence, and the tests do not reveal it because they run at a cadence the bot does
not produce. Since the `Warning` downgrade was justified by that escalation, merging as-is trades a
100%-noise error log for a 0%-signal one. Findings 4–8 are independent and non-blocking.

Not merging. Happy to re-review once 1–3 are addressed.

MohKardan:
## Review findings addressed in `8b3c543`

All three `[blocking]` findings plus suggestions 4–6 and the nits. Finding 7 is left alone —
it is pre-existing behaviour and changing Telegram's 429 handling is not this issue's job.

### 1. The down-alert is now reachable (blocking)

The gap heuristic could not tell "poll succeeded, then failed" from "poll hung for a whole cycle,
then failed", so no single constant could work. Recovery is now established two ways:

- **an update actually arrived** — recorded in `HandleUpdateAsync`, which is unambiguous proof
  polling is alive, and is the signal the error path could never infer on its own;
- **or the gap exceeds one entire poll cycle** — the fallback for a bot with no traffic, now
  derived from `ITelegramBotClient.Timeout` instead of a fixed 30s, so it stays correct if
  `ProxyBotClient`'s timeout ever changes.

A streak resets only when neither holds. Re-running the same probe from the review, one simulated
hour of failures at each cadence:

| cadence | updates arriving | before | after |
|---|---|---|---|
| 31s | no | Error=0 | **Error=1** |
| 50s | no | Error=0 | **Error=1** |
| 71s (#148's measured rate) | no | Error=0 | **Error=1** |
| 110s | no | Error=0 | **Error=1** |
| 130s (real outage cadence) | no | Error=0 | **Error=1** |
| 71s | **yes** | Error=0 | Error=0 |
| 600s (sparse blips, quiet bot) | no | Error=0 | Error=0 |

A real outage escalates exactly once at every cadence; #148's actual pattern — blips with trades
still settling — and sparse overnight blips stay Warning-only. One alert per outage, not one per
failure, so the noise goal is intact.

### 2. `FatalError` no longer takes the transient path (blocking)

Checked before the type dispatch, since the exception on that path is shaped like a routine blip.
`"Polling of updates will stop"` now logs `Error` with the stack trace and returns immediately —
there is no receive loop left to back off for.

### 3. Tests run at the cadence production produces (blocking)

`PollCycle` and `OutageCadence` are derived from `ITelegramBotClient.Timeout` rather than
hardcoded, so no test can pass against a handler that only escalates at a rhythm the bot never
generates. The old `t=100 → t=130` boundary coincidence is gone. New coverage: sustained outage at
the real cadence, #148's pattern with updates flowing, sparse blips on a quiet bot, `FatalError`
severity, backoff progression, and the inner-cause message.

### 4–6. Suggestions

- `_pollingDownAlertLogged` now clears whenever the streak resets, and the streak resets on a real
  recovery signal — so a short recovery no longer disarms the alert permanently.
- The `Warning` line carries the innermost exception message, so a socket reset, a TLS failure and
  a DNS error are now distinguishable instead of ~1200 identical lines.
- Backoff is 1s → 2s → 5s → 10s instead of a flat 10s, which removes ~3.4h/day of added deafness
  while still throttling a sustained failure.

### 8. Nits

Comment moved to the constant it actually describes; `FakeHostApplicationLifetime`'s doc now
matches what it does. `CapturingLogger` duplication left as-is, as noted.

---

`dotnet build TallaEgg.sln` → `0 Warning(s)`. `dotnet test TallaEgg.sln` → **511/511 passing**
(501 on `main`, 10 in this file). Verified locally in a clean worktree.

Issue #147, opened 2026-08-27
Simulator only ever trades MAUA/IRT, so symbol-specific defects survive a clean run


MohKardan:
## What happened

`Simulation.cs` hard-codes one symbol:

```csharp
private const string Symbol = "MAUA/IRT";
```

Every quote it publishes, every trade it accepts, and the wallet it funds (`CurrenciesConstant.Maua`) use it. A full run is 100 users, 120 quotes and 1000 trades — **all gold**.

## Why that matters

On 2026-08-27 a run reported `trades attempted 1000, errors 0` and the database verified clean: no duplicate settlements, no orphans, every wallet reconciling to its transaction history.

**A single manual BTC trade immediately after found a real defect** (#146): `Orders.Amount` was `decimal(18, 2)` while wallets and trades were `decimal(28, 8)`, so a quantity of `2.111` was stored as `2.11` and the collateral difference was locked with nothing holding it and no way to release it.

The simulator could not have found it. **MAUA's precision is 2 — exactly what the column held** — so every gold quantity round-tripped unchanged. The bug only appears on a symbol needing more than two decimal places, and BTC needs eight.

That is the shape of the gap: a run that touches one symbol proves that symbol works, and reads as if it proved the platform works.

## What would close it

Trade across the symbols the platform actually supports, not one. `CurrenciesConstant.AllTradingPairs` already enumerates them with their own `BaseDecimalPlaces`, `MinQuantity` and `MinNotional`, so the run can pick from it rather than naming a symbol.

Two things worth getting right while doing it:

- **Quantities should use each symbol's own precision**, not a shared step. The value of the run is exercising `0.00000001`-scale quantities on BTC and `0.1`-scale on gold in the same pass — a fixed step would flatten the difference that catches these defects.
- **Wallet funding follows the symbol.** Phase 4 funds Toman and MAUA; it needs the base asset of whatever is being traded, and its `CREDIT_` ledger.

## Related

- #146 — the defect this would have caught
- #101 — the simulator itself

Issue #143, opened 2026-08-27
[C-4] Wallet has no concurrency token, but catches the exception as if it had one


MohKardan:
Traces to **C-4** in [`docs/operations/RE_AUDIT_2026-08.md`](../blob/main/docs/operations/RE_AUDIT_2026-08.md), recorded there as *"Partial — Orders only"*.

## What is there

`WalletRepository.cs:107` catches `DbUpdateConcurrencyException`:

```csharp
catch (DbUpdateConcurrencyException ex)
```

## What is not there

Nothing in the Wallet service configures a concurrency token. No `RowVersion`, no `[Timestamp]`, no `IsConcurrencyToken` — verified by searching the whole service.

EF only throws `DbUpdateConcurrencyException` when it detects that a tracked row changed underneath it, and it can only detect that if a concurrency token is configured. **Without one, that catch cannot fire for a lost update.** Two concurrent balance writes silently take the last one.

## Why it is worse than having no handler at all

A reader — human or agent — sees the catch and concludes the path is protected. It is not. The code makes a claim the schema does not support. Orders *is* protected, which makes the asymmetry easy to miss.

## The decision this needs

Either:

1. **Add the token.** `WalletEntity` gets a `RowVersion`, which needs a migration in the Wallet database, and the existing catch becomes real. Note the caution in `CLAUDE.md` about migrating one service's database in isolation.
2. **Delete the catch.** If optimistic concurrency is not wanted here, the handler should go, so nothing suggests a protection that does not exist.

Whichever is chosen, the present state — a handler for an exception that cannot be raised — should not survive.

## Related

- C-3 (transaction atomicity) is resolved on the live path; this is the separate question of two writers racing.
- Orders already has its token, so that side is done.

Issue #124, opened 2026-08-27
Market maker's negative position has no ceiling and no alerting: exposure grows silently


MohKardan:
## Summary

The market maker's position is unbounded by design, and nothing reports on it. Today the account
sits at **IRT −43,290,775** with no `CREDIT_IRT` ledger and therefore no ceiling. That is the
intended model — the market maker manages their own exposure, and the negative IRT balance *is*
the shop's book, representing what customers are owed. This issue is not a request to cap it.

What is missing is visibility. If customers sell heavily, the shop's liability grows silently and
the only way to notice is to query the database.

## Why this is worth a task

The negative balance is the one number in the system that represents real-world obligation. Every
other invariant is enforced somewhere: customers are checked against their credit at order time,
locks are guarded, settlement is atomic and idempotent. This one is deliberately unguarded, which
makes it the only place where a large, unexpected movement produces no signal at all.

Verified during the 6 Shahrivar log/database review: the ledger reconciles exactly and nothing is
wrong. The gap is purely that nobody would be told if it were.

## Suggested scope

- A threshold per asset, configurable, with a sensible default.
- When crossed, notify the market maker **and the other admins** — the point is that exposure
  should not be one person's private knowledge.
- An on-demand view of the current position, so it can be checked without a database query.
  The admin bot menu is the natural home.

Deliberately not proposed: a hard cap that refuses trades. That would change business logic, and
the current behaviour is intended.

## Notes

- Raised from the 6 Shahrivar review. Related to #35 (separating the house/fee account from the
  SuperAdmin role), which touches the same accounting seam.
- Fees are currently zero across all trades by design; the market maker's revenue is the spread.
  So the position, not accumulated commission, is the number that matters here.

Issue #61, opened 2026-07-28
Margin position monitoring: debt is created but never visible, watched, or recorded as settled


MohKardan:
## Summary

The system lets a customer's balance go negative (that is what the `CREDIT_` allowance is for), which means it creates **debt**. But nothing shows that debt, nothing watches it, and nothing records it being paid off. The gold shop carries the risk and has no instrument to see it.

## Two kinds of settlement, and only one exists

| | after the trade settles |
|---|---|
| **Spot** | nobody owes anybody — finished |
| **Margin (credit)** | a **standing debt** remains as a negative balance |

Trade settlement works and is identical for both — that is `SettleTradeAsync`, and credit changes nothing about it.

**Debt settlement is the part that has no model.** In this business it genuinely happens — physically, at the shop counter — but the system has no concept of "settling a position", no record of what was settled against what, and no way to see the debt in the first place. The only mechanism today is the admin manually depositing toman to lift the balance back up, which works but leaves no trace of *why*.

## Evidence from the current database

Net worth per account, valued at the current bid (18,410,360.59 IRT/g), including locked collateral:

| Account | net gold | net IRT | **net worth** |
|---|---|---|---|
| market maker | −95.80 g | +1,785,793,439 | **+22,080,894.478** |
| `09158527483` | +25.80 g | −490,512,029 | **−15,524,725.778** |
| `09037212111` | +70.00 g | −1,295,281,410 | **−6,556,168.700** |

```
customers combined = −22,080,894.478
market maker       = +22,080,894.478
```

Exactly equal and opposite, to the rial. **The ledger is sound** — this is simply the spread the market maker earned on every trade, which is the business model working as intended.

The point is the other one: **both customers are underwater and nothing in the system can show it.** Producing this table required hand-written SQL against two databases.

## What is missing

**1. Visibility.** No endpoint, report or admin screen answers "what does this customer hold, what do they owe, what is their net". The shop owner accepts the risk and has no instrument to see it.

**2. A threshold.** No number anywhere says "coverage dropped below X". Because the credit allowance is converted at the *live* price, a price move alone can push a position past its limit with no trade and no action by anyone — and nothing reacts.

**3. Settlement recording.** When a customer squares up in cash at the shop, the admin makes an ordinary deposit. Nothing links that deposit to the debt it cleared, so the history cannot be reconstructed later.

## Proposed scope — sized for 1–5 shops, not for an exchange

[...]

PR #172, opened 2026-08-30
[Hotfix][Critical] Plausibility band for auto-published quotes — issue #158

### LINKED ISSUE #158: Auto-published quotes have no plausibility band: a glitched price feed becomes tradeable within one tick
From `docs/audit/AUDIT_2026-08b.md` §10 (2026-08-29). Reasoned from code, **not reproduced** — reproducing it needs a mocked price provider, which does not exist yet.

## What happens

`AutoQuotePublisherService` takes whatever the first reference-price provider returns and publishes it. The only rejections are a price `<= 0` and an inverted spread. Nothing checks whether the number is *plausible*.

So a source that glitches once — a unit slip (per-gram vs per-mithqal), a stale cache, a decimal in the wrong place, a partial response parsed as a number — becomes a quote customers can trade against **within one 2-minute tick**.

## Why this one matters more than its severity suggests

Every other open finding costs bookkeeping, operator time, or trust. This one can move money directly, in the shop's disfavour, with no human in the loop: the customer who happens to be looking at the bot when a bad price lands is not doing anything wrong by accepting it.

It is also the only failure on this list that gets *worse* with success. More customers watching the bot means a shorter interval between a bad price appearing and someone trading on it.

## Where

- `src/Order/Orders.Application/Services/AutoQuotePublisherService.cs:107-108` — `decimal.Round(referencePrice.Value * (1 ± halfSpread), 2)`, published as-is
- Providers: `BrsApiPriceProvider`, `NerkhPriceProvider` — first one to answer wins

## Suggested fix

Cheap version, no new infrastructure:

- **Plausibility band** — refuse a reference price outside ±N% of the last published one, and log loudly rather than silently skipping the tick
- **Max move per tick** — clamp, or refuse and hold the previous quote
- **Cold start** — the first tick after a restart has nothing to compare against; either seed from the last stored quote or require an admin quote first

The threshold is a business number, not an engineering one. Gold does not move 20% in two minutes; something that says it did is a broken feed, not a market.

## Acceptance

- A provider returning a wildly out-of-band price does not produce a published quote
- The rejection is visible in logs with the rejected value and the band it violated
- A mocked provider makes this testable — which also closes the audit's "reasoned, not executed" gap on this item

## Related

#90 built the auto-publishing feature. #61 and #124 are about watching exposure after a trade; this is about not creating the bad trade in the first place.

---

## Also in this file: a catch that can never fire

Found 2026-08-30 while tracing this issue. Same method, worth fixing in the same PR rather than on its own.

`PublishIfDueAsync` ends with:

```csharp
catch (ArgumentException ex)
{
    // Quote.Publish's own validation (e.g. a zero/negative price from a source
    // returning garbage) rejects the quote. Logged and skipped, not thrown further —
    // the previous quote stays active until the next tick gets a sane price.
    _logger.LogWarning(ex, "Auto-quote for {Symbol} rejected by Quote.Publish; keeping the previous quote.", symbol);
}
```

`Quote.Publish` throws `BusinessRuleException` in all five of its validation paths, and `BusinessRuleException` is `sealed ... : Exception` — not an `ArgumentException`. **The handler cannot run.**

Nothing crashes: the exception is caught one level up by the loop's `catch (Exception ex)` (`AutoQuotePublisherService.cs:59`), which exists so one bad symbol cannot stop the others. But two things follow:

- The intended message — "keeping the previous quote", at Warning — is never logged. A garbage price that the domain correctly rejected is instead recorded by the generic handler, which reads as a service fault rather than a normal rejection.
- The comment describes behaviour the code does not have, so the next reader is misled about where price rejection is handled.

This is the same shape as #143: a handler for an exception that could not be raised. It matters more here because this issue is about making rejection *visible* — a rejection path that logs the wrong thing undercuts the fix.

Fix the catch to the type actually thrown, and keep the log level and wording the comment intended.

MohKardan:
**Branch Name**: `fix/auto-quote-plausibility-band`
**Linked Issue**: Closes #158

## Description

`AutoQuotePublisherService` published whatever the first reference-price provider returned. The only rejections were a price `<= 0` and an inverted spread, so a source that glitched once became a quote customers could trade against within one two-minute tick. This adds a plausibility band around the last published quote, and fixes an exception handler in the same method that could never run.

## Type of Change
- [x] Hotfix (urgent bug fix)

## Changes Made

- Added `AutoQuotePublisherService.IsPlausible`: a reference price more than **5%** from the mid of the symbol's last published quote is refused, and the tick keeps the previous quote.
- The rejection logs at Warning with the rejected value, the mid it was measured against, the deviation as a percentage, the band it violated, and the quote being kept — the issue's point is that silence is the worst outcome, so the log alone has to say what happened.
- Cold start: a symbol that has never had a quote publishes and logs at Information that the band was not applied. A restart needs no special case — the active quote is read from the database, so it survives one.
- Changed `catch (ArgumentException)` to `catch (BusinessRuleException)` in `PublishIfDueAsync`.
- Nine tests in `AutoQuotePublisherServiceTests`, including one at Bitcoin scale.

### The three decisions, and where they came from

All three were put to the product owner before any code was written, per the issue's "the threshold is a business number, not an engineering one":

| Decision | Answer | Why |
|---|---|---|
| Band width | **±5%** | Gold does not move 5% in two minutes. The failure modes the issue names sit far outside it: a per-mithqal figure read as per-gram is ≈4.33x, a misplaced decimal is 10x. Wide enough that a real gap-up after a quiet period is not refused. |
| Breach behaviour | **Refuse, hold the previous quote** | Clamping publishes a price no source ever reported, which the shop is then bound to honour; and a feed stuck on a bad number would walk the quote towards it one tick at a time until it arrived. |
| Cold start | **Accept and log** | Refusing would mean auto-quote could never bootstrap a newly activated symbol. The exposure is one tick, on a symbol that has never had a quote. |

The band is a `const` rather than configuration or an `AutoQuoteSettings` column: adding a key to `config/appsettings.global.json` would make every deployment that has not been updated fail at startup (this project deliberately does not default missing configuration), and a per-symbol column needs a bot command to go with it. Both are larger than this issue.

### The catch that could never fire

Confirmed before changing it, not taken from the issue text:

- `Quote.Publish` throws `BusinessRuleException` in all five validation paths — `Quote.cs:65,68,71,77,81`.
- `BusinessRuleException` is `sealed class BusinessRuleException : Exception` — `BusinessRuleException.cs:29`. Not an `ArgumentException`.

So the handler was unreachable. Nothing crashed, because the poll loop's `catch (Exception)` at `AutoQuotePublisherService.cs:59` caught it one level up — but a price the domain correctly rejected was recorded at **Error**, reading as a service fault rather than the ordinary rejection it is, and the Warning the comment described was never written. Same shape as #143.

### Unit Tests
- [x] New unit tests written — 9 added

```
Passed!  -  Failed: 0, Passed: 520, Skipped: 0, Total: 520
```

511 before this change, 520 after.

**Every new test was verified to fail without the fix**, so none of them is decoration:

| Fix disabled | Result |
|---|---|
| Band widened to a no-op | 5 failed: `RejectsAReferencePriceAboveTheBand…`, `RejectsAReferencePriceBelowTheBand…`, `ARejectedPriceIsLoggedAsAWarningWithTheValueAndTheBand`, `PublishesAMoveExactlyOnTheBandEdge…`, `TheBandComesFromTheStoredQuote…` |
| `catch (BusinessRuleException)` reverted to `ArgumentException` | 1 failed: `APriceRejectedByQuoteDotPublishIsLoggedAsAWarningInsteadOfThrowing`, with the `BusinessRuleException` escaping uncaught |

The audit marked this finding **"reasoned, not executed"** for want of a mockable provider. One already existed — `AutoQuotePublisherServiceTests` has driven the real tick logic against a stub provider and SQLite in-memory since #90 — so these tests close that gap by execution rather than by reasoning.

The `catch` test triggers the rejection the way the handler's own comment describes: a reference price of `0.001` passes the chain's positive-price check and then rounds to `0.00` on both sides of the spread, so `Quote.Publish` rejects it. That is a real path, not a contrived one.

### Integration / smoke
- [x] `dotnet build TallaEgg.sln` → **0 Warnings, 0 Errors** (TreatWarningsAsErrors is on)
- [x] Full stack run via `driver.ps1 start` + `smoke --users 20 --quotes 20 --trades 60 --seed 158`

```
=== Done in 00:00:18.35. Registered 20 (16 approved), trades attempted 60, errors 0 ===
```

Database checked afterwards:

| Check | Result |
|---|---|
| Duplicate settlements (same `TradeId` twice) | 0 |
| Settlement legs per trade, today's run | 60 references, all exactly 4 legs |
| Wallet balance vs last transaction's `BallanceAfter` | 0 mismatches |
| Transaction chain breaks (`BallanceAfter` → next `BallanceBefore`) | 0 |
| Negative locked balances | 0 |
| Orphan orders from today's run | 0 |

Two anomalies found in the database are **pre-existing and unrelated to this change**: four orders in `Completed` with no trade, and four settlement references carrying two transaction legs instead of four. All eight rows are timestamped `2026-08-27 17:26–17:34`, three days before this branch, and the local databases are never reset between runs (oldest row `2026-08-26`). Today's 60 trades are clean. Flagged rather than fixed here, per scope discipline.

Bitcoin was tested separately because the simulator only trades MAUA/IRT (#147), so a symbol-specific arithmetic problem would survive a clean smoke run. `TheBandBehavesTheSameAtBitcoinScale` exercises the band at ~5.2e10 IRT in both directions, including a 10x decimal slip.

### Environment
- [x] Tested locally on Windows, against local SQL Server Express

## Security Checklist
- [x] No sensitive data in logs — the messages carry prices and a symbol, nothing user-identifying

## Code Quality
- [x] Comments are in English and explain the "why"

## Breaking Changes?
- [x] No breaking changes

No schema change, no configuration change, no API change.

## Deployment Notes

**Pre-Deployment**: nothing. No migration, no new configuration key, no feature flag.

**Post-Deployment Verification**: watch the Orders service log for
`Auto-quote for {Symbol} rejected: reference price …` at Warning. A steady stream of those means
either a genuinely broken feed or a band that is too tight for that symbol — the constant is
`AutoQuotePublisherService.MaxDeviationPercent`.

**Rollback**: revert the commit. Behaviour returns to publishing every price the provider returns.

## Related Issues

- #90 built the auto-publishing feature this guards
- #61 and #124 watch exposure after a trade; this is about not creating the bad trade
- #143 is the same shape as the catch fixed here — a handler for an exception that could not be raised
- #147 is why Bitcoin needed a test of its own

### DIFF
--- a/src/Order/Orders.Application/Services/AutoQuotePublisherService.cs
+++ b/src/Order/Orders.Application/Services/AutoQuotePublisherService.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
@@ -49,6 +50,37 @@ public class AutoQuotePublisherService : BackgroundService
     /// </para>
     /// </summary>
     private const decimal MaxDeviationPercent = 5m;
+
+    /// <summary>
+    /// How many consecutive band rejections mean the feed, and not the market, is what changed.
+    ///
+    /// <para>
+    /// One rejection is a glitch and holding the previous quote is the whole point. Three in a
+    /// row — six minutes of a source insisting on the same implausible level — is not transient,
+    /// and the system cannot tell a real repricing from a persistently broken feed: the provider
+    /// chain returns the first source that answers and never compares two against each other.
+    /// </para>
+    ///
+    /// <para>
+    /// So it stops rather than guesses (product owner, issue #158). Continuing to hold the quote
+    /// would be the worse of the two: during a genuine fast move a stale price is exactly what a
+    /// customer arbitrages, and only in the direction that costs the shop. An outage an admin can
+    /// end in seconds is recoverable; a run of trades at a price nobody meant to offer is not.
+    /// </para>
+    /// </summary>
+    private const int MaxConsecutiveRejections = 3;
+
+    /// <summary>
+    /// Consecutive band rejections per symbol, reset by any tick that publishes.
+    ///
+    /// <para>
+    /// Deliberately in memory rather than on <see cref="AutoQuoteSettings"/>: a restart clearing
+    /// it only costs three more held ticks before the same stop happens, which does not justify a
+    /// column and a migration. The stop itself is persisted, so what a restart must not lose —
+    /// that the shop decided to stop quoting this symbol — is not held here.
+    /// </para>
+    /// </summary>
+    private readonly ConcurrentDictionary<string, int> _consecutiveRejections = new();
 
     private readonly IServiceScopeFactory _scopeFactory;
     private readonly ILogger<AutoQuotePublisherService> _logger;
@@ -124,7 +156,15 @@ public class AutoQuotePublisherService : BackgroundService
         var quoteRepo = scope.ServiceProvider.GetRequiredService<IQuoteRepository>();
 
         if (!IsPlausible(symbol, referencePrice.Value, await quoteRepo.GetActiveAsync(symbol)))
+        {
+            await StopQuotingIfRejectionsPersistAsync(symbol, quoteRepo, settingsRepo, settings);
             return;
+        }
+
+        // Any published tick clears the streak, so only *consecutive* rejections count towards
+        // stopping. A single outlier between good ticks is the transient case the band already
+        // handles by holding the previous quote.
+        _consecutiveRejections.TryRemove(symbol, out _);
 
         // referencePrice is already Toman per traded base unit (a gram of gold, a whole coin, a
         // whole Bitcoin) — each provider does its own unit conversion, so nothing here is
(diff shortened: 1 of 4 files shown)


MohKardan:

Three fixed, one declined with reasoning. Pushed in `7757d16`.

---

### 1. The band latches — **fixed**, and it was worse than reported

Confirmed by execution before fixing anything. A throwaway probe published a quote at 1,000,000, then ran thirty ticks against a perfectly stable market 10% higher:

```
After 30 ticks at a stable 1,100,000 the tradeable quote is
buy=995000.0 sell=1005000.0 ... Total quote rows: 1
```

An hour of simulated time, a market that had not moved once since the first rejection, and the shop still quoting the old price. Also verified the two supporting claims rather than taking them: `PublishedAt` appears nowhere outside ordering and DTO mapping — there is no expiry anywhere — and `QuoteFillService.cs:83` sells against whatever `GetActiveAsync` returns with no age check. The latch is permanent by construction: the anchor only moves when a quote is published, and publishing only happens when the band passes.

The remedy was put to the product owner, because every option is a business trade-off and the system genuinely cannot distinguish a real repricing from a persistently broken feed — `ReferencePriceProviderChain` returns the first source that answers and never compares two against each other.

**Decision: after three consecutive rejections, stop quoting the symbol.** The active quote is deactivated so nothing stale stays tradeable, and auto-quoting is switched off so the next tick cannot publish an unanchored price through the cold-start path. Logged at Error with the recovery instruction.

Stopping beats holding, and beats eventually accepting:

- Holding is what the probe above showed — unbounded exposure at a price nobody meant to offer.
- Accepting after N ticks re-admits exactly the failure this issue is about, six minutes later.
- Stopping fails closed, the way `SettleTradeAsync`'s fee guard already does in this codebase. A quoting outage an admin ends in seconds is recoverable; a run of trades at a stale price is not, and during a fast move a stale quote is precisely what gets arbitraged, in only one direction.

Recovery reuses `AutoQuoteSettings.IsEnabled` — the switch whose own documentation describes it as the one to reach for "so a bad reading or a misbehaving provider can be turned off from inside the product". An admin publishes a price by hand and sends `اتومات روشن`. No deploy, no database access, no new configuration.

The rejection counter is in memory rather than a new column: a restart clearing it costs three more held ticks before the same stop happens, which does not justify a migration. What must survive a restart — that the shop decided to stop quoting — is persisted.

Six new tests: two rejections keep quoting, three stop it, a good price after the stop still publishes nothing, an accepted tick clears the streak so only *consecutive* rejections count, and one symbol stopping leaves another quoting. All six verified to fail with the threshold disabled.

---

### 2. The anchor is the mid, and a manual quote can be arbitrarily wide — **not taking the proposed remedy**

The observation is right: `Quote.Publish` caps no spread, so an admin could publish a quote whose mid is a price no source reported.

Measuring from the nearer of buy/sell instead would fix that, but it makes the safety band's width depend on `SpreadPercent`, which is an unrelated, admin-editable, per-symbol setting. A symbol configured with a 10% spread would silently get a much wider plausibility band than one at 0.5% — the band would stop meaning "5%" and start meaning "5% plus whatever the spread happens to be". That is a worse property for a guard whose whole job is to be a fixed, stated number.

The case that motivates the finding is also self-limiting now. An absurd manual quote makes the band reject honest references — which is finding 1, and finding 1's fix ends it after three ticks by stopping the symbol and calling for a human, which is the right outcome for a symbol whose last published quote is nonsense.

Worth revisiting if per-symbol bands ever land, since the coupling argument disappears once the band is configured next to the spread.

---

### 3. Rounded deviation reads as a contradiction — **fixed**

`decimal.Round(deviationPercent, 2)` → `4`. A marginal breach printed `is 5.00% away ... outside the plausibility band of ±5%`. Four places, and a comment saying why, so nobody helpfully rounds it back.

---

### 4. Log assertions too loose to fail — **fixed**

Correct and worth catching: `Assert.Contains("100", …)` matches inside `"1000000"`, and `Assert.Contains("5", …)` inside `"995000"`, so both would have passed with the deviation and the band absent from the message. Each number is now asserted with the words that give it its meaning:

```csharp
Assert.Contains("reference price 2000000 is", warning.Message);
Assert.Matches(@"is 100(\.0+)?% away from the last published mid 1000000", warning.Message);
Assert.Contains("plausibility band of ±5%", warning.Message);
Assert.Matches(@"Keeping the previous quote \(buy 995000(\.0+)?, sell 1005000(\.0+)?\)", warning.Message);
```

The regexes rather than plain equality are for `decimal` scale: the same value renders as `995000` or `995000.00` depending on where it came from.

Issue #42, opened 2026-07-26
Settlement idempotency is not race-safe: check runs outside the transaction and has no unique constraint


MohKardan:
> Found during the code review of #41 (merged). Not a demo blocker — today's single-instance, sequential outbox processor makes the race practically unreachable — but this must be fixed before real money, because the failure mode is **duplicated money**.

## Summary
`WalletRepository.SettleTradeAsync` guarantees idempotency with a read-then-write check that is (a) executed **outside** the database transaction and (b) **not backed by a unique constraint**. Two concurrent settlements of the same trade can therefore both pass the check and both apply, crediting the participants twice.

## Why this is a problem
Settlement idempotency is the entire reason redelivery is safe. The outbox processor is explicitly allowed to redeliver: if the wallet call succeeds but the `OutboxMessage` row fails to save, the message stays `Pending` and is retried. The design assumes the wallet will detect "already settled" — so idempotency is load-bearing, not a nicety.

Today it is enforced only by a `SELECT` executed before the transaction opens:

```csharp
// src/Wallet/Wallet.Infrastructure/WalletRepository.cs  (~line 207)
var alreadySettled = await _context.Transactions.AnyAsync(t => t.ReferenceId == referenceId);
if (alreadySettled) return (true, "Trade already settled.");
...
await using var tx = await _context.Database.BeginTransactionAsync();   // ~line 232 — transaction starts AFTER the check
```

And there is no database-level safety net — the index is non-unique:

```csharp
// src/Wallet/Wallet.Infrastructure/WalletDbContext.cs  (~line 69)
modelBuilder.Entity<Transaction>().HasIndex(t => t.ReferenceId);   // NOT .IsUnique()
```

**Interleaving that duplicates money:**
```
A: SELECT ReferenceId=T  -> none
B: SELECT ReferenceId=T  -> none          (A hasn't committed yet)
A: settle, COMMIT                          (4 transactions written, balances moved)
B: settle, COMMIT                          (4 more transactions, balances moved AGAIN)
```
Result: both parties credited twice; locked collateral consumed twice (or driven negative). Money is created out of nothing.

[...]

Issue #36, opened 2026-07-25
Credit model: enforce credit limit inside the wallet (single signed ledger), not via a separate CREDIT_ asset + order-time check


MohKardan:
## Summary
The credit system (letting a customer trade on a limit and settle physically offline) is the right idea for the gold-shop model, but its current implementation enforces the credit limit in the wrong place and in a fragile way. This issue proposes migrating to the standard model: a per-wallet **credit limit** with the invariant `Balance >= -CreditLimit` enforced **atomically inside the wallet**, instead of a separate `CREDIT_` asset checked only at order validation.

## How it works today (verified in code)
- **Storage**: credit is a *separate wallet asset* — `CREDIT_MAUA`, `CREDIT_IRR` (`CurrenciesConstant`, `Utils.CREDIT_PREFIX`). The admin `ش` command deposits into `CREDIT_<currency>` (`BotHandlerAdmin.cs`).
- **Validation at order time** — `WalletApiClient.ValidateCreditAndBalanceAsync` (line ~262) sums four balances:
  - buy power (quote): `spotQuote + creditQuote + creditBase * price >= amount * price`
  - sell power (base): `spotBase + creditBase + creditQuote / price >= amount`
- **Lock**: `WalletEntity.LockBalance` lets the spot `Balance` go negative (guard commented out: "چون اعتبار داریم…").
- **Settlement**: only touches the spot assets (MAUA/IRR). `CREDIT_` is never modified. The debt is represented by the negative spot `Balance`.

So usage is recorded as a negative spot balance, and the `CREDIT_` asset acts as a static limit that validation adds to availability.

### 1. The limit is enforced only at order validation, not in the wallet (highest risk)
There is no `Balance >= -CreditLimit` invariant in `WalletEntity`. `LockBalance` and `ConsumeLockedBalance` drive `Balance` negative with no check; the only hard guard (`DecreaseBalance`) floors at **zero**, not at `-credit`, and is bypassed on the trade path. If any code path skips `ValidateCreditAndBalanceAsync` or that check is wrong, a user can exceed their credit limit and nothing stops it. The limit must be an invariant the wallet enforces itself.

### 2. Validation and lock are not atomic (concurrency hole)
Validation is one HTTP call; lock is a separate one. Two concurrent orders can both pass validation against the same credit, then both lock → the user exceeds the limit. (Same family as the audit's C-4/C-5 concurrency findings.)

### 3. Cross-asset credit conversion at live price
[...]

PR #173, opened 2026-08-30
[Hotfix][Critical] Idempotency for admin top-ups and deductions — issue #157

### LINKED ISSUE #157: Deposits and withdrawals have no idempotency key: one retry double-credits a customer
Finding **F1** of [`docs/audit/AUDIT_2026-08b.md`](../blob/main/docs/audit/AUDIT_2026-08b.md) (2026-08-29, methodology v7). Reproduced by execution.

> **Corrected 2026-08-30.** This issue originally proposed a unique index on
> `(UserId, Asset, ReferenceId)` as the fix. That would do nothing. Tracing the live path
> afterwards showed the bot never sends a `ReferenceId` at all, so every deposit row holds
> `NULL` and no two of them can collide. The defect is larger than the audit described and the
> remedy is different — it needs a product decision first. The original reproduction still
> stands; only the proposed fix was wrong.

## What happens

`POST /api/wallet/deposit` accepts a `ReferenceId`, stores it on the transaction row, and never checks whether it has been used before. Calling it twice with the same reference credits the wallet twice.

The audit's harness ran exactly that: `IncreaseBalanceAsync` twice with one `ReferenceId` produced a balance of **2,000,000** instead of 1,000,000, and **two** transaction rows sharing that reference.

`POST /api/wallet/withdrawal` has the same shape.

## The live path is worse: there is no key at all

`BotHandlerAdmin.cs:71` — the admin charge command — sends this and nothing else:

```csharp
await _walletApi.DepositeAsync(new WalletRequest
{
    Asset  = CurrenciesConstant.CreditAssetFor(currency),
    Amount = amount,
    UserId = userDto.Id
});
```

No `ReferenceId`. `WalletApiClient.DepositeAsync` serializes the request as-is and adds nothing. So in production the column is always `NULL`.

`Transaction.Create` concedes the same point: it requires a non-empty `referenceId` only for `TransactionType.Trade`. Deposits are allowed to have none, and do.

**Consequence:** a unique index over `ReferenceId` cannot help. `NULL`s do not collide, so it would constrain nothing; and there is no key for an endpoint-level check to compare either. Idempotency has to be given to this path before it can be enforced.

## Why it can happen in practice

The bot issues a single `PostAsync` with no automatic retry, so this is not a machine-driven duplicate. The realistic trigger is a **lost response**: the server commits, the reply never arrives, the admin sees no confirmation and re-sends the charge command.

That re-send is a *new* Telegram message with the same content. Any key derived from the message id would differ, and would not deduplicate anything. The key has to come from the content.

## The decision this needs before any code

> **If an admin charges the same customer the same amount twice within N minutes, is that one duplicate or two real charges?**

The answer sets the deduplication window, and there is no safe default: too short and the lost-response case slips through; too long and a legitimate second top-up is silently swallowed. **Ask the product owner and wait for an answer — do not pick a number.**

## Then, and only then, the mechanical part

Settlement already solves the enforcement half, and the audit re-proved it works: `TradeSettlement.HasKey(s => s.TradeId)` (`WalletDbContext.cs:89`) makes the database itself the barrier, and a second settlement of the same trade returns *success* with "Trade already settled", moving no money. Deposits should behave the same way — a duplicate is a success that does nothing, never an error the admin has to interpret.

Once a key exists, the constraint can enforce it.

## One trap for whoever implements this

**Do not put a unique index on `ReferenceId` alone.** Settlement writes **four** transaction rows sharing one reference (`WalletRepository.cs:328`, `referenceId = tradeId.ToString()`) — buyer/quote, buyer/base, seller/base, seller/quote. A bare unique index would break every trade in the system.

Those four rows are distinct in `(UserId, Asset, ReferenceId)`. Verify that before relying on it.

## Code

- `src/Wallet/Wallet.Api/Program.cs:180` — the endpoint, no dedup
- `src/Wallet/Wallet.Application/WalletService.cs:39` — `IncreaseBalanceAsync`, stores `refId` and checks nothing
- `src/Wallet/Wallet.Infrastructure/WalletDbContext.cs:77` — `HasIndex(t => t.ReferenceId)`, indexed but not unique
- `TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs:71` — the caller that sends no key

## Acceptance

- The admin charge path sends a deduplication key derived from the command's content
- A second charge carrying that same key leaves the balance unchanged and returns success
- A trade still settles, writing its four legs under one reference
- Tests covering both, alongside the existing settlement-idempotency test
- Withdrawal closed the same way

## Not blocked on engineering

Nothing here waits on #36, #105 or a server. It waits on one product answer about the window.

MohKardan:
**Branch Name**: `fix/deposit-withdrawal-idempotency`
**Linked Issue**: Closes #157

## Description

`POST /api/wallet/deposit` accepted a `ReferenceId`, stored it on the transaction row, and never checked whether it had been used before, so calling it twice with the same reference credited the wallet twice. Withdrawal had the same shape. This gives the admin path a deduplication key, makes the wallet apply a reference only once, and puts a unique index behind it.

## Type of Change
- [x] Hotfix (urgent bug fix)

## The two facts the issue asks to be re-checked first

Both verified against this branch's code and against the live local database, not taken from the issue text.

**1. The bot sent no key at all.** [`BotHandlerAdmin.cs:71`](../blob/main/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs#L71) sent `Asset`, `Amount` and `UserId` and nothing else; the deduction branch at line 153 the same. `WalletRequest` has had a `ReferenceId` property the whole time and nobody filled it. In the local database:

| | rows | with a null `ReferenceId` |
|---|---|---|
| Deposits | 39 | **39** |
| Withdrawals | 3 | **3** |

A unique index over `ReferenceId` would have constrained nothing, because no two NULLs collide, and there was no key for an endpoint check to compare either. Confirmed.

**2. Settlement writes four legs under one reference.** [`WalletRepository.cs:328`](../blob/main/src/Wallet/Wallet.Infrastructure/WalletRepository.cs#L328) sets `referenceId = tradeId.ToString()` and lines 434/439/444/449 write buyer-quote, buyer-base, seller-base and seller-quote under it. Confirmed — and confirmed further that those four are distinct in `(WalletId, ReferenceId)`, since settlement refuses a self-trade at line 361 so the four wallets are always four different rows. Every reference in the local `Transactions` table was distinct under that pair.

Rather than trust that reasoning, the trap was reproduced: switching the index to `ReferenceId` alone **failed nine settlement tests**, including `SettleTradeAsync_HappyPath`, both idempotency tests and `DifferentTrades_BetweenTheSameParties_BothSettle`. The issue's warning is exactly right, and `SettlementSurvivesReferenceIndexTests` now pins it.

## Changes Made

- **`AdminAdjustmentKey`** (new, `TallaEgg.Core`) — builds the key from the command's content (customer, asset, amount) plus a five-minute bucket. Deposits and withdrawals are namespaced apart so that topping up and deducting the same amount in one window are never confused.
- **`BotHandlerAdmin`** — both the charge (`ش`) and deduct (`د`) commands now send that key.
- **`WalletRepository.ApplyWithIdempotencyAsync`** — applies a wallet change and its transaction, treating an already-recorded reference as a success that moves nothing and returns the original transaction. Same shape as `SettleTradeAsync`: fast-path check, then the database as the actual guarantee, then a duplicate-key catch that reports the winner.
- **`WalletService.IncreaseBalanceAsync` / `DecreaseBalanceAsync`** — route through it.
- **`WalletDbContext`** — unique index on `(WalletId, ReferenceId)`, filtered to non-null, plus migration `20260830210630_AddTransactionReferenceUniqueIndex`.

### Why five minutes

The product owner's number, for the failure the issue describes: the wallet commits, the reply is lost, the admin sees no confirmation and re-sends. That is seconds later, comfortably inside five minutes, while two genuinely separate top-ups of exactly the same amount to the same customer that close together are rare. Too short lets the retry through; too long silently swallows a real charge.

### Why the key contains a time bucket, and what that costs

The re-send is a *new* Telegram message, so anything derived from the message id would differ and deduplicate nothing — the key has to come from the content. But content alone cannot separate a re-send from a genuine second top-up of the same amount next week, so time has to be in the key, and the timestamp is rounded down to a five-minute bucket.

A bucket is not a sliding window. Two sends that straddle a boundary — 12:09:58 and 12:10:03 — get different keys and the duplicate is not caught. The exposure is the retry gap divided by the window, so a few seconds against five minutes. Closing it would mean the wallet querying by key prefix, trading a clean contract for that last few percent. `ARetryThatStraddlesABucketBoundaryIsNotCaught` pins the limitation rather than hiding it, and the failure mode is the one that exists today, not a new one.

### Why the index filter is not optional

SQL Server treats NULLs as **equal** in a unique index, so an unfiltered `(WalletId, ReferenceId)` unique index would allow only one reference-less transaction per wallet — breaking every lock, unlock and unreferenced deposit after the first. Confirmed in the created index:

```
name              : IX_Transactions_WalletId_ReferenceId
is_unique         : True
filter_definition : ([ReferenceId] IS NOT NULL)
Cols              : WalletId, ReferenceId
```

### Unit Tests
- [x] New unit tests written — 25 added

```
Passed!  -  Failed: 0, Passed: 536, Skipped: 0, Total: 536
```

511 before this change, 536 after.

**Verified load-bearing by removing the protection**, in three separate configurations:

| What was disabled | Result |
|---|---|
| The application pre-check only | All 7 idempotency tests still pass — proving the guarantee really is the database, not the check |
| Pre-check **and** unique index | 4 fail, including `TheSameDepositReferenceTwiceCreditsOnce` with **Expected 1000000, Actual 2000000.0** — the audit's exact reproduction |
| Index narrowed to `ReferenceId` alone | 9 settlement tests fail — the trap the issue warns about |

### Live check against the running stack

Migration applied to the real SQL Server database at startup, then the endpoints driven directly:

```
POST /api/wallet/deposit (first)  -> success, balanceAfter 1000000, trackingCode 54JB6X381ES9
POST /api/wallet/deposit (repeat) -> success, balanceAfter 1000000, trackingCode 54JB6X381ES9

balance      : 1000000.00000000   (credited once)
transactions : 1

a different reference still credits  -> 2000000
withdrawal, sent twice               -> deducted once, 1 transaction row
```

The repeat returns the *original* transaction — same tracking code, same balances, same `UpdatedAt` — so an admin who re-sent sees the confirmation they should have seen the first time, and nothing to interpret.

That `UpdatedAt` is a fix this check surfaced. On the first live run the repeat answered with a timestamp 145ms later than the stored one, because `IncreaseBalance` adjusts `Balance` and stamps `UpdatedAt` before the repository sees the entity and that change is never saved on the idempotent path. The entity is now reloaded, so the response describes the wallet as stored. `TheRepeatReportsTheWalletAsStoredRatherThanAsItWouldHaveBeen` covers it.

### Simulator

```
=== Done in 00:00:15.51. Registered 20 (16 approved), trades attempted 60, errors 0 ===
new outbox failures: 0
```

Database afterwards:

| Check | Result |
|---|---|
| Duplicate settlements | 0 |
| Rows colliding on `(WalletId, ReferenceId)` | 0 |
| Settlement legs per trade, this run | 72 references, all exactly 4 |
| Wallet balance vs last transaction's `BallanceAfter` | 0 mismatches |
| Transaction chain breaks | 0 |
| Negative locked balances | 0 |
| Orphan orders from this run | 0 |

The simulator drives the real `BotHandler`, so its Phase 5 admin charge/discharge produced real keys in the database — the fix working end to end, not just in unit tests:

```
admin-deposit:7d80a4ad6ec74f10a43cc3a5815803f4:CREDIT_MAUA:2:213079071
admin-withdrawal:7d80a4ad6ec74f10a43cc3a5815803f4:IRT:17820:213079071
...
```

Withdrawals now carry a reference on 3 of 3 rows, against 0 of 3 before. The 36 deposits still holding NULL are the simulator's own `FundWalletsAsync`, which calls the API directly rather than through the admin command — outside this issue's scope.

### Environment
- [x] Tested locally on Windows against local SQL Server Express

## Security Checklist
- [x] No sensitive data in logs — the key contains a user id, an asset and an amount, no phone number or name
- [x] Input validation unchanged; the key is built server-side from already-validated values

## Code Quality
- [x] No new dependencies

## Breaking Changes?
- [x] No breaking changes

`ReferenceId` was already on `WalletRequest` and optional; callers that send none behave exactly as before, which `DepositsWithNoReferenceAreAllApplied` pins.

## Deployment Notes

**Pre-Deployment**: one migration, `20260830210630_AddTransactionReferenceUniqueIndex`, applied automatically at Wallet.Api startup.

**It can fail on a database that already holds a colliding pair.** Check before deploying:

```sql
SELECT WalletId, ReferenceId, COUNT(*)
FROM Transactions
WHERE ReferenceId IS NOT NULL
GROUP BY WalletId, ReferenceId
HAVING COUNT(*) > 1;
```

This returned 0 rows locally. If it returns rows in production, they are duplicate applications of one reference — the defect this issue is about — and need reconciling before the index can be created.

**Post-Deployment Verification**: send the same `ش` command twice within five minutes and confirm the customer's credit moves once. `sys.indexes` should show `IX_Transactions_WalletId_ReferenceId` as unique with the `IS NOT NULL` filter.

**Rollback**: `dotnet ef database update 20260828140441_AddWalletConcurrencyToken` drops the index, then revert the commit. Nothing else has to be undone: the keys already written are ordinary reference strings that older code ignores.

## Related Issues

- #42 built the settlement idempotency this copies
- #36 is unaffected — no per-asset balance rule is introduced here

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -1,4 +1,4 @@
-﻿using System.Text;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using TallaEgg.Core;
@@ -68,11 +68,19 @@ namespace TallaEgg.TelegramBot
                 var userDto = await _usersApi.GetUserAsync(phone);
                 if (userDto != null)
                 {
+                    var creditAsset = CurrenciesConstant.CreditAssetFor(currency);
+
                     var result = await _walletApi.DepositeAsync(new TallaEgg.Core.Requests.Wallet.WalletRequest
                     {
-                        Asset = CurrenciesConstant.CreditAssetFor(currency), // شارژ ادمین همیشه اعتباری‌ست
+                        Asset = creditAsset, // شارژ ادمین همیشه اعتباری‌ست
                         Amount = amount,
-                        UserId = userDto.Id
+                        UserId = userDto.Id,
+
+                        // Without this the column was always NULL in production and the same charge
+                        // sent twice credited twice (issue #157). The realistic duplicate is an admin
+                        // re-sending after a reply that never arrived, so the key comes from what they
+                        // typed rather than from the message.
+                        ReferenceId = AdminAdjustmentKey.ForDeposit(userDto.Id, creditAsset, amount, DateTime.UtcNow)
                     });
                     if (result.Success)
                     {
@@ -154,7 +162,10 @@ namespace TallaEgg.TelegramBot
                     {
                         Asset = currency,
                         Amount = amount,
-                        UserId = userDto.Id
+                        UserId = userDto.Id,
+
+                        // Same deduplication as the charge command above (issue #157).
+                        ReferenceId = AdminAdjustmentKey.ForWithdrawal(userDto.Id, currency, amount, DateTime.UtcNow)
                     });
                     if (result.Success)
                     {
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -92,20 +92,29 @@ namespace TallaEgg.TelegramBot
                         var amountText = $"{PersianFormat.Amount(amount, currency)} {PersianFormat.Unit(currency)}";
                         var newCreditText = $"{PersianFormat.Amount(result.Data?.BalanceAfter ?? 0m, currency)} {PersianFormat.Unit(currency)}";
 
+                        // A deduplicated repeat reports success, because the charge did happen — on the
+                        // earlier send, with nothing moving on this one. Telling the customer their
+                        // credit rose again would be a lie about their own money, so only the admin is
+                        // told, and told that it was a repeat (issue #157).
+                        var alreadyApplied = result.Data?.WasAlreadyApplied ?? false;
+
                         await _messenger.SendAsync(
                            message.Chat.Id,
-                           string.Format(BotMsgs.MsgAdminChargeDone,
+                           string.Format(alreadyApplied ? BotMsgs.MsgAdminChargeAlreadyApplied : BotMsgs.MsgAdminChargeDone,
                                PersianFormat.Asset(currency),
                                amountText,
                                PersianFormat.Ltr(PersianFormat.ToPersianDigits(phone)),
                                newCreditText));
 
-                        await _messenger.SendAsync(
-                           userDto.TelegramId,
-                           string.Format(BotMsgs.MsgUserCreditIncreased,
-                               PersianFormat.Asset(currency),
-                               amountText,
-                               newCreditText));
+                        if (!alreadyApplied)
+                        {
+                            await _messenger.SendAsync(
+                               userDto.TelegramId,
+                               string.Format(BotMsgs.MsgUserCreditIncreased,
+                                   PersianFormat.Asset(currency),
+                                   amountText,
+                                   newCreditText));
+                        }
                     }
                     else
                     {
@@ -177,20 +186,27 @@ namespace TallaEgg.TelegramBot
                         var deductAmountText = $"{PersianFormat.Amount(amount, currency)} {PersianFormat.Unit(currency)}";
                         var newBalanceText = $"{PersianFormat.Amount(result.Data?.BalanceAfter ?? 0m, currency)} {PersianFormat.Unit(currency)}";
 
+                        // Same as the charge command above: a deduplicated repeat moved nothing, so the
+                        // customer is not told a second deduction happened (issue #157).
+                        var alreadyApplied = result.Data?.WasAlreadyApplied ?? false;
+
                         await _messenger.SendAsync(
                                message.Chat.Id,
-                               string.Format(BotMsgs.MsgAdminDeductDone,
+                               string.Format(alreadyApplied ? BotMsgs.MsgAdminDeductAlreadyApplied : BotMsgs.MsgAdminDeductDone,
                                    PersianFormat.Asset(currency),
                                    deductAmountText,
                                    PersianFormat.Ltr(PersianFormat.ToPersianDigits(phone)),
                                    newBalanceText));
 
-                        await _messenger.SendAsync(
-                               userDto.TelegramId,
-                               string.Format(BotMsgs.MsgUserBalanceDeducted,
-                                   PersianFormat.Asset(currency),
-                                   deductAmountText,
-                                   newBalanceText));
+                        if (!alreadyApplied)
+                        {
+                            await _messenger.SendAsync(
+                                   userDto.TelegramId,
+                                   string.Format(BotMsgs.MsgUserBalanceDeducted,
+                                       PersianFormat.Asset(currency),
+                                       deductAmountText,
+                                       newBalanceText));
+                        }
 
 
                     }
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Constants.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Constants.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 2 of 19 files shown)


MohKardan:

All four were real and all four are fixed, in `161a18c`.

---

### 1. Repeated deduction refused for a deduction that had succeeded — **fixed**

The most serious of the four, and it made the fix actively worse than the bug in one case. The check ran *after* `wallet.DecreaseBalance(amount)`, which enforces its own floor, so a re-sent command taking most of a balance was refused.

Reproduced before and after, by moving the check back:

```
Failed  ARepeatedDeductionOfMostOfTheBalanceStillSucceeds
  TallaEgg.Core.ErrorHandling.BusinessRuleException : مقدار کسر از حساب بیشتر از حد مجاز است
```

Deducting 800 of a customer's 1000 leaves 200; the retry asks `DecreaseBalance` to take 800 from 200 and is refused — and the admin is told the deduction failed when it had already gone through, which is precisely the interpretation problem this issue exists to remove. The review is also right that my own test dodged it by deducting 400 from 1000.

Both `IncreaseBalanceAsync` and `DecreaseBalanceAsync` now check before touching the entity, through a shared `AlreadyAppliedAsync`. The repository keeps its own check and duplicate-key catch, which are what make the *concurrent* case safe — the service check is not the guarantee, only the thing that stops a repeat from being rejected by a domain rule before it can be recognised.

### 2. The bot told the customer money moved when it had not — **fixed**

Correct, and a defect this PR introduced: before it, the second send genuinely credited, so `MsgUserCreditIncreased` was true. Now it would be a false statement to a customer about their own balance.

`WalletBallanceDTO` carries `WasAlreadyApplied`, set from whether the transaction the wallet returned is the one this call created. Both the charge and deduct branches now:

- send the admin `MsgAdminChargeAlreadyApplied` / `MsgAdminDeductAlreadyApplied` instead of the "done" message — not an error, since the charge did happen, plus a line saying to wait a few minutes if a second one was actually intended;
- send the customer **nothing**, because nothing moved.

`ADeduplicatedChargeTellsTheAdminButNotTheCustomer` and `AnOrdinaryChargeStillTellsBoth` cover both directions.

Confirmed on the live endpoint:

```json
{"balanceAfter":1000000,"trackingCode":"OKQAIZPOQ0GN","wasAlreadyApplied":false}
{"balanceAfter":1000000.00000000,"trackingCode":"OKQAIZPOQ0GN","wasAlreadyApplied":true}
```

### 3. Benign duplicates logged at Error — **fixed**

Right, and the comparison to #148 is apt: a reference arriving twice is the normal outcome this design expects, and routing it through `catch (Exception) { LogError; throw; }` on the way to being absorbed would put a stack trace in the log — and a Telegram alert — for a success. `UpdateWalletAsync` now recognises it, logs at Debug, and rethrows for `ApplyWithIdempotencyAsync` to absorb.

### 4. Index key over the 900-byte limit — **fixed**

`ReferenceId` mapped to `nvarchar(450)` = 900 bytes, plus 16 for `WalletId` = 916, past the limit on SQL Server 2014 and earlier. Bounded to 256 (512 bytes, 528 with the key). The local server is SQL Server 2025, where the limit is 1700 and the original would have worked — the "plausible" marking was the right call, and the fix costs nothing.

No truncation risk: the longest `ReferenceId` in the local database is **70** characters, and 0 rows exceed 256. Trade ids are 36, admin adjustment keys around 110.

The migration is regenerated as `20260830213754_AddTransactionReferenceUniqueIndex` and carries the `AlterColumn` as well as the index. The earlier one had already been applied locally, so it was reverted with `dotnet ef database update` before being replaced, and the new one re-applied cleanly at startup:

```
migration applied  : 20260830213754_AddTransactionReferenceUniqueIndex
ReferenceId column : nvarchar(256)
```

---

### Re-verification after the fixes

```
Passed!  -  Failed: 0, Passed: 541, Skipped: 0, Total: 541
```

536 before these fixes, 541 after — five more tests, of which `ARepeatedDeductionOfMostOfTheBalanceStillSucceeds` was verified to fail against the code as reviewed.

Simulator, with the outbox drained either side:

```
=== Done in 00:00:12.55. Registered 20 (18 approved), trades attempted 60, errors 0 ===
new outbox failures: 0
```

| Check | Result |
|---|---|
| Duplicate settlements | 0 |
| Colliding `(WalletId, ReferenceId)` pairs | 0 |
| Settlement legs per trade | 72 references, all exactly 4 |
| Balance vs last transaction | 0 mismatches |
| Transaction chain breaks | 0 |
| Negative locked balances | 0 |
| Orphan orders from this run | 0 |

Issue #175, opened 2026-08-31
Simulator runs are not independent: DataReset races in-flight settlements, and smoke reports green anyway


MohKardan:
Found on 2026-08-31 while running the simulator repeatedly for #158 and #157. Reproduced by execution, including on unmodified `main`.

**This is a test-harness defect, not a money bug.** The product behaved correctly throughout — see "The product was fine" below. What is broken is that the harness can corrupt its own database, and then report the run as green anyway.

## What happens

Running two simulations back to back, with no pause between them, does two things:

1. The previous run's queued settlements fail permanently — 94 of them in one session.
2. The next run's reset can crash outright with a foreign-key violation.

Both come from the same root cause: **the run is not finished when the simulator says it is.** The summary line is printed when the trading phase ends, but settlement happens asynchronously through the Orders outbox and drains on its own schedule, tens of seconds later.

## Bug 1 — `DataReset` deletes data that queued outbox messages still need

`DataReset.RunAsync` (`DataReset.cs:36-49`) deletes Users, Wallets, Transactions, Trades and Orders for `TelegramId < 0`. It does **not** touch `OutboxMessages`, and it does not wait for in-flight messages to finish.

So the previous run's settlement messages are still queued, and when the processor reaches them the wallets they need are gone:

```
Wallet settlement rejected: Buyer locked IRT (0) is less than required (10544633).
```

Five retries each, then `Status = Failed`. In one session that produced **94** permanently failed messages, all with that error:

| Day | Failed outbox messages |
|---|---|
| 2026-08-27 | 16 (pre-existing, unrelated) |
| 2026-08-30 | **94 (this)** |

`OutboxMessages` is never cleaned by the reset, so they accumulate indefinitely and every later inspection of `Status = 2` shows them.

### The race, on top of that

`DeleteWalletDataAsync` (`DataReset.cs:65-93`) deletes `Transactions` and then `Wallets` in two separate statements. The outbox processor writes new `Transactions` rows between them, and the second delete fails:

```
The DELETE statement conflicted with the REFERENCE constraint
"FK_Transactions_Wallets_WalletId". The conflict occurred in database
"TallaEggWallet", table "dbo.Transactions", column 'WalletId'.
```

This killed a run on unmodified `main`, so it is not branch-specific.

[...]

Issue #174, opened 2026-08-31
Market maker's wallet is a hot row: quote fills intermittently fail after the concurrency retry is exhausted


MohKardan:
Found while working on #158 and #157 (2026-08-31). Reproduced by execution, on `main`, unmodified.

## What happens

A customer accepts a quote, and the fill fails with:

> در حال حاضر امکان انجام این معامله نیست.

That message comes from `QuoteFillService.cs:172`, the branch where the **market maker's** counterparty order could not be created. The customer's own order is cancelled and their collateral released, so nothing is left half-done — but a perfectly valid trade was refused, and the customer is told to try later with no reason they can act on.

## Why

The market maker is the counterparty to *every* quote fill in the market, so its wallet rows are a single hot row for the whole shop. `WalletRepository.WithConcurrencyRetryAsync` retries an optimistic-concurrency loss three times (`ConcurrencyAttempts = 3`, 20/40ms backoff). Under a burst, three attempts are not enough, `DbUpdateConcurrencyException` escapes, and order creation fails.

The wallet log shows both halves during a run: 13 `Concurrent write to the {Asset} wallet ... retrying from a fresh read` warnings, and several `DbUpdateConcurrencyException` stack traces that got past the retry limit.

This is the intended failure mode of the retry limit, and the comment on `ConcurrencyAttempts` says so — "a wallet contended by more than that at once is a load problem, not a retry problem". The observation here is that the market maker's wallet *is* that wallet by design, on every single trade, so the load problem is structural rather than exceptional.

## Reproduction

`driver.ps1 smoke --users 20 --quotes 20 --trades 60` — 60 fills in about five seconds, all against one market maker.

Failure rate on cold first-runs after `driver.ps1 start`:

| | runs | failed |
|---|---|---|
| `main` (c1da35d) | 3 | 1 |
| a feature branch that does not touch this path | 4 | 3 |

The failing trade index varies between runs at the same seed (#36, #39, #50), which is what marks it as timing-dependent rather than a data or seed problem. The branch/main split above is not meaningful — the branch's changes were shown not to execute in this path at all (no log output from them in any run) — it is the same flake in both.

## Where

- `src/Wallet/Wallet.Infrastructure/WalletRepository.cs` — `ConcurrencyAttempts = 3`, `WithConcurrencyRetryAsync`
- `src/Order/Orders.Application/Services/QuoteFillService.cs:164-173` — the counterparty order, and the message the customer sees

## Worth deciding

[...]

PR #176, opened 2026-08-31
[Hotfix] Deduplicated admin adjustments reported a stale balance and logged nothing — follow-up to #157

MohKardan:
**Branch Name**: `fix/deduplicated-adjustment-reporting`
**Linked Issue**: Follow-up to #157

## Description

Two defects in the idempotency #157 added, both found by driving the real bot rather than by reading the code: a deduplicated repeat showed the admin a stale balance under a label saying "current", and left no trace at all in the wallet service log.

## Type of Change
- [x] Hotfix (urgent bug fix)

---

## Bug 1 — a stale balance labelled "current"

A repeat reports the *original* transaction. That is deliberate and matches settlement. But that transaction's `BalanceAfter` is the balance at that earlier moment, and anything that has happened since is not in it.

Observed against the running bot, in this order:

| step | effect |
|---|---|
| deduct 500 | 184,887,309 → 184,886,809 |
| deduct 1500 | 184,886,809 → **184,885,309** |
| re-send the 500 (×3) | deduplicated, nothing moved |

Each repeat answered `موجودی فعلی: ۱۸۴٬۸۸۶٬۸۰۹` — the balance from before the 1500 came out. The customer's actual balance was 184,885,309.

The number was right for what it described and wrong for what it was labelled. **An admin re-sends precisely when they are unsure what has taken effect**, so this is the worst moment to hand them a figure whose label says "now" and whose value means "then". Here the gap was 1,500 toman because that is what happened to be deducted in between; nothing bounds it.

### Fix

`WalletBallanceDTO` gains `CurrentBalance` — what the wallet holds — alongside `BalanceAfter`, which keeps its meaning as what the reported operation produced. The two are equal except on a repeat, so nothing else changes shape.

The bot quotes `CurrentBalance` in the two already-applied messages only. `MsgAdminChargeDone` and `MsgAdminDeductDone` are untouched, because on an ordinary application the two figures are the same number.

The labels are also now specific — `اعتبار کنونی کاربر` / `موجودی کنونی کاربر` rather than `اعتبار فعلی` / `موجودی فعلی` — so the sentence says whose balance and as of when.

## Bug 2 — the repeat was invisible in the wallet service

`/code-review` on #157 found that a repeated deduction was refused by `DecreaseBalance` when the amount exceeded what was left. The fix was to move the idempotency check up into `WalletService`, before the entity is touched. Correct, but it meant the ordinary repeat now returns *before* reaching the repository — and the repository held the only log line.

So the case that actually happens became silent in the system of record. The bot log has the reply; the wallet service, which is where anyone would look to confirm what happened to a balance, had nothing. Confirmed on the session above: three deduplicated repeats, zero lines in `wallet-api-*.log`.

### Fix

`AlreadyAppliedAsync` logs the hit at Information with the reference, the wallet's asset and owner, and the id of the transaction being reported. The repository keeps its own line for the concurrent case, which is a different path and still reachable.

`WalletService` now takes an `ILogger<WalletService>`. Only two places construct it outside DI, both tests, both updated. (`.audit-work/repro` also constructs it but is untracked and outside the solution.)

### Unit tests
- [x] New unit tests written — 8 added
- [x] All unit tests pass

```
Passed!  -  Failed: 0, Passed: 563, Skipped: 0, Total: 563
```

555 before this change, 563 after.

**Verified load-bearing.** With both fixes reverted:

```
Failed  DepositIdempotencyTests.ADeduplicatedRepeatIsLogged
Failed  AdminChargeCommandTests.ADeduplicatedChargeQuotesTheBalanceAsItStandsNow
Failed  AdminChargeCommandTests.ADeduplicatedDeductionQuotesTheBalanceAsItStandsNow
```

`ARepeatReportsTheBalanceAsItStandsNow_NotAsTheOriginalLeftIt` reproduces the exact live sequence at the service level: deduct 500, deduct 1500, re-send the 500, then assert that `BalanceAfter` is still 9,500 (the original, correctly unchanged) while `CurrentBalance` is 8,000 (what the wallet holds).

Two tests guard against over-correcting: `WithNothingInBetweenTheRepeatsTwoFiguresAgree` and `AnOrdinaryApplicationReportsTheBalanceItProduced` pin that the ordinary path reads exactly as before. `AnOrdinaryApplicationLogsNothing` keeps the new log line a signal rather than noise.

### Live check

The same sequence through the running Wallet API:

```
deduct  500 -> balanceAfter 9500  currentBalance 9500  alreadyApplied False
deduct 1500 -> balanceAfter 8000  currentBalance 8000  alreadyApplied False
re-send 500 -> balanceAfter 9500  currentBalance 8000  alreadyApplied True

what the wallet actually holds      : 8000
what the repeat reported as current : 8000
the stale figure it used to report  : 9500
```

## Security Checklist
- [x] No sensitive data in logs — the new line carries a reference key, an asset, a user id and a transaction id; no phone number or name

## Code Quality
- [x] Follows `STANDARDS.md`
- [x] No new dependencies

## Breaking Changes?
- [x] No breaking changes

`CurrentBalance` is an added field; existing consumers reading `BalanceAfter` are unaffected, and it keeps its old meaning. `WalletService`'s constructor changed, but it is only ever resolved through DI in production.

## Deployment Notes

No migration, no configuration, no schema change. Revert the commit to roll back.

**Post-deployment verification**: send the same `ش` or `د` command twice inside five minutes with something else in between, and check the second reply quotes the balance the customer actually holds. `wallet-api-*.log` should carry one Information line per deduplicated repeat.

## Related

- #157 introduced the idempotency this corrects
- #174 and #175 are the two unrelated defects found in the same session

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -98,13 +98,20 @@ namespace TallaEgg.TelegramBot
                         // told, and told that it was a repeat (issue #157).
                         var alreadyApplied = result.Data?.WasAlreadyApplied ?? false;
 
+                        // The repeat reports the original transaction, so its BalanceAfter is the credit
+                        // as it stood at that earlier moment. Anything charged or deducted since is
+                        // missing from it, which is why the repeat quotes CurrentBalance instead.
+                        var creditNowText = alreadyApplied
+                            ? $"{PersianFormat.Amount(result.Data?.CurrentBalance ?? 0m, currency)} {PersianFormat.Unit(currency)}"
+                            : newCreditText;
+
                         await _messenger.SendAsync(
                            message.Chat.Id,
                            string.Format(alreadyApplied ? BotMsgs.MsgAdminChargeAlreadyApplied : BotMsgs.MsgAdminChargeDone,
                                PersianFormat.Asset(currency),
                                amountText,
                                PersianFormat.Ltr(PersianFormat.ToPersianDigits(phone)),
-                               newCreditText));
+                               creditNowText));
 
                         if (!alreadyApplied)
                         {
@@ -190,13 +197,19 @@ namespace TallaEgg.TelegramBot
                         // customer is not told a second deduction happened (issue #157).
                         var alreadyApplied = result.Data?.WasAlreadyApplied ?? false;
 
+                        // Same reason as the charge command above: on a repeat the figure that means
+                        // "what this customer holds" is CurrentBalance, not the earlier BalanceAfter.
+                        var balanceNowText = alreadyApplied
+                            ? $"{PersianFormat.Amount(result.Data?.CurrentBalance ?? 0m, currency)} {PersianFormat.Unit(currency)}"
+                            : newBalanceText;
+
                         await _messenger.SendAsync(
                                message.Chat.Id,
                                string.Format(alreadyApplied ? BotMsgs.MsgAdminDeductAlreadyApplied : BotMsgs.MsgAdminDeductDone,
                                    PersianFormat.Asset(currency),
                                    deductAmountText,
                                    PersianFormat.Ltr(PersianFormat.ToPersianDigits(phone)),
-                                   newBalanceText));
+                                   balanceNowText));
 
                         if (!alreadyApplied)
                         {
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -101,8 +101,14 @@ namespace TallaEgg.TelegramBot
                         // The repeat reports the original transaction, so its BalanceAfter is the credit
                         // as it stood at that earlier moment. Anything charged or deducted since is
                         // missing from it, which is why the repeat quotes CurrentBalance instead.
+                        //
+                        // Falling back to BalanceAfter, not to zero: a wallet that predates
+                        // CurrentBalance sends no such field, and the services restart independently.
+                        // The stale figure is the thing being fixed here, but telling an admin their
+                        // customer holds nothing would be the worse of the two answers.
+                        var creditNow = result.Data?.CurrentBalance ?? result.Data?.BalanceAfter ?? 0m;
                         var creditNowText = alreadyApplied
-                            ? $"{PersianFormat.Amount(result.Data?.CurrentBalance ?? 0m, currency)} {PersianFormat.Unit(currency)}"
+                            ? $"{PersianFormat.Amount(creditNow, currency)} {PersianFormat.Unit(currency)}"
                             : newCreditText;
 
                         await _messenger.SendAsync(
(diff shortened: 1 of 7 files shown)


MohKardan:

Both fixed in `83d0ac3`. The first is worse than "low" — it turns a wrong-but-plausible number into a wrong-and-alarming one, in a deployment state this project actually has.

---

### 1. `?? 0m` on a brand-new response field — **fixed**

The scenario is real, not theoretical: `scripts/windows-services/install-services.ps1` installs the services individually, so they restart independently and a bot running ahead of the Wallet API is an ordinary state, not a mistake. In that window the response carries no `currentBalance`, a non-nullable `decimal` takes it as `0`, and the deduplicated reply tells the admin their customer holds nothing.

That is a worse failure than the one this PR fixes. A stale balance is at least a number the customer had at some point; a zero is a number they never had, arriving in the reply an admin reads when they are already unsure what happened.

And the review's reasoning about why is exactly right: a non-nullable `decimal` cannot distinguish "the wallet did not send this" from "the wallet holds nothing". So:

- `WalletBallanceDTO.CurrentBalance` is now `decimal?`, with the deployment reason written into the XML doc so nobody helpfully un-nullables it.
- Both call sites read `result.Data?.CurrentBalance ?? result.Data?.BalanceAfter ?? 0m`. A version-skewed deployment degrades to the behaviour that shipped in #157 — the stale figure — rather than to a fabricated zero.

`AWalletThatSendsNoCurrentBalanceFallsBackInsteadOfReportingZero` covers it, with a stub switch (`OmitCurrentBalance`) that answers the way an older wallet does. Verified load-bearing: restoring `?? 0m` fails it.

### 2. `AnOrdinaryChargeStillQuotesWhatItJustProduced` could not fail — **fixed**

Correct. The stub defaulted `CurrentBalance` to `request.Amount`, so both figures were 100 and the handler's conditional made no difference to the assertion. The stub is now given 999, the test asserts 100 appears and 999 does not, and it therefore fails if the handler stops distinguishing the two cases.

---

### One of my own, while fixing the above

The first version of the new fallback test asserted `DoesNotContain(PersianFormat.Amount(0m, …))`. That renders as `۰`, which is a substring of `۱۰۰` — so it failed against entirely correct behaviour. Same class of mistake as the loose assertions caught on #172, in the opposite direction: there a bare substring passed when it should have failed, here it failed when it should have passed. It is anchored to its label now:

```csharp
Assert.Contains("اعتبار کنونی کاربر: " + PersianFormat.Amount(100m, "CREDIT_SEKE_BAHAR"), reply);
```

---

### Re-verification

```
Passed!  -  Failed: 0, Passed: 564, Skipped: 0, Total: 564
```

Build: 0 warnings, 0 errors. Simulator on the final build:

```
=== Done in 00:00:24.34. Registered 20 (20 approved), trades attempted 60, errors 0 ===
new outbox failures: 0
```

And the live sequence once more, through the running Wallet API:

```
re-send 500 -> balanceAfter 9500  currentBalance 8000  alreadyApplied True
what the wallet actually holds      : 8000
what the repeat reported as current : 8000
the stale figure it used to report  : 9500
```

Thanks for the note on the missing BOM in `WalletLazyCreationTests.cs` — agreed it is diff noise rather than a defect, and the Persian literals in it are covered by tests that still pass.

PR #177, opened 2026-08-31
[Hotfix] Admin deduction acts on the spot wallet while top-up acts on credit — make them mirrors

MohKardan:
**Branch Name**: `fix/symmetric-admin-credit-commands`
**Linked Issue**: Related to #36 (evidence recorded [there](https://github.com/MohKardan/TallaEgg/issues/36#issuecomment-5474366458))

## Description

`ش` (top-up) and `د` (deduct) wrote to **different wallets from the same Persian word**: `ش` credited `CREDIT_<X>`, `د` debited plain `<X>`. Both now act on the credit ledger, and both default to gold.

## Type of Change
- [x] Hotfix (urgent bug fix)

## The problem

```csharp
// ش  — top-up
Asset = CurrenciesConstant.CreditAssetFor(currency)   // always CREDIT_<X>

// د  — deduct
Asset = currency                                       // always plain <X>
```

The help text presented them as a matched pair, down to identical examples:

```
ش [شمارهٔ تلفن] [مقدار] [نوع]     نمونه: ش ۰۹۱۲۱۲۳۴۵۶۷ ۵۰۰۰۰۰ تومان
د [شمارهٔ تلفن] [مقدار] [نوع]     نمونه: د ۰۹۱۲۱۲۳۴۵۶۷ ۵۰۰۰۰۰ تومان
```

So an admin correcting a mistaken top-up reaches for `د <phone> <amount> <same word>`, and gets: the mistaken credit untouched, and the customer's **real** position debited instead. One mistake becomes two, under `✅ کسر از موجودی انجام شد`.

From live data on a test account:

```
CREDIT_MAUA    100.00      ← the limit the admin granted
MAUA           -39.81      ← the position that limit is backing
```

`د <phone> 100 آبشده` would have left the 100 alone and taken the spot to `-139.81`.

The two also disagreed about an omitted asset — `ش` defaulted to gold, `د` to Toman — so even the shorthand forms meant different things.

## Changes Made

- **`د` targets `CREDIT_<X>`**, like `ش`. Its deduplication key follows.
- **`د` defaults to gold**, like `ش`.
- **Both refuse a credit name.** `اعتبار آبشده` resolves to `CREDIT_MAUA`, which `ش` then prefixed again into `CREDIT_CREDIT_MAUA` — not an asset, and it failed at the wallet as a confusing "wallet not found". New `CurrenciesConstant.IsCreditAsset` makes this an explicit check.
- **The `GetPersianNamesList` comment is corrected.** It claimed both commands add the prefix themselves; that was true of `ش` only. `د` passed the resolved code through unprefixed, so it read credit names correctly while the list hid them from the admin — the only form that could reduce a credit line was the only form nobody was told about.
- **Wording follows behaviour**: the admin confirmation, the customer notification and the deduplicated-repeat message all say credit rather than balance, and both help texts now state which ledger they act on and what an omitted asset means.

## What this removes, and why it is not a loss

Deducting a **spot** balance from the bot is gone.

That is the asymmetry being removed, not a capability. `BotHandlerAdmin` is the only place the bot moves money at all, and the top-up has always meant credit — so there has never been a way to *add* to a spot balance. A deduction that could reach one had no counterpart. Spot balances move by trading and settlement, or through the wallet API directly (`POST /api/wallet/withdrawal`).

## Why now, rather than waiting for #36

#36 replaces this model entirely — credit becomes a `CreditLimit` column with `Balance >= -CreditLimit`, and none of the above can be expressed, let alone typed wrong. That is the right fix.

But #36 is blocked on an unanswered business question ("should 50g of gold credit also fund IRR purchases?"), and engineering is paused for customer discovery, so it may sit for a while. The trap is live in the meantime and costs a customer's real assets when it fires. This is deliberately the smaller, interim change; the evidence is recorded on #36 so the two stay connected.

### Unit tests
- [x] New unit tests written — 11 added
- [x] All unit tests pass

```
Passed!  -  Failed: 0, Passed: 575, Skipped: 0, Total: 575
```

564 before, 575 after.

**The suite had nothing covering which asset `د` targeted** — all 564 existing tests pass unchanged against the new behaviour. That is why the asymmetry survived: it was invisible to every test in the project.

**Verified load-bearing.** Restoring the old behaviour fails 7 of the new tests:

```
Failed  ChargingAndDeductingTheSameWordsReachTheSameWallet(asset: "")
Failed  ChargingAndDeductingTheSameWordsReachTheSameWallet(asset: "سکه")
Failed  ChargingAndDeductingTheSameWordsReachTheSameWallet(asset: "تومان")
Failed  ChargingAndDeductingTheSameWordsReachTheSameWallet(asset: "بیت‌کوین")
Failed  DeductingWithNoCurrencyGivenDefaultsToGoldCredit
Failed  TheDeductionKeyNamesTheCreditLedger
Failed  DeductingTargetsTheCreditLedgerNotTheSpotWallet
```

`ChargingAndDeductingTheSameWordsReachTheSameWallet` states the broken property directly rather than restating the implementation: charge and deduct with the same words, assert the same asset, across four inputs including the omitted one.

### Integration
- [x] Simulator

```
=== Done in 00:00:20.48. Registered 20 (18 approved), trades attempted 60, errors 0 ===
new outbox failures: 0
```

## Security Checklist
- [x] No sensitive data in logs or messages

## Code Quality
- [x] Follows `STANDARDS.md`
- [x] No new dependencies

## Breaking Changes?
- [x] **Breaking** — `د` changes meaning

`د <phone> <amount> <asset>` now reduces the customer's **credit** rather than their spot balance, and an omitted asset means gold rather than Toman. Anyone with the old muscle memory will find the command does something different.

This is intended and was decided deliberately: the previous meaning was the defect. The confirmation messages state which ledger moved, so a wrong assumption is visible in the reply rather than silent.

**Migration path**: for a genuine spot adjustment, call the wallet API directly.

## Deployment Notes

No migration, no configuration, no schema change. Revert the commit to roll back.

**Post-deployment verification**: `ش <phone> 100 سکه` then `د <phone> 100 سکه` and confirm the customer's `CREDIT_SEKE_BAHAR` returns to where it started and their `SEKE_BAHAR` is untouched.

## Related

- #36 — the model this is an interim fix for, with the field evidence
- #157 / #176 — the deduplication these commands carry

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -58,7 +58,10 @@ namespace TallaEgg.TelegramBot
                     : CurrenciesConstant.Maua; // مقدار پیش‌فرض
                 var currency = CurrenciesConstant.ResolveCurrencyCode(currencyInput);
 
-                if (currency is null)
+                // A credit name resolves, so without this "ش ... اعتبار آبشده" would be prefixed a
+                // second time into CREDIT_CREDIT_MAUA and fail at the wallet as an unknown asset.
+                // Both commands already mean the credit ledger, so naming it adds nothing.
+                if (currency is null || CurrenciesConstant.IsCreditAsset(currency))
                 {
                     await _messenger.SendAsync(message.Chat.Id,
                         string.Format(BotMsgs.MsgAdminInvalidCurrency, currencyInput, CurrenciesConstant.GetPersianNamesList()));
@@ -165,12 +168,17 @@ namespace TallaEgg.TelegramBot
                 // The input may be a Persian name or a currency code.
                 // The old default was a Persian word that never matched any asset code, so the
                 // deduction ran against a wallet that did not exist.
+                //
+                // Gold, matching the charge command. The default used to be Toman, so the two
+                // commands disagreed about what an omitted asset meant as well as about which
+                // ledger they touched.
                 var currencyInput = match.Groups["currency"].Success
                     ? match.Groups["currency"].Value
-                    : CurrenciesConstant.Toman; // مقدار پیش‌فرض
+                    : CurrenciesConstant.Maua; // مقدار پیش‌فرض
                 var currency = CurrenciesConstant.ResolveCurrencyCode(currencyInput);
 
-                if (currency is null)
+                // Same guard as the charge command: a credit name would be prefixed twice.
+                if (currency is null || CurrenciesConstant.IsCreditAsset(currency))
                 {
                     await _messenger.SendAsync(message.Chat.Id,
                         string.Format(BotMsgs.MsgAdminInvalidCurrency, currencyInput, CurrenciesConstant.GetPersianNamesList()));
@@ -180,14 +188,28 @@ namespace TallaEgg.TelegramBot
                 var userDto = await _usersApi.GetUserAsync(phone);
                 if (userDto != null)
                 {
+                    // The credit ledger, exactly as the charge command writes to it. These two
+                    // commands used to disagree: ش credited CREDIT_<X> while د debited plain <X>,
+                    // from the same Persian word and with the same help text and examples. An admin
+                    // undoing a mistaken top-up with the obvious symmetric command therefore left the
+                    // credit untouched and took the customer's real position instead — one mistake
+                    // becoming two, under a ✅ confirmation (issue #36 carries the evidence).
+                    //
+                    // The bot has never been able to add to a spot balance: this is the only place it
+                    // moves money, and the charge command has always meant credit. So a deduction that
+                    // could reach a spot balance was the asymmetry, not a capability being removed —
+                    // nothing here has a counterpart that adds. Spot balances move by trading, or
+                    // through the wallet API directly.
+                    var creditAsset = CurrenciesConstant.CreditAssetFor(currency);
+
                     var result = await _walletApi.WithdrawalAsync(new TallaEgg.Core.Requests.Wallet.WalletRequest
                     {
-                        Asset = currency,
+                        Asset = creditAsset,
                         Amount = amount,
                         UserId = userDto.Id,
 
                         // Same deduplication as the charge command above (issue #157).
-                        ReferenceId = AdminAdjustmentKey.ForWithdrawal(userDto.Id, currency, amount, DateTime.UtcNow)
+                        ReferenceId = AdminAdjustmentKey.ForWithdrawal(userDto.Id, creditAsset, amount, DateTime.UtcNow)
                     });
                     if (result.Success)
                     {
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -51,7 +51,7 @@ namespace TallaEgg.TelegramBot
             if (resolved is null)
             {
                 await _messenger.SendAsync(chatId,
-                    string.Format(BotMsgs.MsgAdminInvalidCurrency, typed, CurrenciesConstant.GetPersianNamesList()));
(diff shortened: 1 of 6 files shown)


MohKardan:

Three fixed, one deliberately left, in `82c3e37`. The first was a regression this branch introduced and the review was right to lead with it.

---

### 1. `د … تومان` now targets a non-existent asset — **fixed**, and it was worse than reported

Verified before fixing, against the running wallet:

```
CREDIT_IRT  -> REJECTED
CREDIT_MAUA -> OK
IRT         -> OK
```

Credit ledgers are minted per tradable **base** asset (the loop in `BuildCurrencies`), and Toman is a quote currency, so `CREDIT_IRT` is not a currency. Making `د` a mirror pointed it at that, and the admin got the generic `خطا در بروزرسانی`.

**`ش` has the same defect and always has.** It has always built `CREDIT_IRT` for `تومان` and always failed at the wallet — while the bot's own help offered exactly `ش ۰۹۱۲۱۲۳۴۵۶۷ ۵۰۰۰۰۰ تومان` as its example. So the branch did not create the Toman problem, only made it symmetric.

Also worth recording, since it bears on #36: `ValidateCreditAndBalanceAsync` reads `CREDIT_<quote>` and would use it if it existed (`WalletApiClient.cs:283`). Half the cross-asset credit design is unreachable rather than absent.

**Decision** (product owner): refuse Toman explicitly rather than invent `CREDIT_IRT`, because whether a quote currency should carry credit is the open question on #36, and enabling it here would pre-empt that.

So there are now three refusals with three sentences, because they send the admin to three different places:

| input | message |
|---|---|
| an unrecognised word | `نوع «…» شناسایی نشد` — check the spelling |
| a credit name (`اعتبار آبشده`) | `لازم نیست «…» را بنویسید` — the commands add it themselves |
| Toman | `«تومان» دفتر اعتبار ندارد` — real asset, no credit ledger |

`CurrenciesConstant.HasCreditLedger` answers the third by asking what exists, deciding nothing.

### 2. `MsgAdminHelp` still said "کسر از موجودی" — **fixed**

Correct, and the sharpest of the findings: it is the help an operator actually reads, it was the one message not reworded, and it is exactly the misleading-text trap this PR exists to close. It now says `کسر از اعتبار (معکوس ش)`, uses an asset that works, and states the default.

### 3. The credit-name guard reused "unrecognised" — **fixed**

Right. `اعتبار آبشده` was previously the *only* working way to reduce a credit line, and reporting it as unrecognised — against a permitted-names list that deliberately omits credit names — left the admin with nothing to correct toward. It has its own message now.

### 4. The ceiling can be cut below the position it backs — **left, deliberately**

Confirmed, and the review is right that this branch makes it the default form rather than an undocumented one.

Left alone on the product owner's call: that invariant belongs inside the wallet (`Balance >= -CreditLimit`), which is what #36 proposes, and adding a second copy of it in the bot handler is the pattern that issue exists to remove. Recorded on #36 rather than patched here.

---

### Two mistakes of my own, caught while fixing the above

Worth stating rather than quietly correcting:

**The new help text said `طلا`**, which `ResolveCurrencyCode` does not know — MAUA's Persian name is `آبشده` and it has no aliases. That is the defect this PR is about, help text teaching a command that cannot work, reintroduced inside the fix for it. The new tests caught it.

**A bulk replace of that word damaged 26 unrelated Persian strings** — `طلافروشی` became `آبشده‌فروشی`, `اطلاع` became `اآبشده‌ع`. Reverted the file and redid it as four targeted line edits; `طلافروشی` (10) and `اطلاع` (7) verified intact afterwards.

---

### Re-verification

```
Passed!  -  Failed: 0, Passed: 578, Skipped: 0, Total: 578
```

564 on `main`, 575 at the first review, 578 now. Build: 0 warnings, 0 errors.

New tests: `TomanIsRefusedBecauseItHasNoCreditLedger` (both commands) and `AnUnrecognisedWordStillGetsTheSpellingMessage`, which keeps the three refusals distinct rather than letting them collapse back into one message.

Simulator on the final build:

```
=== Done in 00:00:17.79. Registered 20 (18 approved), trades attempted 60, errors 0 ===
new outbox failures: 0
```

And the premise re-checked against the running services afterwards: `CREDIT_IRT` still rejected, which is why Toman is refused up front instead of at the wallet.

On the dropped BOM: agreed it is noise. `CurrenciesConstant.cs` is written back with one now.

PR #178, opened 2026-08-31
[Feat] Hold an out-of-band quote for admin approval instead of stopping the symbol — issue #158

### LINKED ISSUE #158: Auto-published quotes have no plausibility band: a glitched price feed becomes tradeable within one tick

MohKardan:
**Branch Name**: `feat/quote-band-approval`
**Linked Issue**: Closes #158

## Description

The plausibility band added in #172 refused an out-of-band tick and, after three in a row, deactivated the quote and switched auto-quote off. This replaces that with an approval flow: the price is held, the previous quote stays in force, and an admin is asked. It also extends the band to manually typed prices, which it never covered.

## Type of Change
- [ ] Hotfix (urgent bug fix)
- [x] Feature (new capability)
- [ ] Refactor
- [ ] Performance
- [ ] Documentation
- [ ] Security

---

## Why the stop had to go

Both problems were observed in a live session, not reasoned about.

**It made a suspicious price into a silent outage.** The symbol stopped quoting; the only thing that said so was a log line.

**And it could be walked around.** Deactivating the quote left the symbol with no anchor. An admin re-enabling auto-quote therefore hit the *cold-start* path — which by design skips the band — and published the very price that had just been refused three times. From the live log:

```
13:26:39 [WRN] rejected: 21,929,682 is 417.64% from the last mid 4,236,504
13:28:39 [WRN] rejected: 1166.60% from the last mid 1,731,381
13:30:40 [WRN] rejected
13:30:40 [ERR] stopped after 3 consecutive prices outside ±5%; 1 quote deactivated, auto-quote disabled
13:38:45 [INF] Auto-published quote for MAUA/IRT: buy 21,836,633, sell 21,946,089
```

That last line is the price the three warnings above it were about. The Error message asked for a manual quote before re-enabling; nothing enforced it.

## And why the manual path needed the band

`POST /api/quotes` applied only `Quote.Publish`'s own rules — prices above zero, buy not above sell. Nothing compared against the current quote, so a mistyped zero published straight through. That is worse than the machine case the band was built for: a manual quote is tradeable the instant it lands, and nobody is watching a log for it.

## Changes Made

**Domain (`Orders.Core`)**
- `QuotePlausibility` — the band, moved out of the publisher so both paths share one rule. Compares mid to mid, so a spread change does not read as a price move.
- `PendingQuote` — a proposal, deliberately not a `Quote`: nothing tradeable exists until somebody approves. Enforces `Quote.Publish`'s validation up front, refuses a second answer, and refuses an expired one.

**Orders**
- `PendingQuotes` table + migration `20260831111250_AddPendingQuotes`.
- `POST /api/quotes` holds an out-of-band price instead of publishing it, returning `PublishQuoteResult` with either `Published` or `Pending`.
- `GET /api/quotes/pending`, `POST /api/quotes/pending/{id}/approve`, `.../reject`.
- `AutoQuotePublisherService` holds instead of stopping, and expires stale proposals on the same tick that creates them.
- `IQuoteRepository.DeactivateActiveAsync` removed — it existed only for the stop.

**Bot**
- `PendingQuoteNotifierService` polls every 30s and asks every Admin/SuperAdmin with confirm/reject buttons.
- Callback handler; the proposal id travels in the callback data, because the message can sit for minutes, several symbols can be waiting, and any admin may answer.
- A manual quote is answered in the admin's own chat instead — they are already there.
- `PendingQuoteMessage` builds the question for both sources.

### The decisions, and where they came from

| Decision | Answer | Why |
|---|---|---|
| On breach | **Ask, never disable** | Deciding what publishes is the band's job; switching the feature off is the admin's |
| Proposal lifetime | **5 minutes** | Gold moves. A price approved half an hour later is not the market, which is its own version of the mistake being prevented |
| A newer bad price | **Supersedes** | The admin should decide about the newest price the shop has seen, not work through a backlog |
| Who is asked | **Every Admin + SuperAdmin** | Anyone who can publish a quote can judge one; first to answer decides |

### Why the bot polls rather than Orders pushing

Orders has no Telegram dependency and the bot exposes no HTTP endpoint. Polling is the direction every other bot-to-service call already runs in, and it survives a bot restart without Orders needing to retry.

### Unit tests
- [x] 37 tests added; the four describing the old stop behaviour replaced
- [x] All pass

```
Passed!  -  Failed: 0, Passed: 624, Skipped: 0, Total: 624
```

587 before, 624 after.

### Live check against the running stack

```
1) in-band manual quote  -> needsApproval=False  published
2) one zero too many     -> needsApproval=True   deviation=900%  band=5%
   quote in force is still the previous one
3) awaiting approval     -> 1 proposal, source=Manual
4) after a newer bad one -> still 1 awaiting (superseded, not queued)
5) after reject          -> previous quote still in force
6) after approve         -> the held price published
7) second approve        -> HTTP 400 "این مظنه پیش‌تر بررسی شده است."
```

Migration applied to the real SQL Server at startup; the bot's notifier started and the approval messages arrived with working buttons.

### Simulator

```
=== Done in 00:00:34.22. Registered 20 (16 approved), trades attempted 60, errors 0 ===
new outbox failures: 0
```

### Two defects the operator's testing found in this branch, both fixed

**The approval message showed gold in grams only.** An admin who typed `333,502,239` per mesghal was asked to confirm `76,989,297.52` — arithmetically correct and a number they had never seen, in a message whose entire purpose is asking whether the number is right. It now shows both units, following `MsgAdminQuotePublished`'s existing layout.

**And it labelled the price with the asset's weight unit**, so it read as `76,989,297.52 گرم` — a quantity of gold rather than toman per gram.

`PendingQuoteMessageTests` pins both, including the exact figures from that session.

### One thing checked and cleared

The operator reported the feed's gold price looked ~10x wrong. It is not: `brsapi.ir` returns `IR_GOLD_MELTED = 96,484,000` and `IR_GOLD_18K = 22,265,800` in the same response, and their ratio is **4.3333** against `GramsPerMesghal = 4.3318` — so the melted figure is per mesghal and the division lands on brsapi's own per-gram number to within 0.03%. The reported figure was the same price expressed in rial.

## Security Checklist
- [x] No hardcoded secrets, passwords, API keys or tokens
- [x] No sensitive data in logs — prices, symbols and user ids only
- [x] The approve/reject callbacks check the responder is an operator before acting: a callback is just a string, and anyone who has seen one could send it back

## Code Quality
- [x] Follows `STANDARDS.md`
- [x] No new dependencies

## Breaking Changes?
- [x] Breaking changes (see below)

`POST /api/quotes` now returns `ApiResponse<PublishQuoteResult>` instead of `ApiResponse<QuoteDto>`; the published quote moves to `data.published`. The only caller is the bot's `OrderApiClient`, updated here. A 200 no longer means the quote is live — callers must check `pending` first.

`IOrderApiClient.PublishQuoteAsync` gains a third tuple element, and `IUsersApiClient` gains `GetUsersByRoleAsync`.

## Deployment Notes

**Pre-Deployment**: one migration, `20260831111250_AddPendingQuotes`, applied automatically at Orders.Api startup. New table only — nothing existing is altered.

**Order**: deploy Orders before the bot. A bot running ahead of Orders gets 404s from the pending endpoints, which its client treats as "nothing waiting"; the reverse is harmless too, since proposals simply wait.

**Post-Deployment Verification**: publish a quote more than 5% from the current one and confirm it is held rather than published, that every admin is asked, and that approving publishes it. `PendingQuotes` should show one row per question with its outcome.

**Rollback**: revert the commit and `dotnet ef database update 20260827175732_WidenOrderAmountAndPriceTo28x8`. Held proposals are lost, which is safe: none of them was ever a published price.

## Related

- #158 is the issue this closes; #172 was the first attempt, whose stop rule this replaces
- #90 built the auto-publishing this guards

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandler.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandler.cs
@@ -467,6 +467,26 @@ namespace TallaEgg.TelegramBot
             var telegramId = callbackQuery.From?.Id ?? 0;
             var data = callbackQuery.Data ?? "";
 
+            // Answering a held quote carries the proposal id after a colon, so it cannot be matched
+            // by the constant switch below (issue #158). Handled first, and only for an admin: the
+            // buttons are only ever sent to admins, but a callback is just a string and anyone who
+            // has seen one could send it back.
+            if (data.StartsWith(InlineCallBackData.approve_quote, StringComparison.Ordinal) ||
+                data.StartsWith(InlineCallBackData.reject_quote, StringComparison.Ordinal))
+            {
+                var responder = await _usersApi.GetUserAsync(telegramId);
+
+                if (responder is null || !IsOperator(responder))
+                {
+                    await _messenger.AnswerCallbackAsync(callbackQuery.Id, BotMsgs.MsgCallbackMessageGone);
+                    return;
+                }
+
+                await _messenger.AnswerCallbackAsync(callbackQuery.Id, "");
+                await HandlePendingQuoteDecisionAsync(chatId, responder.Id, data);
+                return;
+            }
+
             switch (data)
             {
                 case InlineCallBackData.buy_spot:
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -805,10 +805,19 @@ namespace TallaEgg.TelegramBot
             // published and the price shown cannot drift apart (issue #65).
             var quote = QuoteMessage.Prepare(asset, buyPrice, sellPrice);
 
-            var (published, publishMessage) = await _orderApi.PublishQuoteAsync(
+            var (published, publishMessage, pending) = await _orderApi.PublishQuoteAsync(
                 asset, quote.BuyPricePerGram, quote.SellPricePerGram, userId);
 
-            if (published)
+            if (pending is not null)
+            {
+                // Too far from the current quote to publish unasked. The band applies to a price
+                // an admin typed as much as to one the feed produced — a mistyped zero does the
+                // same damage, and a manual quote is tradeable the moment it lands (issue #158).
+                // The admin is right here, so the question goes straight back to them.
+                var (text, keyboard) = PendingQuoteMessage.Build(pending, DateTime.UtcNow);
+                await _messenger.SendAsync(chatId, text, keyboard);
+            }
+            else if (published)
             {
                 await _messenger.SendAsync(chatId, quote.Text);
             }
@@ -817,6 +826,46 @@ namespace TallaEgg.TelegramBot
                 await _messenger.SendAsync(chatId,
                     string.Format(BotMsgs.MsgAdminQuoteFailed, publishMessage));
             }
+        }
+
+        /// <summary>
+        /// Answers a quote the plausibility band is holding (issue #158).
+        ///
+        /// <para>
+        /// The proposal id travels in the callback data rather than in conversation state,
+        /// because the message can sit in Telegram for minutes, several symbols can be waiting at
+        /// once, and any admin may be the one who answers — none of which a per-conversation
+        /// "current pending quote" would survive.
+        /// </para>
+        ///
+        /// <para>
+        /// "Already answered" and "expired" both come back as refusals with a sentence written for
+        /// the admin, and both are ordinary: two admins can press at once, and a button in an old
+        /// message is exactly what the expiry window exists to stop.
+        /// </para>
+        /// </summary>
+        private async Task HandlePendingQuoteDecisionAsync(long chatId, Guid adminUserId, string callbackData)
+        {
+            var separator = callbackData.IndexOf(':');
+            if (separator < 0 || !Guid.TryParse(callbackData[(separator + 1)..], out var pendingQuoteId))
+            {
+                _logger.LogWarning("Unparseable pending-quote callback: {CallbackData}", callbackData);
+                return;
+            }
+
+            var approving = callbackData.StartsWith(InlineCallBackData.approve_quote, StringComparison.Ordinal);
+
+            var (success, message) = approving
+                ? await _orderApi.ApprovePendingQuoteAsync(pendingQuoteId, adminUserId)
+                : await _orderApi.RejectPendingQuoteAsync(pendingQuoteId, adminUserId);
+
+            if (!success)
+            {
+                await _messenger.SendAsync(chatId, string.Format(BotMsgs.MsgAdminQuoteResolveFailed, message));
+                return;
+            }
+
+            await _messenger.SendAsync(chatId, message);
         }
 
         /// <summary>
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/BotHandlerAdmin.cs
@@ -855,6 +855,12 @@ namespace TallaEgg.TelegramBot
 
             var approving = callbackData.StartsWith(InlineCallBackData.approve_quote, StringComparison.Ordinal);
 
+            // Read before answering: once resolved it is no longer in the awaiting list, and the
+            // confirmation needs its symbol and prices. A null here is not fatal — the server's own
+            // sentence is still shown — so a failed read costs detail, not the answer.
+            var pending = (await _orderApi.GetPendingQuotesAsync())
+                ?.FirstOrDefault(p => p.Id == pendingQuoteId);
+
             var (success, message) = approving
                 ? await _orderApi.ApprovePendingQuoteAsync(pendingQuoteId, adminUserId)
                 : await _orderApi.RejectPendingQuoteAsync(pendingQuoteId, adminUserId);
@@ -865,7 +871,16 @@ namespace TallaEgg.TelegramBot
                 return;
             }
 
-            await _messenger.SendAsync(chatId, message);
+            // The server's sentence says what happened but not to which symbol or at what price,
+            // and an admin may have several questions open at once. The confirmation is rebuilt
+            // from the proposal so it names both — in the same units the question used.
+            var answered = pending is null
+                ? message
+                : approving
+                    ? PendingQuoteMessage.Approved(pending)
+                    : PendingQuoteMessage.Rejected(pending);
+
+            await _messenger.SendAsync(chatId, answered);
         }
 
         /// <summary>
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/IOrderApiClient.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/IOrderApiClient.cs
@@ -57,8 +57,11 @@ public interface IOrderApiClient
     // ── Automatic quotes (issue #90) ────────────────────────────────────────────
 
     /// <summary>A symbol's current automatic-quote settings.</summary>
-    /// <summary>Quotes the plausibility band is holding until an admin answers (issue #158).</summary>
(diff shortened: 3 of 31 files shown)


MohKardan:
## Review response

All eight findings are addressed in 15ed991. Six are fixed outright; two are answered rather than
fixed, and I have said plainly below which is which and why.

### Fixed

**1 — An approval or rejection produced no confirmation naming what was decided.**
The bot replied with the server's own sentence, which identifies no symbol and no price. Several
proposals can be open at once, so an admin holding two messages could not tell which one they had
just answered. `PendingQuoteMessage.Approved`/`Rejected` now build a confirmation naming the asset
and, for gold, the prices in both mesghal and gram — the same shape as the question, so the reply
is checkable against the message that prompted it.

**2 — The notifier re-asked every admin every tick.**
It keyed "have I asked about this?" on the proposal id, but a feed stuck on a wrong price
supersedes its proposal and mints a new id every two minutes, so the key never repeated and every
admin was messaged all night. It is now keyed on the symbol, with a fifteen-minute quiet period —
longer than the proposal's own five minutes, because re-asking the moment one expires restores the
flood a proposal at a time. Nothing is lost by waiting: the button carries whichever proposal is
live when it is pressed, so an admin answering at minute fourteen is answering about the newest
price, not the one that prompted the message.

**3 — A failed poll was read as "nothing is waiting".**
`GetPendingQuotesAsync` returns null when Orders cannot be reached, which the loop treated like an
empty list — clearing the record of what had been asked and re-broadcasting every open proposal to
every admin on the next poll. One network blip became a broadcast. It now returns early on null.

**4 — Two admins pressing at the same instant could both publish.**
The in-memory `Status == Pending` check does not survive two threads that both read Pending before
either wrote. `PendingQuote` now carries a `Version` concurrency token, bumped by `Approve`,
`Reject`, `Supersede` and `Expire`, so the second writer loses in the database. The repository
catches `DbUpdateConcurrencyException` and raises the sentence a person should see
("این مظنه پیش‌تر بررسی شده است.") rather than an EF exception. The migration was regenerated to
include the column — the first one predated the token.

**5 — No test covered the token.**
`EveryStateChangeBumpsTheConcurrencyToken` and `AFreshProposalStartsAtVersionZero` in
`PendingQuoteTests`. The first pins that all four transitions bump it, not just approval; a
`Supersede` that forgot to would let a superseded proposal be approved by a stale reader.

**6 — A second operator lookup was added where one existed.**
The notifier now calls the existing `IUsersApiClient.GetOperatorTelegramIdsAsync()`.

### Answered, not fixed

**7 — A persistently out-of-band feed leaves the shop trading on a stale quote.**
Correct, and deliberate. This is the product owner's decision on #158, given in these words: if a
quote falls outside the band it must not be written to the database and must not disable
auto-quoting — an admin is asked instead. The previous behaviour (stop the symbol after three
rejections, from #172) is what this PR replaces, at their instruction. So the consequence stands:
if the feed is wrong for an hour and no admin answers, customers trade for an hour on the last
approved price.

What this PR does about it is make the situation loud rather than silent. A held proposal with no
reachable admin is logged at Error, not Warning, because a shop with a held quote and nobody to ask
is stuck on a price it cannot move. Anything further — a hard age limit on the standing quote, or
suspending trading rather than quoting — changes what customers can do and is not mine to decide
inside a fix. Worth raising with the owner separately.

**8 — Manual quotes at a different scale from the standing quote are all held.**
Also correct, and the intended effect of extending the band to the manual path. It showed up in the
simulator: all 120 of its Phase 6 quotes were held, because the quote standing in my local database
(≈33.4M/gram, left by an earlier approval test) is ~87% away from the ≈4.26M/gram scale the
simulator publishes at. The simulator's own successive prices are 2.43% apart on average — inside
the band — so against a database whose standing quote matches its scale it publishes normally. It
is the mismatch with the pre-existing quote that trips it, which is the band working.

Recording it because the next person to run the simulator against a realistically-priced database
will see Phase 6 stop publishing and reasonably think something broke.

### Verification

- `dotnet build TallaEgg.sln` — 0 warnings, 0 errors (`TreatWarningsAsErrors` is on)
- `dotnet test TallaEgg.sln` — 626 passed, 0 failed
- Live run against the three services with the regenerated migration applied: in-band manual quote
  publishes; a quote with one zero too many is held at 900% against a 5% band; a newer bad proposal
  supersedes rather than queues; rejection leaves the previous quote in force; approval publishes;
  a second approval is refused.
- Simulator: 993 trades, 7 errors, all of them the expected "موجودی یا اعتبار شما برای این معامله
  کافی نیست" from users who ran out. Outbox drained to zero with no new failures. 0 duplicate
  settlements, 0 colliding `(WalletId, ReferenceId)`, every settled reference has exactly 4 legs,
  0 transaction-chain breaks, 0 wallet/last-transaction balance mismatches, 0 negative locked
  balances, and 0 orphan orders created during the run.

Issue #105, opened 2026-08-09
[Deployment] Production launch checklist: secrets, CORS, and hardening


MohKardan:
## Summary

A roundup of small but easy-to-forget items that came up while closing #33 and #31. Each one is
a real gap the moment this goes live and public. Filing now so they don't get lost between
conversations.

> **Updated 2026-08-28.** Re-checked against the code and the repository settings rather than
> against this list, and most of it was already done. #69 and #70 are closed. Secret scanning and
> push protection are enabled — verified through the GitHub API, not assumed. What remains is two
> items, and **neither is engineering work**:
>
> - the API key needs somewhere to be set, and
> - CORS needs a domain to allow.
>
> **No server has been provisioned.** That is the whole of what holds this issue — and, through
> it, the audit's Operational readiness row at 4/10. Everything the code can do is done: addresses
> come from configuration, the services install as Windows Services, authentication and the CORS
> whitelist are wired, and there is global error handling with trace ids. What is missing is a
> machine. Until there is one, this row cannot move, and no amount of code will move it.

> **Updated 2026-09-03.** A server now exists, and all four services are installed and running on
> it — so the paragraph above is out of date and is kept only as the record of where this stood.
> Every item was re-tested against the deployment; the evidence is in the comment of 2026-09-03.
> **Four of the five items are done**, including the one that was holding this issue: the
> Production code path has now executed, and an unauthenticated request is refused with 401. The
> fifth — `Cors:AllowedOrigins` — was never blocked on a server; it was blocked on #97, and it has
> been carried there. **This issue is closed on that basis.** What the deployment surfaced that this
> list never had is enumerated in the comment, unfiled: redeploying after a merge, database backups,
> cold-boot ordering against SQL Server, and log collection.

## Checklist

- [x] ~~**Generate the real `TALLAEGG_API_KEY` value and set it in the production environment.**~~
      **Done 2026-09-03** — set for all four services on the deployed machine, and confirmed by the
      APIs refusing an unauthenticated request with 401 rather than 500. Evidence in the comment of
      2026-09-03. **The rationale below is wrong in one respect and is left here only as the
      record**: the guard does *not* throw at startup — it runs on the first request (#214), so a
      service coming up is not by itself evidence that the key is set.
      *Originally: blocked on a server existing.* Deferred deliberately in #33:
      `APIKeyConstant.RequireTallaEggApiKey()` throws at startup in Production if it is unset, so
[...]

PR #180, opened 2026-08-31
[Hotfix] The shared config file overrode environment variables and the command line — issue #159

### LINKED ISSUE #159: The shared config file overrides environment variables, so no setting can be varied per host
From `docs/audit/AUDIT_2026-08b.md` §10 (2026-08-29). Read from registration order, not run.

## What happens

Every service calls `builder.Configuration.AddJsonFile(sharedConfigPath, ...)` **after** `WebApplication.CreateBuilder` has already registered the default providers. In ASP.NET Core, last registration wins — so the shared file takes precedence over **environment variables and command-line arguments**, which is the opposite of what everyone assumes.

Example: `src/Wallet/Wallet.Api/Program.cs:31`. The same shape is in every service.

## Why it matters

Ports, connection strings and URLs cannot be varied per host. To run a second instance, or to point staging at a different database, you must edit `config/appsettings.global.json` itself — the one file that holds live credentials and is deliberately untracked (#33).

That makes the ordinary act of deploying require editing the most sensitive file in the system, by hand, on the server. It is exactly the kind of friction that ends with someone copying a file they should not copy.

Nothing is broken today because everything runs on one machine. This becomes real the moment a server exists.

## Suggested fix

Register the shared file *before* the environment-variable and command-line providers, or re-add those two after it:

```csharp
builder.Configuration.AddJsonFile(sharedConfigPath, optional: false, reloadOnChange: true);
builder.Configuration.AddEnvironmentVariables();
builder.Configuration.AddCommandLine(args);
```

Then the file stays the source of truth for everything shared, and a host can override the handful of things that are genuinely per-host without touching it.

## Acceptance

- Setting a connection string or port through an environment variable actually takes effect
- The shared file still supplies every value not overridden
- Applied consistently across all services and the bot, not just one

## Related

#105 — this belongs on the production checklist; the first deployment is when it stops being theoretical.

MohKardan:
**Branch Name**: `fix/config-precedence-159`
**Linked Issue/Task**: Closes #159

---

## Description

Every API host registered `config/appsettings.global.json` — and then the
`Services:{ApplicationName}` section flattened out of it — *after*
`WebApplication.CreateBuilder` had already registered the environment-variable and
command-line providers. Configuration providers are last-wins, so the shared file
outranked both, and no setting could be varied per host without hand-editing the one
file that holds live credentials and is deliberately untracked (#33).

This re-registers `AddEnvironmentVariables()` and `AddCommandLine(args)` immediately
after `AddInMemoryCollection(flattened)`, which is the order the bot and the simulator
already use. The file stays the source of truth for everything a host does not
explicitly override.

---

## Type of Change
- [x] Hotfix (urgent bug fix)

Marked hotfix only in the sense that it is a defect fix; nothing is on fire, because
everything still runs on one machine. It stops being theoretical the first time a
server exists.

---

## Verified before writing, not assumed

The claim was checked against all seven hosts rather than taken from the audit text. It
held exactly:

| Host | Before |
|---|---|
| `src/Order/Orders.Api/Program.cs` | file + flattened section registered last |
| `src/User/Users.Api/Program.cs` | file + flattened section registered last |
| `src/Wallet/Wallet.Api/Program.cs` | file + flattened section registered last |
| `src/TallaEgg/TallaEgg.Api/Program.cs` | file + flattened section registered last |
| `src/Affiliate/Affiliate.Api/Program.cs` | file + flattened section registered last |
| `TelegramBot/…Infrastructure/Program.cs` | already calls `AddEnvironmentVariables()` after both — correct |
| `TelegramBot/…Simulator/Program.cs` | already calls `AddEnvironmentVariables()` after both — correct |

One detail the suggested fix in the issue does not spell out, and which decides where
the two lines go: it is not enough to register the environment after `AddJsonFile`. The
flattening step copies the file's values onto the root-level keys the services actually
read (`ConnectionStrings:WalletDb`, `OrdersApiUrl`, …) through a second provider, so an
override registered between the file and the flattening would still be overridden. Both
new calls therefore go after `AddInMemoryCollection`, exactly as the bot has them.

---

## Changes Made

- `src/{Order/Orders.Api, User/Users.Api, Wallet/Wallet.Api, TallaEgg/TallaEgg.Api, Affiliate/Affiliate.Api}/Program.cs`:
  re-register `AddEnvironmentVariables()` and `AddCommandLine(args)` after the shared
  file and its flattened section.
- `tests/TallaEgg.AllServices.Tests/ConfigurationPrecedenceTests.cs`: new, 21 tests.

No change to the bot or the simulator — see below.

---

## Decisions this change had to make

**`AddCommandLine(args)` as well as environment variables — yes.** Checked what the
hosts read from `args` first: the five APIs read nothing from it themselves, they only
hand it to `CreateBuilder(args)`, which already registers a command-line provider (when
`args` is non-empty) among the defaults. So this adds no new parsing of anything that
was not already being parsed — it cannot introduce an argument-format failure that did
not already exist — and it changes behaviour only for someone passing a switch who
already expected it to win. Confirmed by running `Wallet.Api` with
`--Services:Wallet.Api:Urls:0=http://localhost:61934`: it bound to 61934.

**`Affiliate.Api` included, although it is outside the audit's scope.** The change is
mechanical and it had the same defect. Leaving it out would leave one service running
the opposite precedence rule from the other four — the specific inconsistency the third
acceptance criterion warns about — and it would have to be found again later. It boots
clean with the change (its EF migrations run as before).

**The bot and the simulator were left alone.** Both already register the environment
after the file and its flattened section, so the defect this issue is about is not
present in either. The bot does have a smaller related gap: `Host.CreateDefaultBuilder(args)`
registers a command-line provider, but early enough that the shared file still outranks
it. Nothing in the bot reads a setting from `args` today, so it is a latent
inconsistency rather than a live problem, and per `STANDARDS.md` on scope discipline it
is flagged here rather than fixed inline. Say the word and it is a one-line follow-up.

**`ConfigurationGuard` was checked, not bypassed.** `RequireConnectionString` reads the
*merged* configuration, so an environment override reaches it like any other value: a
real override satisfies the guard, and an override set to blank still fails startup as
it should. The `Services:{ApplicationName}` existence check is unaffected — it runs
against the file, which is still `optional: false`.

---

## Observation, not a change

`ResolveSharedConfigPath` is copy-pasted verbatim into all seven `Program.cs` files, and
the section-flattening block into six of them. That duplication is why this bug needed
fixing in five places instead of one, and it is what will make the next configuration
change cost five edits again. A shared bootstrap helper would be the real answer.
Deliberately not done here — this issue is about provider order, and widening it into a
startup refactor across every service is exactly what `STANDARDS.md` says not to do.
Worth its own issue.

---

### Unit Tests
- [x] New unit tests written

```
dotnet build TallaEgg.sln --configuration Release
-> Build succeeded. 0 Warning(s) 0 Error(s)

dotnet test TallaEgg.sln --configuration Release --no-build
-> Passed! Failed: 0, Passed: 668, Skipped: 0, Total: 668
```

`ConfigurationPrecedenceTests` is in two halves on purpose, because either half alone
would be worth little:

- **Behavioural** (4 tests) — builds the same provider chain the hosts build, against a
  throwaway JSON file shaped like the shared one, and asserts that an environment
  variable beats the file, that a command-line switch does, that a port set through the
  environment reaches `Services:{App}:Urls`, and that an un-overridden key still comes
  from the file. A negative control reproduces the old order and asserts the file wins,
  so the other three cannot pass for a reason unrelated to precedence.
- **Source guard** (17 theory cases) — reads all seven host `Program.cs` files and fails
  if `AddEnvironmentVariables` (all seven) or `AddCommandLine` (the five APIs) is ever
  registered before the shared file or before the flattening. Comment lines are stripped
  first, so a comment mentioning a provider can neither satisfy nor defeat it. A third
  case (added in decbb19, from the review) covers the *second* ordering the port override
  rests on: the hosts capture `serviceSection` early and read `Urls` off it late, and that
  only works because a section reads through to its root on every access, so `UseUrls`
  must stay below `AddEnvironmentVariables`.

The guard was confirmed non-vacuous: reverting `Wallet.Api/Program.cs` alone turns
exactly its two theory cases red.

```
Failed …EveryApiHost_RegistersTheCommandLine_AfterTheSharedFile(hostProgram: "src/Wallet/Wallet.Api/Program.cs")
Failed …EveryHost_RegistersTheEnvironment_AfterTheSharedFile(hostProgram: "src/Wallet/Wallet.Api/Program.cs")
```

### The risk here is boot, not tests

A wrong change to configuration composition breaks every service at startup and no unit
test would show it, so the whole stack was actually run.

**All five services start, no errors, bound to the ports from the file:**

```
[INF] Now listening on: http://localhost:5136    (Users.Api)
[INF] Now listening on: http://localhost:60933   (Wallet.Api)
[INF] Now listening on: http://localhost:5140    (Orders.Api)
[INF] Now listening on: http://localhost:5135    (TallaEgg.Api)
[INF] Now listening on: http://localhost:60812   (Affiliate.Api)
-> 0 [ERR]/[FTL] lines across all five logs
```

**The bot starts and reaches its polling loop:**

```
[INF] Bot connection successful: @…
[INF] Bot is now running and listening for messages...
[INF] PendingQuoteNotifierService started (poll every 30s).
```

**End to end through the real handlers and APIs** (`driver.ps1 smoke`, 5 users /
5 quotes / 10 trades):

```
=== Done in 00:00:29.8. Registered 5 (4 approved), trades attempted 10, errors 0 ===
```

**A real override, applied by hand, at boot** — the acceptance criterion of the issue:

| What was set | Result |
|---|---|
| `Services__Wallet.Api__Urls__0=http://localhost:61933` | `Now listening on: http://localhost:61933`; port 60933 (the file's) never opened |
| `--Services:Wallet.Api:Urls:0=http://localhost:61934` on the command line | `Now listening on: http://localhost:61934`; port 60933 never opened |
| `ConnectionStrings__WalletDb` pointed at a hostname that exists nowhere | startup failed inside `Migrator.MigrateAsync` with "server was not found" — EF used the environment's value, not the file's, which the same service had started against minutes earlier |

Before this change the environment variable would simply have been ignored and the
service would have come up on 60933 against the file's database.

### Environment
- [x] Tested locally on Windows
- [ ] Tested on staging environment (no staging exists yet — see `STANDARDS.md` §7)

---

## Security Checklist

- [x] `.gitignore` blocks sensitive files (config, .env, secrets)
- [x] Code compiles without warnings (`TreatWarningsAsErrors` is on; 0 warnings)

`config/appsettings.global.json` is untouched and uncommitted, and no value from it
appears in this PR, the tests, or the commit message. The test's throwaway config file
is generated per-test in the temp directory with invented values and deleted afterwards
— CI has no shared config file and this suite still must not need one.

Worth noting the direction of travel: the point of this change is that the next
deployment does **not** have to edit that file to move a port or a database, which is
the friction that ends with someone copying it somewhere they should not.

---

## Code Quality

- [x] Consistent indentation and formatting

`FindRepositoryRoot` is duplicated from `SolutionMembershipTests` rather than extracted,
because extracting it would mean editing a test this change has no business touching.
Noted in the test's own XML docs.

---

## Documentation

- [ ] Updated relevant `.md` files in `docs/` — none needed. `AGENT.md`'s Configuration
      section describes the shared file as the source of truth, which is still true. It
      is now the source of truth *for what a host does not override*, which is the
      normal ASP.NET Core expectation and not something that document asserted otherwise.

---

## Breaking Changes?

- [x] No breaking changes

Nothing changes for any host that sets no environment variable or command-line switch,
which is every host today. Verified by running the whole stack: same ports, same
databases, same simulator result.

The one thing to be aware of: a *pre-existing* environment variable that happens to
collide with a key in the shared file now wins where it previously lost. That is the
intended behaviour, and there is nothing of the sort on this machine or in CI, but it is
the shape of surprise this change makes possible.

---

## Deployment Notes

**Pre-Deployment**: nothing. No migrations, no new environment variables, no feature flag.

**Post-Deployment Verification**:
- [x] Application starts successfully
- [x] Smoke tests pass
- [x] Logs show no errors

**How to use it** (the point of the issue). Any key a service reads at the root — connection
strings, `*ApiUrl`, `AutoQuote:*` — can be set either way, because the flattening runs over a
configuration that already contains the default environment-variable provider:

```
ConnectionStrings__WalletDb=Server=…;Database=…;
Services__Wallet.Api__ConnectionStrings__WalletDb=Server=…;Database=…;   # equivalent
```

A port must use the nested path, because the hosts read `Services:{ApplicationName}:Urls` off
the section rather than through the flattened copy:

```
Services__Wallet.Api__Urls__0=http://0.0.0.0:60933
```

`ASPNETCORE_URLS` and `--urls` remain inert: `UseUrls` is fed unconditionally from the file and
writes through `UseSetting`, which bypasses the provider chain entirely. That is a different
mechanism from the one this issue describes and it is not addressed here — see the review thread
on this PR for the evidence and the one-line fix it would take.

**Rollback Plan**: revert the commit and restart. There is no state to undo — the change
is two lines of provider registration per service.

---

## Related Issues

- Closes #159 (audit finding, `docs/audit/AUDIT_2026-08b.md` §10)
- #105 — this belongs on the production checklist; the first deployment is when it stops
  being theoretical
- #33 — why `config/appsettings.global.json` is untracked, and why editing it on a server
  is the thing worth avoiding

### DIFF
--- a/src/Affiliate/Affiliate.Api/Program.cs
+++ b/src/Affiliate/Affiliate.Api/Program.cs
@@ -38,6 +38,15 @@ var flattened = serviceSection.AsEnumerable(true)
 
 builder.Configuration.AddInMemoryCollection(flattened);
 
+// Re-registered after the shared file and the section flattened from it so that last-wins
+// puts a host on top of both. WebApplication.CreateBuilder registers these two, but ahead of
+// the AddJsonFile above, which left the file outranking them: no port, URL or connection
+// string could be varied per host without hand-editing config/appsettings.global.json, the
+// one file that holds live credentials and is deliberately untracked (#33). The file stays
+// the source of truth for every value a host does not explicitly override (issue #159).
+builder.Configuration.AddEnvironmentVariables();
+builder.Configuration.AddCommandLine(args);
+
 // نمادهای معاملاتی از appsettings.global.json (بخش Symbols) — نه پیش‌فرض‌های کامپایل‌شده.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
--- a/src/Order/Orders.Api/Program.cs
+++ b/src/Order/Orders.Api/Program.cs
@@ -54,6 +54,15 @@ var flattened = serviceSection.AsEnumerable(true)
 
 builder.Configuration.AddInMemoryCollection(flattened);
 
+// Re-registered after the shared file and the section flattened from it so that last-wins
+// puts a host on top of both. WebApplication.CreateBuilder registers these two, but ahead of
+// the AddJsonFile above, which left the file outranking them: no port, URL or connection
+// string could be varied per host without hand-editing config/appsettings.global.json, the
+// one file that holds live credentials and is deliberately untracked (#33). The file stays
+// the source of truth for every value a host does not explicitly override (issue #159).
+builder.Configuration.AddEnvironmentVariables();
+builder.Configuration.AddCommandLine(args);
+
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
--- a/src/TallaEgg/TallaEgg.Api/Program.cs
+++ b/src/TallaEgg/TallaEgg.Api/Program.cs
@@ -42,6 +42,15 @@ var flattened = serviceSection.AsEnumerable(true)
 
 builder.Configuration.AddInMemoryCollection(flattened);
 
+// Re-registered after the shared file and the section flattened from it so that last-wins
+// puts a host on top of both. WebApplication.CreateBuilder registers these two, but ahead of
+// the AddJsonFile above, which left the file outranking them: no port, URL or connection
+// string could be varied per host without hand-editing config/appsettings.global.json, the
+// one file that holds live credentials and is deliberately untracked (#33). The file stays
+// the source of truth for every value a host does not explicitly override (issue #159).
+builder.Configuration.AddEnvironmentVariables();
+builder.Configuration.AddCommandLine(args);
+
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
--- a/src/User/Users.Api/Program.cs
+++ b/src/User/Users.Api/Program.cs
@@ -53,6 +53,15 @@ var flattened = serviceSection.AsEnumerable(true)
 
 builder.Configuration.AddInMemoryCollection(flattened);
 
+// Re-registered after the shared file and the section flattened from it so that last-wins
+// puts a host on top of both. WebApplication.CreateBuilder registers these two, but ahead of
+// the AddJsonFile above, which left the file outranking them: no port, URL or connection
+// string could be varied per host without hand-editing config/appsettings.global.json, the
+// one file that holds live credentials and is deliberately untracked (#33). The file stays
+// the source of truth for every value a host does not explicitly override (issue #159).
+builder.Configuration.AddEnvironmentVariables();
+builder.Configuration.AddCommandLine(args);
+
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
--- a/src/Wallet/Wallet.Api/Program.cs
+++ b/src/Wallet/Wallet.Api/Program.cs
@@ -49,6 +49,15 @@ var flattened = serviceSection.AsEnumerable(true)
 
 builder.Configuration.AddInMemoryCollection(flattened);
 
+// Re-registered after the shared file and the section flattened from it so that last-wins
+// puts a host on top of both. WebApplication.CreateBuilder registers these two, but ahead of
+// the AddJsonFile above, which left the file outranking them: no port, URL or connection
+// string could be varied per host without hand-editing config/appsettings.global.json, the
+// one file that holds live credentials and is deliberately untracked (#33). The file stays
+// the source of truth for every value a host does not explicitly override (issue #159).
+builder.Configuration.AddEnvironmentVariables();
+builder.Configuration.AddCommandLine(args);
+
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
--- a/tests/TallaEgg.AllServices.Tests/ConfigurationPrecedenceTests.cs
+++ b/tests/TallaEgg.AllServices.Tests/ConfigurationPrecedenceTests.cs
@@ -167,6 +167,24 @@ public class ConfigurationPrecedenceTests
     }
 
     /// <summary>
+    /// The port override rests on a second ordering that no behavioural test can reach: the hosts
+    /// capture <c>serviceSection</c> early but read <c>Urls</c> off it late, and a
+    /// <see cref="IConfigurationSection"/> reads through to its root on every access, so the read
+    /// picks up providers registered after the capture. Move the <c>UseUrls</c> block up beside
+    /// the flattening it reads from and the port override dies silently — with every other test
+    /// in this file still green.
+    /// </summary>
+    [Theory]
+    [MemberData(nameof(ApiHosts))]
+    public void EveryApiHost_ReadsItsUrls_AfterTheEnvironmentIsRegistered(string hostProgram)
+    {
+        var code = ReadHostProgram(hostProgram);
+
+        AssertRegisteredLast(code, hostProgram, "GetSection(\"Urls\")", "AddEnvironmentVariables");
+        AssertRegisteredLast(code, hostProgram, "UseUrls", "AddEnvironmentVariables");
+    }
+
+    /// <summary>
     /// Builds configuration the way the five API hosts do, in their order: the shared file, then
     /// the service's own section flattened into the root, then the host's own overrides.
     /// </summary>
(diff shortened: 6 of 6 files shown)


MohKardan:
## Response to `/code-review high` on this PR

Three findings. One is refuted by experiment, one is confirmed and is not fixed here, one is
fixed in decbb19.

---

### 1. "Nested `Services:{App}:…` overrides are silently ignored" — **refuted**

The reasoning was that `flattened` is a `ToDictionary` snapshot taken before the new providers
exist, so a nested environment variable cannot reach the root key the code reads. The snapshot
part is true; the conclusion is not, and the missing step is one line above the flattening:

```csharp
var builder = WebApplication.CreateBuilder(args);   // already registers AddEnvironmentVariables()
…
var serviceSection = builder.Configuration.GetSection($"Services:{applicationName}");
```

The section is enumerated off a `ConfigurationManager` that **already contains the default
unprefixed environment-variable provider**. A nested variable is therefore visible inside the
section at flattening time and is flattened onto the root key along with the file's values. This
predates the PR — it is how the flattening has always behaved.

Tested rather than argued, on `Wallet.Api`, whose connection strings live at the *root* of the
shared file and not under its section (so the flattening is the only possible route):

| Environment variable | Result |
|---|---|
| *(none — control)* | starts normally on 60933 |
| `Services__Wallet.Api__ConnectionStrings__WalletDb` = an unreachable host | **fails at `Migrator.MigrateAsync`, "server was not found"** — the nested override reached EF |
| `Services__Orders.Api__ConnectionStrings__WalletDb` = the same unreachable host | starts normally — the wrong application name is correctly ignored |

The third row is the discriminator: the application-name prefix is what decides, which is only
possible if the flattening is what carried the value. Reproduced twice.

So both spellings work for a flat key — `AutoQuote__BrsApiKey` and
`Services__Orders.Api__AutoQuote__BrsApiKey` alike — and the guidance in
`BrsApiPriceProvider.cs:44` is not misleading. I have added the flat form to the deployment notes
anyway, since it is the shorter one.

---

### 2. "`ASPNETCORE_URLS` / `--urls` are still overridden" — **confirmed, and deliberately not fixed here**

Correct, and worth having found. Verified:

```
ASPNETCORE_URLS=http://localhost:61935  ->  [INF] Now listening on: http://localhost:60933
                                            port 61935 never opened
```

`builder.WebHost.UseUrls(urls)` is fed unconditionally from the file's section, and `UseUrls`
writes through `UseSetting`, which outranks the provider chain this PR just fixed. So the
standard spelling of "put this instance on another port" is still inert; only
`Services__{App}__Urls__0` works, which is what the deployment notes document and what was
verified end to end.

Not fixed in this PR, for two reasons. It is a different mechanism from the one #159 describes —
provider registration order versus a host builder call that bypasses providers entirely — and the
fix is a behavioural change to how all five services choose their listen address, which is the
boot path where a mistake takes the whole platform down. That is not something to slip into a
mechanical reordering. The issue's acceptance criterion, a port set through an environment
variable actually taking effect, is met.

The fix, when it is done, is small and should be its own change:

```csharp
// A host that set ASPNETCORE_URLS or --urls has already said where to listen; the file's
// value must not overwrite it.
if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKey]) && urls is { Length: > 0 })
{
    builder.WebHost.UseUrls(urls);
}
```

Raising it with the owner before either opening a follow-up issue or extending this PR.

---

### 3. "The port test cannot catch the property the hosts depend on" — **valid, fixed in decbb19**

Also correct. The hosts capture `serviceSection` at line 33 and read `Urls` off it at line 65,
and that works only because a `ConfigurationSection` reads through to its root on every access,
so the late read sees providers registered after the capture. Nothing pinned that: moving the
`UseUrls` block up beside the flattening block it reads from would have killed the port override
with all 16 tests still green.

Added `EveryApiHost_ReadsItsUrls_AfterTheEnvironmentIsRegistered`, a source guard over the five
API hosts asserting that both `GetSection("Urls")` and `UseUrls` come after
`AddEnvironmentVariables`. Confirmed non-vacuous by actually performing that move:

```
Failed …EveryApiHost_ReadsItsUrls_AfterTheEnvironmentIsRegistered(hostProgram: "src/Wallet/Wallet.Api/Program.cs")
```

Suite is now 668 tests, 0 failures, Release, 0 warnings.

Issue #184, opened 2026-09-01
DataReset wipes the participants but leaves their settlements queued, so each simulator run orphans the last few


MohKardan:
Found while fixing #174 (2026-09-01). It cost most of that investigation, and it will cost the next
one the same way, which is why it is worth its own issue.

## What happens

`DataReset` — the simulator's Phase 0 — deletes everything a previous run left behind, scoped to
`TelegramId < 0`: `Transactions`, `Wallets`, `Trades`, `Orders`, `Users`. It does **not** delete
`OutboxMessages`.

Settlement is queued, not immediate: the outbox polls every 5s and claims 20 at a time, while a
60-trade run finishes in about 13 seconds. A run therefore almost always ends with settlements still
queued. The next run's reset then removes the wallets those settlements are for, and they can never
succeed:

```
Wallet settlement rejected: Buyer locked IRT (0) is less than required (42148277).
```

They retry five times, are marked `Failed`, and stay there. Ten consecutive runs on this machine left
609 of them.

## Why it matters

**1. It silently removes the contention a simulator run exists to exercise.**

`ClaimDueMessagesAsync` orders by `CreatedAt`, oldest first. Once a backlog of dead messages exists,
each poll spends its batch on those instead of the current run's, so the run's own settlements are
starved. Settlement stops happening *during* the run — and settlement is the background writer that
contends with quote fills for the market maker's wallet row.

The result is a run that looks healthy and proves nothing. Concretely, while diagnosing #174:

| | runs | lock collisions observed | fills refused |
|---|---|---|---|
| with a dirty queue | 10 | 0 | 0 |
| with the queue drained between runs | 5 | 11 | **2** |

Ten green runs were recorded before the cause was found. The bug was there the whole time; the dirty
queue was hiding it. Any future concurrency or settlement work hits exactly this.

**2. It strands collateral on the market maker.**

Both sides of a trade lock collateral, and only settlement consumes it. When settlement can never
run, the lock is permanent. The customer's lock disappears with their deleted wallet row, but the
market maker is a real account (`TelegramId 6389449308`, positive by design since the filter was
corrected) so it is never wiped, and its locks accumulate across every run.

The arithmetic matches exactly, which confirms the mechanism and rules out any fault in
lock/unlock/consume:

| | locked on the market maker's wallet | sum over failed settlements |
|---|---|---|
| MAUA | `446.98` | `446.98` — 298 settlements where it was the seller |
[...]

PR #187, opened 2026-09-01
[Fix] ASPNETCORE_URLS, --urls and the bot's command line now reach the host — issue #181

### LINKED ISSUE #181: ASPNETCORE_URLS is still ignored, and the bot's command line is still outranked by the shared file
Follow-up to #159 / PR #180, which fixed provider registration order in the five APIs. Two
places still let the shared file beat a host, both found while verifying that change. Neither is
a live problem today — everything runs on one machine — and both become real at the first
deployment, same as #159.

## 1. `ASPNETCORE_URLS` and `--urls` are still ignored (five APIs)

Every API host does this:

```csharp
var urls = serviceSection.GetSection("Urls").Get<string[]>();
if (urls is { Length: > 0 })
{
    builder.WebHost.UseUrls(urls);
}
```

`UseUrls` writes through `UseSetting`, which bypasses the configuration provider chain entirely,
so the file's value wins over the host no matter where the providers are registered. #180 could
not reach this.

Verified on `Wallet.Api` after #180 was merged:

```
ASPNETCORE_URLS=http://localhost:61935  ->  [INF] Now listening on: http://localhost:60933
                                            port 61935 never opened
```

A port *can* be set per host today, but only through the bespoke key #180 documented
(`Services__Wallet.Api__Urls__0`), which works and is tested. The standard spelling — the one a
deployment script, a service definition or a container image would use, and the one every
ASP.NET Core operator reaches for first — silently does nothing. Silently is the problem: the
service starts, reports healthy, and listens somewhere other than where it was told.

Likely fix, in each of the five hosts:

```csharp
// A host that set ASPNETCORE_URLS or --urls has already said where to listen; the file's
// value must not overwrite it.
if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKey]) && urls is { Length: > 0 })
{
    builder.WebHost.UseUrls(urls);
}
```

Deliberately not folded into #180: that change was a mechanical reordering of provider
registration, this one changes how all five services choose their listen address, which is the
boot path where a mistake takes the whole platform down at once. It wants its own change and its
own run of the full stack.

## 2. The bot's command line is outranked by the shared file

`TelegramBot/TallaEgg.TelegramBot.Infrastructure/Program.cs` calls `Host.CreateDefaultBuilder(args)`,
which registers a command-line provider among the defaults — before `ConfigureAppConfiguration`
adds the shared file and its flattened section. So the file outranks `--Key=value` there, the
same defect #159 described, in the one host #180 left alone.

The environment-variable half is already correct in the bot; it has called
`AddEnvironmentVariables()` after the flattening all along, and that is what #180 aligned the
APIs with. Nothing in the bot reads a setting from `args` today, so this is a latent
inconsistency rather than a live problem, which is why it was flagged rather than fixed inline
(`STANDARDS.md`, scope discipline).

Fix is one line in `ConfigureAppConfiguration`, beside the existing call:

```csharp
configBuilder.AddInMemoryCollection(flattened);
configBuilder.AddEnvironmentVariables();
configBuilder.AddCommandLine(args);   // <- this
```

`ConfigurationPrecedenceTests.EveryApiHost_RegistersTheCommandLine_AfterTheSharedFile` already
exists and covers the five APIs; adding the bot to that theory's data is the test for this.

## Acceptance

- `ASPNETCORE_URLS=…` sets the listen address on all five services, and the file still supplies it
  when the host says nothing
- `--urls` does the same
- `--Key=value` on the bot's command line beats the shared file
- All five services and the bot start; simulator runs end to end
- Regression tests in `tests/TallaEgg.AllServices.Tests/ConfigurationPrecedenceTests.cs`

## Related

- #159 / #180 — provider registration order, fixed
- #105 — production checklist; this belongs on it for the same reason #159 did

MohKardan:
**Branch Name**: `fix/host-urls-precedence-181`
**Linked Issue/Task**: Closes #181

---

## Description

Two places where the shared file still beat the host, both left standing by #159 / PR #180.

`UseUrls` writes through `UseSetting`, which bypasses the configuration provider chain
entirely, so the five API hosts applied `config/appsettings.global.json`'s `Urls` over
`ASPNETCORE_URLS` and `--urls` no matter how the providers were ordered. The service
started, reported healthy, and listened somewhere other than where it had been told —
silently, which is the part that matters. Each host now falls back to the file only when
the host named no address of its own.

The bot's `Host.CreateDefaultBuilder(args)` registers a command-line provider among the
defaults, which run before `ConfigureAppConfiguration` adds the shared file, so the file
outranked `--Key=value` there. `AddCommandLine(args)` is now registered beside the
`AddEnvironmentVariables()` call that was already in the right place.

A third change follows from the first and is not optional: the five `launchSettings.json`
files no longer name an `applicationUrl`. See Decisions — without it this PR would have
broken `manual-test-run`.

---

## Type of Change
- [x] Hotfix (urgent bug fix)

Marked hotfix only in the sense that it is a defect fix; nothing is on fire, because
everything still runs on one machine. It stops being theoretical at the first deployment,
same as #159.

---

## Changes Made

- `src/{Order/Orders.Api, User/Users.Api, Wallet/Wallet.Api, TallaEgg/TallaEgg.Api, Affiliate/Affiliate.Api}/Program.cs`:
  apply the configured `Urls` only when `builder.Configuration[WebHostDefaults.ServerUrlsKey]`
  is empty — i.e. only when neither `ASPNETCORE_URLS` nor `--urls` named an address.
- `TelegramBot/TallaEgg.TelegramBot.Infrastructure/Program.cs`: `configBuilder.AddCommandLine(args)`
  after `AddInMemoryCollection(flattened)` and `AddEnvironmentVariables()`.
- `src/*/Properties/launchSettings.json` (the same five): `applicationUrl` removed from every
  profile. Nothing else in those files changes — `ASPNETCORE_ENVIRONMENT=Development`, which
  is what the driver and the CI workflow actually need from them, stays.
- `tests/TallaEgg.AllServices.Tests/ConfigurationPrecedenceTests.cs`: 15 new cases
  (4 behavioural, 5 + 5 source-guard theory cases, and the bot added to the command-line guard).

---

## Verified before writing, not assumed

The pre-fix behaviour was reproduced on `main` rather than taken from the issue text, so
that "after" has something to be measured against. Both held exactly:

| On `main`, `Wallet.Api` | Result |
|---|---|
| `ASPNETCORE_URLS=http://localhost:61935`, run the built exe | `Now listening on: http://localhost:60933` — port 61935 never opened |
| Bot, `-- --OrderApiUrl=http://localhost:5140/api/proof-of-181` | `Order API URL: http://localhost:5140/api` — the file's value; the switch ignored |

---

## Decisions this change had to make

**Launch profiles had to stop naming an address, or this PR breaks CI.** `dotnet run` hands
a profile's `applicationUrl` to the process as `ASPNETCORE_URLS` — indistinguishable, by
construction, from a deployment setting it. While `UseUrls` was unconditional that made no
difference: nothing a profile said could reach the server. The moment a host's own address
is honoured, the profile silently outranks the shared file, and three of the five profiles
name an https endpoint that would then have to be bound wherever the stack runs.
`.github/workflows/manual-test-run.yml` starts these services with plain `dotnet run` on
`ubuntu-latest` with no `dotnet dev-certs` step, so `Wallet.Api` would have failed to start
there and taken the job with it. First seen on this machine as a behaviour change:

```
Wallet.Api, dotnet run, before -> http://localhost:60933
Wallet.Api, dotnet run, with the guard but the profile still naming an address
                               -> https://localhost:60932   (from launchSettings)
                                  http://localhost:60933
```

Removing `applicationUrl` is the smaller of the two available repairs — the other being
`--no-launch-profile` plus an explicit `ASPNETCORE_ENVIRONMENT` in the workflow, in
`driver.ps1` and in the skill's documentation, which would also leave the trap live for
anyone typing `dotnet run` by hand. It leaves the shared file as the single source of truth
for where a service listens, which is what `AGENT.md` already says it is, and it keeps
`dotnet run` binding exactly what it binds today (evidence below). A new test keeps
`applicationUrl` out, with the reason in its own docs, since a `.json` file has nowhere to
put a comment.

**The guard reads configuration, not the web host.** `builder.Configuration[WebHostDefaults.ServerUrlsKey]`
sees both spellings, because `WebApplication.CreateBuilder` registers the
`ASPNETCORE_`-prefixed environment provider and the command-line provider, and both land on
the same `urls` key. The file cannot collide with it: the section's `Urls` is an array, so
flattening produces `Urls:0`, never a value at `Urls` itself.

**`Affiliate.Api` included, although the affiliate feature is dormant and outside the
audit's scope.** Same reason as #180: the change is mechanical, the service had the same
defect, and leaving it out would leave one service choosing its listen address by a
different rule from the other four — the inconsistency the acceptance criteria warn about.
It boots clean and honours both overrides (table below).

**The simulator was left alone.** It binds no address and parses its own
`--users`/`--trades` switches in `SimulationOptions.FromArgs`, which a command-line
configuration provider has no part in. The bot's own `launchSettings.json` keeps its
`applicationUrl` for the same reason: it is a generic host with no Kestrel, so the value has
never reached anything, and this PR does not change that.

**`ResolveSharedConfigPath` is still copy-pasted into all seven `Program.cs` files**, and it
is still why a configuration fix costs five or six edits instead of one. Untouched here for
the same reason #180 left it: this issue is about the boot path's precedence, and widening
it into a startup refactor across every service is what `STANDARDS.md` says not to do.

---

## Flagged, not fixed

- **`ASPNETCORE_HTTP_PORTS` / `ASPNETCORE_HTTPS_PORTS` are still silently overridden.** The
  guard reads only `urls`, and `urls` outranks `HTTP_PORTS` in Kestrel's binder, so a host
  that names its port the other way still gets the file's address instead — the same
  listen-somewhere-else shape #181 is about, in a spelling the issue does not mention. It
  matters because the official `mcr.microsoft.com/dotnet/aspnet` images set `HTTP_PORTS` by
  default, so a containerised deployment would hit it first. Extending the same condition to
  those two keys is a few characters per host; raised here rather than done, the way #180
  raised this issue.
- **`Orders.Api` now has two launch profiles, `http` and `https`, that do the same thing.**
  Its `https` profile existed only to name an address, and it never bound one. Deleting it
  is tidy-up outside this issue.

---

### Unit Tests
- [x] New unit tests written

```
-> Build succeeded. 0 Warning(s) 0 Error(s)     (TreatWarningsAsErrors is on)

dotnet test TallaEgg.sln
-> Passed! Failed: 0, Passed: 694, Skipped: 0, Total: 694
```

The suite's existing two-halves shape is kept, because either half alone would be worth
little here:

- **Behavioural** (4 new) — through a real `WebApplicationBuilder`, not a hand-built
  provider chain, since the whole question is about providers `WebApplication.CreateBuilder`
  registers and about `UseUrls` writing back to the same key. `ASPNETCORE_URLS` wins,
  `--urls` wins, and — the one that matters most — with neither set the file still supplies
  the address. A negative control reproduces the old unconditional call and asserts the file
  wins, so the other three cannot pass for a reason unrelated to the guard.
- **Source guards** (11 new theory cases) — the five hosts must reach
  `WebHostDefaults.ServerUrlsKey` before `UseUrls`; no API launch profile may name an
  `applicationUrl`; and the bot joins the existing command-line ordering theory. Comment
  lines are stripped before matching, as before, so a comment naming a provider can neither
  satisfy nor defeat an assertion.

Each new guard was confirmed non-vacuous by reverting exactly one file and watching exactly
its own case go red:

```
Failed …EveryApiHost_AppliesTheConfiguredUrls_OnlyWhenTheHostNamedNoAddress(hostProgram: "src/Wallet/Wallet.Api/Program.cs")
Failed …EveryHostThatParsesTheCommandLine_RegistersIt_AfterTheSharedFile(hostProgram: "TelegramBot/…/Program.cs")
Failed …NoApiLaunchProfile_NamesAListenAddress(launchSettings: "src/Wallet/Wallet.Api/Properties/launchSettings.json")
```

### The risk here is boot, not tests

This is the path where a mistake takes all five services down at once and no unit test shows
it, so every service was actually run — each built executable, three ways: with
`ASPNETCORE_URLS`, with `--urls`, and with neither.

```
Users.Api      env     listening=http://localhost:61936   filePort(5136)=False  overridePort=True   errors=0
Users.Api      switch  listening=http://localhost:61936   filePort(5136)=False  overridePort=True   errors=0
Users.Api      none    listening=http://localhost:5136    filePort(5136)=True   overridePort=False  errors=0
Wallet.Api     env     listening=http://localhost:61935   filePort(60933)=False overridePort=True   errors=0
Wallet.Api     switch  listening=http://localhost:61935   filePort(60933)=False overridePort=True   errors=0
Wallet.Api     none    listening=http://localhost:60933   filePort(60933)=True  overridePort=False  errors=0
Orders.Api     env     listening=http://localhost:61937   filePort(5140)=False  overridePort=True   errors=0
Orders.Api     switch  listening=http://localhost:61937   filePort(5140)=False  overridePort=True   errors=0
Orders.Api     none    listening=http://localhost:5140    filePort(5140)=True   overridePort=False  errors=0
TallaEgg.Api   env     listening=http://localhost:61938   filePort(5135)=False  overridePort=True   errors=0
TallaEgg.Api   switch  listening=http://localhost:61938   filePort(5135)=False  overridePort=True   errors=0
TallaEgg.Api   none    listening=http://localhost:5135    filePort(5135)=True   overridePort=False  errors=0
Affiliate.Api  env     listening=http://localhost:61939   filePort(60812)=False overridePort=True   errors=0
Affiliate.Api  switch  listening=http://localhost:61939   filePort(60812)=False overridePort=True   errors=0
Affiliate.Api  none    listening=http://localhost:60812   filePort(60812)=True  overridePort=False  errors=0
```

The third row of each group is the acceptance criterion this change most risked breaking: no
host address anywhere, and the file still decides. `filePort` and `overridePort` are read
from `Get-NetTCPConnection`, not from the log, so a service that printed one address and
bound another would show up.

**All five services and the bot start together, on the same ports as before this PR, and the
bot reaches its polling loop** (`driver.ps1 start` plus `dotnet run` for the other two and
the bot, i.e. through the launch profiles):

```
users:     [INF] Now listening on: http://localhost:5136
wallet:    [INF] Now listening on: http://localhost:60933
orders:    [INF] Now listening on: http://localhost:5140
tallaegg:  [INF] Now listening on: http://localhost:5135
affiliate: [INF] Now listening on: http://localhost:60812
bot:       [INF] Bot is now running and listening for messages...
-> 0 [ERR]/[FTL] lines across all six logs
-> https ports 60932 / 7296 / 60811 / 57545: not listening, as before this PR
```

**End to end through the real handlers and APIs**, with all six hosts up
(`driver.ps1 smoke`, 5 users / 5 quotes / 10 trades):

```
=== Done in 00:00:10.4. Registered 5 (4 approved), trades attempted 10, errors 0 ===
Simulation completed with errors 0; 10 settlement(s) completed, no new failures.
```

Settlement is included, not just the trading phase — the driver drains the outbox and
requires `Completed` to have grown and `Failed` not to have (#175 / #184).

**The bot's half, at runtime:**

```
bot -- --OrderApiUrl=http://localhost:5140/api/proof-of-181
[INF] Order API URL: http://localhost:5140/api/proof-of-181     <- the switch
[INF] Users API URL: http://localhost:5136/api                  <- still the file
```

The second line is the half that is easy to break while fixing the first: everything the
command line did not name still comes from the shared file. On `main`, the same switch
produced the file's value on both lines.

### Environment
- [x] Tested locally on Windows
- [ ] Tested on staging environment (no staging exists yet — see `STANDARDS.md` §7)

---

## Security Checklist

- [x] `.gitignore` blocks sensitive files (config, .env, secrets)
- [x] Code compiles without warnings (`TreatWarningsAsErrors` is on; 0 warnings)

`config/appsettings.global.json` is untouched and uncommitted. The only values from it that
appear anywhere in this PR are the five service ports, which are already in
`config/appsettings.global.example.json` and `AGENT.md`. The tests generate their own
throwaway config file in a temp directory with invented values and delete it afterwards — CI
has no shared config file and this suite still must not need one.

The direction of travel is the same as #159's: a deployment can now move a listen address
with the standard `ASPNETCORE_URLS`, instead of hand-editing the one file that holds live
credentials on the server.

---

## Code Quality

- [x] Consistent indentation and formatting

---

## Documentation

- [ ] Updated relevant `.md` files in `docs/` — none needed. `AGENT.md`'s Configuration
      section already describes the shared file as the source of truth for what a host does
      not override, which is now true of the listen address as well. Its note that the
      services "also listen on an HTTPS port, which nothing in the system uses" describes
      the launch profiles, and was already untrue of what the services bound; it is now
      untrue in the same way and for the same reason, so it is left for whoever revises that
      list.

---

## Breaking Changes?

- [x] No breaking changes

Nothing changes for a host that names no address — the deployed shape today, every row of
the `none` group above, and every service in the full-stack run. Two things are worth
knowing about:

1. **A pre-existing `ASPNETCORE_URLS`, `DOTNET_URLS` or bare `URLS` in an environment now
   wins** where it previously lost. That is the intended behaviour and there is nothing of
   the sort on this machine or in CI, but it is the shape of surprise this change makes
   possible.
2. **Visual Studio / Rider no longer read a browse address out of the launch profile.** They
   run the project the same way and the service binds the same port; only the "open a
   browser at this URL" hint is gone.

---

## Deployment Notes

**Pre-Deployment**: nothing. No migrations, no new environment variables, no feature flag.

**Post-Deployment Verification**:
- [x] Application starts successfully
- [x] Smoke tests pass
- [x] Logs show no errors

**How to use it.** The bespoke spelling #180 documented still works, and the standard one
now works too:

```
ASPNETCORE_URLS=http://0.0.0.0:60933                 # or --urls http://0.0.0.0:60933
Services__Wallet.Api__Urls__0=http://0.0.0.0:60933   # still honoured
```

If both are set, `ASPNETCORE_URLS` wins — the host's own spelling is the one a deployment
script reaches for, and the nested key is the file's shape.

**Rollback Plan**: revert the two commits and restart. There is no state to undo — the
change is one condition per service, one provider registration in the bot, and one key
removed from five launch profiles.

---

## Related Issues

- Closes #181
- #159 / PR #180 — provider registration order; this is the part that order could not reach
- #105 — production checklist; this belongs on it for the same reason #159 did
- #33 — why `config/appsettings.global.json` is untracked, and why editing it on a server is
  the thing worth avoiding

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Program.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Program.cs
@@ -75,6 +75,12 @@ public class Program
 
                 configBuilder.AddInMemoryCollection(flattened);
                 configBuilder.AddEnvironmentVariables();
+
+                // Host.CreateDefaultBuilder(args) registers a command-line provider among its
+                // defaults, which run before this callback — so the shared file outranked
+                // --Key=value, the same defect #159 fixed in the five APIs. Re-registered here
+                // beside the environment, which was already in the right place (issue #181).
+                configBuilder.AddCommandLine(args);
             })
             .ConfigureServices((context, services) =>
             {
--- a/src/Affiliate/Affiliate.Api/Program.cs
+++ b/src/Affiliate/Affiliate.Api/Program.cs
@@ -50,8 +50,12 @@ builder.Configuration.AddCommandLine(args);
 // نمادهای معاملاتی از appsettings.global.json (بخش Symbols) — نه پیش‌فرض‌های کامپایل‌شده.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
+// UseUrls writes through UseSetting, which bypasses the configuration providers entirely, so
+// calling it unconditionally let the file's address beat ASPNETCORE_URLS and --urls however the
+// providers were ordered — the one override #159 could not reach. The file now supplies the
+// listen address only when the host has not already named one (issue #181).
 var urls = serviceSection.GetSection("Urls").Get<string[]>();
-if (urls is { Length: > 0 })
+if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKey]) && urls is { Length: > 0 })
 {
     builder.WebHost.UseUrls(urls);
 }
--- a/src/Affiliate/Affiliate.Api/Properties/launchSettings.json
+++ b/src/Affiliate/Affiliate.Api/Properties/launchSettings.json
@@ -5,8 +5,7 @@
       "launchBrowser": true,
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
-      },
-      "applicationUrl": "https://localhost:60811;http://localhost:60812"
+      }
     }
   }
 }
--- a/src/Order/Orders.Api/Program.cs
+++ b/src/Order/Orders.Api/Program.cs
@@ -66,8 +66,12 @@ builder.Configuration.AddCommandLine(args);
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
+// UseUrls writes through UseSetting, which bypasses the configuration providers entirely, so
+// calling it unconditionally let the file's address beat ASPNETCORE_URLS and --urls however the
+// providers were ordered — the one override #159 could not reach. The file now supplies the
+// listen address only when the host has not already named one (issue #181).
 var urls = serviceSection.GetSection("Urls").Get<string[]>();
-if (urls is { Length: > 0 })
+if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKey]) && urls is { Length: > 0 })
 {
     builder.WebHost.UseUrls(urls);
 }
--- a/src/Order/Orders.Api/Properties/launchSettings.json
+++ b/src/Order/Orders.Api/Properties/launchSettings.json
@@ -5,7 +5,6 @@
       "commandName": "Project",
       "dotnetRunMessages": true,
       "launchBrowser": false,
-      "applicationUrl": "http://localhost:5140",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       }
@@ -14,7 +13,6 @@
       "commandName": "Project",
       "dotnetRunMessages": true,
       "launchBrowser": false,
-      "applicationUrl": "https://localhost:7140;http://localhost:5140",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       }
--- a/src/TallaEgg/TallaEgg.Api/Program.cs
+++ b/src/TallaEgg/TallaEgg.Api/Program.cs
@@ -54,8 +54,12 @@ builder.Configuration.AddCommandLine(args);
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
+// UseUrls writes through UseSetting, which bypasses the configuration providers entirely, so
+// calling it unconditionally let the file's address beat ASPNETCORE_URLS and --urls however the
+// providers were ordered — the one override #159 could not reach. The file now supplies the
+// listen address only when the host has not already named one (issue #181).
 var urls = serviceSection.GetSection("Urls").Get<string[]>();
-if (urls is { Length: > 0 })
+if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKey]) && urls is { Length: > 0 })
 {
     builder.WebHost.UseUrls(urls);
 }
--- a/src/TallaEgg/TallaEgg.Api/Properties/launchSettings.json
+++ b/src/TallaEgg/TallaEgg.Api/Properties/launchSettings.json
@@ -12,7 +12,6 @@
       "commandName": "Project",
       "dotnetRunMessages": true,
       "launchBrowser": true,
-      "applicationUrl": "https://localhost:7296;http://localhost:5135",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       }
--- a/src/User/Users.Api/Program.cs
+++ b/src/User/Users.Api/Program.cs
@@ -65,8 +65,12 @@ builder.Configuration.AddCommandLine(args);
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
+// UseUrls writes through UseSetting, which bypasses the configuration providers entirely, so
+// calling it unconditionally let the file's address beat ASPNETCORE_URLS and --urls however the
+// providers were ordered — the one override #159 could not reach. The file now supplies the
+// listen address only when the host has not already named one (issue #181).
 var urls = serviceSection.GetSection("Urls").Get<string[]>();
-if (urls is { Length: > 0 })
+if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKey]) && urls is { Length: > 0 })
 {
     builder.WebHost.UseUrls(urls);
 }
--- a/src/User/Users.Api/Properties/launchSettings.json
+++ b/src/User/Users.Api/Properties/launchSettings.json
@@ -5,8 +5,7 @@
       "launchBrowser": false,
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
-      },
-      "applicationUrl": "http://localhost:5136"
+      }
     }
   }
 }
--- a/src/Wallet/Wallet.Api/Program.cs
+++ b/src/Wallet/Wallet.Api/Program.cs
@@ -61,8 +61,12 @@ builder.Configuration.AddCommandLine(args);
 // Trading symbols come from appsettings.global.json (Symbols section), not compiled-in defaults.
 TallaEgg.Core.CurrenciesConstant.Configure(builder.Configuration);
 
+// UseUrls writes through UseSetting, which bypasses the configuration providers entirely, so
+// calling it unconditionally let the file's address beat ASPNETCORE_URLS and --urls however the
+// providers were ordered — the one override #159 could not reach. The file now supplies the
+// listen address only when the host has not already named one (issue #181).
 var urls = serviceSection.GetSection("Urls").Get<string[]>();
-if (urls is { Length: > 0 })
+if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKey]) && urls is { Length: > 0 })
 {
     builder.WebHost.UseUrls(urls);
 }
--- a/src/Wallet/Wallet.Api/Properties/launchSettings.json
+++ b/src/Wallet/Wallet.Api/Properties/launchSettings.json
@@ -5,8 +5,7 @@
       "launchBrowser": true,
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
-      },
-      "applicationUrl": "https://localhost:60932;http://localhost:60933"
+      }
     }
   }
 }
--- a/tests/TallaEgg.AllServices.Tests/ConfigurationPrecedenceTests.cs
+++ b/tests/TallaEgg.AllServices.Tests/ConfigurationPrecedenceTests.cs
@@ -1,4 +1,7 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
 
 namespace TallaEgg.AllServices.Tests;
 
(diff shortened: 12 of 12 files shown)


MohKardan:
## Self-review (`/code-review high`), and what came of it

Three findings, all real. Two are fixed in 647c8c8; one is flagged in the PR body rather
than fixed.

**1. `dotnet run` would have honoured `launchSettings.json`, and that breaks
`manual-test-run` — fixed.** The workflow starts these services with plain `dotnet run` on
`ubuntu-latest` and has no `dotnet dev-certs` step, so once the guard let a host's address
through, `Wallet.Api` would have tried to bind `https://localhost:60932` with no development
certificate and failed to start, taking the `check wallet` step and the job with it.
`driver.ps1` starts them the same way, so any machine without a trusted dev cert would hit
it too. `applicationUrl` is now removed from all five API launch profiles, which leaves the
shared file as the only thing naming a listen address and keeps `dotnet run` binding exactly
what it bound before this PR — verified by rerunning the whole stack: five services on the
same http ports, no https port bound, 0 errors, smoke green. A new test
(`NoApiLaunchProfile_NamesAListenAddress`) keeps `applicationUrl` out and carries the reason,
since JSON has nowhere to put a comment.

**2. `ASPNETCORE_HTTP_PORTS` / `ASPNETCORE_HTTPS_PORTS` are still silently overridden —
flagged, not fixed.** Correct, and it is the same shape of defect: the guard reads only
`urls`, and `urls` outranks `HTTP_PORTS` in Kestrel's binder, so a host that names its port
that way still gets the file's address. It matters mainly for containers, since the official
`aspnet` images set `HTTP_PORTS` by default. #181 names `ASPNETCORE_URLS` and `--urls`
specifically, and this is a different key with its own deployment story, so it is raised in
the PR body the way #180 raised this issue rather than folded in — the fix would be a few
characters per host if you want it here instead.

**3. The "host named no address" tests cleared only `ASPNETCORE_URLS` — fixed.** The chain
also honours `DOTNET_URLS` and a bare `URLS`, either of which would have turned those two
tests red on a machine that had one set, with a message pointing at the shared file. Both are
now cleared alongside `ASPNETCORE_URLS`, through a small `NoAddressInTheEnvironment` helper.

PR #201, opened 2026-09-02
[Hotfix][Critical] BOT_DIRECT_CONNECTION=1 now actually bypasses the proxy — issue #199

### LINKED ISSUE #199: BOT_DIRECT_CONNECTION=1 does not bypass the proxy, but says it does
## What is wrong

`BOT_DIRECT_CONNECTION=1` does not bypass the system proxy. It prints a message saying it did.

`TelegramBot/TallaEgg.TelegramBot.Infrastructure/ProxyBotClient.cs:16-21`:

```csharp
if (Environment.GetEnvironmentVariable("BOT_DIRECT_CONNECTION") == "1")
{
    Console.WriteLine("🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)");
    var directClient = new HttpClient { Timeout = TimeSpan.FromSeconds(120) };
    return new TelegramBotClient(token, directClient);
}
```

`new HttpClient()` uses the stock `HttpClientHandler`, which defaults to `UseProxy = true` and `Proxy = null`. On Windows that resolves through `HttpClient.DefaultProxy` → `HttpWindowsProxy`, which reads the **same WinInet settings** that `WebRequest.GetSystemWebProxy()` reads on line 26. Nothing anywhere sets `UseProxy = false`.

So the flag changes *how* the proxy is resolved — `DefaultProxy` instead of an explicitly constructed handler — but the request still goes through the proxy.

## Why it matters

The flag exists for one scenario, described in its own comment: a machine that can reach Telegram directly (a TUN-mode VPN) while a system HTTP proxy is still configured, where the proxy is stable on short requests and drops `getUpdates` mid-poll with *"The response ended prematurely"*.

In exactly that scenario the flag does not help. The operator:

1. sets `BOT_DIRECT_CONNECTION=1`,
2. sees `🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)`,
3. watches the long poll drop again,
4. and now has a confirmed-good workaround crossed off the list on false evidence.

That is worse than the flag not existing.

## The fix

```csharp
var handler = new HttpClientHandler { UseProxy = false };
var directClient = new HttpClient(handler) { Timeout = TimeSpan.FromSeconds(120) };
```

## Two stale claims to correct at the same time

- **The console message** promises a bypass that does not happen.
- **The XML comment on lines 11-15** says "Set `BOT_DIRECT_CONNECTION=1` to bypass the proxy and connect directly", which is the same claim.

Both become true once `UseProxy = false` is set.

## Related, in the same method

Worth handling in the same PR, since they are the same twenty lines:

- **The exception fallback loses the timeout.** Line 57 returns `new TelegramBotClient(token)` with no `HttpClient`, so it gets the .NET default of **100 seconds** — the value the other three paths deliberately raise to 120 s because `getUpdates` is a long poll. `TelegramBotHostedService` derives its polling-recovery gap from `_botClient.Timeout`, so silently landing on this path also shifts the down-alert threshold.
- **`🔧 Using proxy: {proxyUri}` is printed unconditionally**, on line 29, *before* the check on line 31 that decides whether a proxy applies. When none does, `GetProxy` returns the destination itself and the line reads `🔧 Using proxy: https://api.telegram.org/` — which means *no proxy*, but reads as the opposite. Moving the write inside the `if` makes the output mean what it says.
- **A null `GetProxy` result takes the proxy branch.** `proxyUri` of `null` is `!= new Uri("https://api.telegram.org")`, so the code builds a proxy handler and prints `🔧 Using proxy: ` with nothing after it.

## Provenance

Found while rewriting `NetworkTroubleshooting.md` in #198. That document originally stated the bypass as a guarantee, on the strength of the code comment; it now describes the flag as worth trying rather than reliable, and points here. Once this is fixed, that hedge should be removed.

MohKardan:
**Branch Name**: `fix/direct-connection-really-bypasses-proxy`
**Linked Issue**: #199

---

## Description

`BOT_DIRECT_CONNECTION=1` printed `🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)`
and then sent every request through the system proxy anyway. `new HttpClient()` takes the stock
handler, which leaves `UseProxy = true` with a null `Proxy`, so on Windows it resolves through
`HttpClient.DefaultProxy` — the same WinInet settings `WebRequest.GetSystemWebProxy()` reads.
Nothing set `UseProxy = false`.

The code change is one line. The value of this PR is the evidence, because a flag that lies is
worse than a flag that does not exist: the operator this flag exists for sets it, reads the
success message, watches the long poll drop again, and crosses a working workaround off the list
on false evidence. That is what happened, and "the message printed" is what made it believable —
so this PR does not repeat the mistake of asserting the bypass without showing it.

---

## Type of Change
- [x] Hotfix (urgent bug fix)

---

## Changes Made

- `ProxyBotClient.ChooseConnection` — new `internal` seam returning `(HttpClientHandler?, string)`:
  the handler to build the client on and the line that describes it. The bypass path returns
  `new HttpClientHandler { UseProxy = false }`.
- `ProxyBotClient.CreateHttpClient` — new `internal` seam, the single place the 120 s long-poll
  timeout is applied.
- The exception fallback now supplies an `HttpClient` instead of `new TelegramBotClient(token)`,
  so it keeps 120 s rather than silently taking the .NET default of 100 s. That mattered beyond
  the poll itself: `TelegramBotHostedService.PollingRecoveryGap` derives from
  `ITelegramBotClient.Timeout`, so falling onto that path moved the down-alert threshold too.
- `🔧 Using proxy: {uri}` is printed only when a proxy actually applies. It used to print
  unconditionally, *before* the check, so "no proxy" surfaced as
  `🔧 Using proxy: https://api.telegram.org/` — which means no proxy and reads as its opposite.
  That case now prints `🔧 No system proxy applies to api.telegram.org`.
- A null `GetProxy` result is treated as no proxy. `null != new Uri(...)`, so it used to take the
  proxy branch and print `🔧 Using proxy: ` with nothing after it.
- The fallback message no longer says "direct connection" — the stock handler it falls back to
  still proxies. Behaviour there is unchanged, only the wording.
- The two Persian comments in this method are gone with the code that carried them.
- `InternalsVisibleTo("TallaEgg.AllServices.Tests")` on the bot project, matching
  `Orders.Application` and `TallaEgg.TelegramBot.Simulator`.
- `NetworkTroubleshooting.md`: the #198 hedge is removed, the four-path table matches the new
  behaviour (including the 100 s row, now 120 s), and the console-output section says what each
  of the three lines now means.

### Why a seam, and why the smallest one

`CreateWithProxy` returns `ITelegramBotClient`, which exposes neither its handler nor its proxy
settings — so `UseProxy = false` cannot be asserted through it, and no test could ever have caught
this bug from outside. That is exactly why a confident, false message survived. Two `internal`
statics is the least that makes the claim checkable: `ChooseConnection` returns the handler *and*
the message as one value, so both the proxy decision and the sentence claiming it are assertable;
`CreateHttpClient` makes the timeout assertable on every path including the fallback. Nothing else
in the bot's network layer is touched.

---

### Unit Tests

- [x] New unit tests written — `tests/TallaEgg.AllServices.Tests/ProxyBotClientTests.cs`, 10 cases

```
Build succeeded.
    0 Warning(s)
    0 Error(s)

dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 718, Skipped: 0, Total: 718
```

**The tests fail against the previous code.** Reverting the bypass branch to the stock handler:

```
Failed ProxyBotClientTests.ChooseConnection_BypassRequested_TurnsProxyingOff
Failed ProxyBotClientTests.ChooseConnection_BypassRequested_SendsNoTrafficToAConfiguredProxy
Failed!  - Failed: 2, Passed: 8, Skipped: 0, Total: 10
```

and reverting the null-`GetProxy` handling:

```
Failed ProxyBotClientTests.ChooseConnection_GetProxyReturnsNull_ReportsNoProxy
Failed!  - Failed: 1, Passed: 9, Skipped: 0, Total: 10
```

`SendsNoTrafficToAConfiguredProxy` is socket-level, not a property assertion: it attaches a
loopback proxy to the bypass handler, requests a closed port, and requires the listener to see
nothing while a control request through a proxying handler does reach it. Loopback only, no
outbound traffic, no process-wide state.

### End-to-end

Run on Windows 10 + .NET 9 on 2026-09-02, on a machine that is exactly the case the flag exists
for: WinInet reports a system proxy at `127.0.0.1:10808`, `netsh winhttp` reports direct access,
and Telegram is reachable **only** through that proxy:

```
curl.exe -x http://127.0.0.1:10808 https://api.telegram.org/bot0:0/getMe   →  http_code=401  remote=127.0.0.1:10808
curl.exe --noproxy "*"             https://api.telegram.org/bot0:0/getMe   →  http_code=000  (exit 7, could not connect)
```

That makes the request outcome self-verifying: **reaching Telegram at all proves the proxy was
used.** No message is being taken at its word anywhere below.

#### (a) With the flag set, traffic really does not go through the proxy

The production `ProxyBotClient.CreateWithProxy` driven in a real process, with every TCP
connection the process opened listed alongside the result:

```
CASE: fixed code, BOT_DIRECT_CONNECTION=1
🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)
client.Timeout = 00:02:00
RESULT after 3.1s: RequestException: Exception during making request
---- TCP connections opened by the process ----
  10.10.34.35:443 [SynSent]          ← api.telegram.org, direct. No 127.0.0.1:10808.
```

and the same harness against the **pre-fix** code, which is the bug itself:

```
CASE: pre-fix code, BOT_DIRECT_CONNECTION=1
🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)
client.Timeout = 00:02:00
RESULT after 1.2s: ApiRequestException: Unauthorized     ← reached Telegram, so it proxied
---- TCP connections opened by the process ----
  127.0.0.1:10808 [Established]      ← the proxy it claimed to have bypassed
```

Same message, opposite behaviour. After the fix the connection goes to Telegram's address and
never to the proxy.

The real bot shows the same thing, and comes with its own control — the startup diagnostics build
their own `HttpClient` and so still proxy, while only the `ProxyBotClient`-created client goes
direct:

```
dotnet run --project TelegramBot/TallaEgg.TelegramBot.Infrastructure   (BOT_DIRECT_CONNECTION=1)

🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)
✅ https://api.telegram.org - Status: OK          ← diagnostics' own HttpClient, still proxied
🔧 Testing with proxy settings...
System Proxy: http://127.0.0.1:10808/
[WRN] Could not delete webhook
Telegram.Bot.Exceptions.RequestException: Exception during making request
 ---> System.Net.Http.HttpRequestException: No connection could be made because the target
      machine actively refused it. (api.telegram.org:443)
```

The bot client failing to reach `api.telegram.org:443` while everything sharing the machine's
proxy succeeds is the bypass, observed. On this machine that means the bot cannot run with the
flag set — correctly, because here the proxy is doing real work. `NetworkTroubleshooting.md` now
says so.

#### (b) Without the flag, the previous behaviour is untouched

This is the regression that matters: the server deployment relies on the proxy path.

```
dotnet run --project TelegramBot/TallaEgg.TelegramBot.Infrastructure   (flag unset)

🔧 Using proxy: http://127.0.0.1:10808/
[INF] Webhook deleted successfully
[INF] Bot connection successful: @…
[INF] Bot is now running and listening for messages...
[INF] Application started. Press Ctrl+C to shut down.
```

Harness, same path: `RESULT: ApiRequestException: Unauthorized`, one TCP connection, to
`127.0.0.1:10808 [Established]`. Proxy path unchanged, and the client is still bound to the proxy
explicitly rather than resolving one ambiently.

#### (c) The console output is true in all three cases

| Case | Printed | True? |
|---|---|---|
| `BOT_DIRECT_CONNECTION=1` | `🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)` | yes — process opened no connection to the proxy, see (a) |
| A proxy applies | `🔧 Using proxy: http://127.0.0.1:10808/` | yes — and that is where the connection went, see (b) |
| No proxy applies | `🔧 No system proxy applies to api.telegram.org` | yes — was `🔧 Using proxy: https://api.telegram.org/` |

The third row was produced in a real process without editing the machine's WinInet settings:
`WebRequest.GetSystemWebProxy()` resolves through `HttpClient.DefaultProxy`, so replacing that
with an empty `WebProxy` reproduces "no proxy configured" faithfully.

```
WebRequest.GetSystemWebProxy().GetProxy(telegram) = https://api.telegram.org/
🔧 No system proxy applies to api.telegram.org
```

`GetProxy` returning null is covered by unit test only — no real `IWebProxy` on hand returns it.

### Environment
- [x] Tested locally (Windows 10, .NET 9, live WinInet proxy at `127.0.0.1:10808`)
- [x] Feature flag working as expected — that is the whole PR

---

## Security Checklist

- [x] No hardcoded secrets, passwords, API keys, or tokens — the harness token is a fake with a
      valid shape (`123456789:AAAA…`), and no real token appears in the diff or in this
      description
- [x] No sensitive data in logs or error messages — the bot username is redacted above; the
      evidence needs only the endpoints
- [x] `.gitignore` blocks sensitive files — `config/appsettings.global.json` was copied into the
      worktree to run the bot and is untracked, as designed in #33
- [x] Code compiles without warnings (`TreatWarningsAsErrors`, `--no-incremental`, 0 warnings)

---

## Code Quality

- [x] Code follows naming conventions (`STANDARDS.md`); tests named `Method_Scenario_ExpectedResult`
- [x] Comments in English, explaining why — the two Persian comments in this method are gone
- [x] No new dependencies

Deliberately **not** changed, per `STANDARDS.md` scope discipline: this file's namespace is
`TallaEgg.TelegramBot`, the name of a project that was deleted, and it logs through
`Console.WriteLine` rather than Serilog. Both are wrong and neither is this issue. Say the word
and I will open issues for them.

---

## Documentation

- [x] `NetworkTroubleshooting.md` updated — the #198 hedge removed, the path table corrected, and
      a new section on what each console line means and what each outcome of setting the flag
      tells you
- [x] Code comments in English

---

## Breaking Changes?

- [x] No breaking changes

`BOT_DIRECT_CONNECTION` is unset on the server, which takes the proxy path — unchanged and
verified in (b). The flag's behaviour changes for anyone who sets it, which is the point: it now
does what it always said.

---

## Deployment Notes

**Pre-Deployment**: none. No migration, no configuration, no new environment variable.

**Post-Deployment Verification**: the bot's first console line should read
`🔧 Using proxy: …` on the server exactly as before, and the bot should reach Telegram.

**Rollback**: revert the commit. The flag returns to being a no-op that claims otherwise.

---

## Related Issues

- Closes #199
- Follows #198, which documented this bug rather than fixing it and pointed here
- Touches the timeout `TelegramBotHostedService` reads for its polling-recovery gap (#148)

---

## Final Checklist

- [x] All tests pass locally
- [x] Documentation updated

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/NetworkTroubleshooting.md
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/NetworkTroubleshooting.md
@@ -11,34 +11,35 @@ configured port.)
 
 | Condition | What you get | Timeout |
 |---|---|---|
-| `BOT_DIRECT_CONNECTION=1` | `HttpClient` with the **default** handler | 120 s |
+| `BOT_DIRECT_CONNECTION=1` | `HttpClient` on a handler with `UseProxy = false` | 120 s |
 | A system proxy applies | `HttpClient` with `Proxy` set explicitly from `WebRequest.GetSystemWebProxy()` | 120 s |
-| No system proxy applies | `HttpClient` with the default handler | 120 s |
-| Anything above throws | Bare `TelegramBotClient(token)`, no `HttpClient` supplied | **100 s** |
+| No system proxy applies | `HttpClient` on the stock handler | 120 s |
+| Anything above throws | `HttpClient` on the stock handler | 120 s |
 
 The 120 s matters because `getUpdates` is a long poll and the .NET default of 100 s is too tight
-for it. Note the last row: the exception fallback drops to that 100 s default, and
-`TelegramBotHostedService` derives its polling-recovery gap from `_botClient.Timeout`, so a
-silent fall onto that path shifts the down-alert threshold too.
+for it. Every row gets it now, the exception fallback included: that row used to supply no
+`HttpClient` at all and silently take the 100 s default, which moved the down-alert threshold
+with it, because `TelegramBotHostedService` derives its polling-recovery gap from
+`_botClient.Timeout` (#199).
 
-### Two things the console output will not tell you
+### Reading the console output
 
-**`🔧 Using proxy: …` is printed unconditionally**, on line 29, *before* the check on line 31
-that decides whether a proxy is actually in play. When no proxy applies, `GetProxy` returns the
-destination itself and you get `🔧 Using proxy: https://api.telegram.org/` — which means *no
-proxy*. Do not read that line as evidence of one.
+One line is printed, and it means what it says. Until #199 two of the three did not, so output
+from a build older than that fix cannot be read this way.
 
-**`BOT_DIRECT_CONNECTION=1` does not disable proxying.** It prints
-`🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)` and constructs
-`new HttpClient { Timeout = … }` — the default `HttpClientHandler`, which on Windows has
-`UseProxy = true` and `Proxy = null`, so it falls through to `HttpClient.DefaultProxy` and reads
-the same WinInet settings. Nothing sets `UseProxy = false`.
-
-What the flag really changes is *how* the proxy is resolved — `DefaultProxy` instead of an
-explicit `WebRequest.GetSystemWebProxy()` handler — which is sometimes enough to shake off a
-misbehaving handler, and sometimes not. **Treat it as worth trying, not as a guaranteed bypass.**
-A real bypass needs `new HttpClient(new HttpClientHandler { UseProxy = false })`; that gap is
-tracked as an issue, not fixed here.
+- **`🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)`** — the handler has
+  `UseProxy = false`, so nothing the bot sends is proxied: not the system proxy, not one
+  configured elsewhere in .NET. Before #199 this line printed over a stock handler, which on
+  Windows leaves `UseProxy = true` with a null `Proxy` and so falls through to
+  `HttpClient.DefaultProxy` — the same WinInet settings the proxy path reads. It proxied, and
+  said it did not.
+- **`🔧 Using proxy: <uri>`** — a system proxy applies to `api.telegram.org` and the client is
+  bound to it. Printed only in that case; it used to be printed unconditionally, *before* the
+  check that decides whether a proxy is in play.
+- **`🔧 No system proxy applies to api.telegram.org`** — `GetProxy` answered with the destination
+  itself, or with null. Both mean no proxy. The first used to surface as
+  `🔧 Using proxy: https://api.telegram.org/`, which says *no proxy* and reads as its opposite;
+  the second took the proxy branch outright and printed `🔧 Using proxy: ` with nothing after it.
 
 ## The failure this file exists for
 
@@ -62,8 +63,15 @@ The build is not optional — `dotnet run --no-build` against a stale `bin` runs
 already changed, which here would mean running a binary that predates the flag and concluding it
 does not work.
 
-If the drops continue, the proxy is still in the path — see the note above — and the next step is
-to clear it machine-wide or disconnect the VPN.
+Two outcomes, and they mean different things:
+
+- **The drops continue.** The proxy is not what was dropping them — with the flag set the bot's
+  client does not use one. Look at the VPN and the direct route instead. Note that the flag covers
+  the bot's Telegram client only; the startup diagnostics build their own `HttpClient`, and
+  everything else on the machine still follows the system proxy.
+- **The bot stops connecting entirely**, with `No connection could be made because the target machine
+  actively refused it. (api.telegram.org:443)` or a timeout. The machine cannot reach Telegram
+  directly after all, so the proxy was doing real work. Unset the flag; the fix is elsewhere.
 
 ## When nothing connects at all
 
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/ProxyBotClient.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/ProxyBotClient.cs
@@ -6,6 +6,16 @@ namespace TallaEgg.TelegramBot
 {
     public class ProxyBotClient
     {
+        private static readonly Uri TelegramApiRoot = new("https://api.telegram.org");
+
+        // getUpdates is a long poll, so the .NET default of 100 s is too tight. Every path below
+        // reaches this through CreateHttpClient, the exception fallback included: that fallback
+        // used to hand TelegramBotClient no HttpClient at all and silently take the 100 s
+        // default, and TelegramBotHostedService derives its polling-recovery gap from
+        // ITelegramBotClient.Timeout, so landing there moved the down-alert threshold too
+        // (issue #199).
+        internal static readonly TimeSpan RequestTimeout = TimeSpan.FromSeconds(120);
+
         public static ITelegramBotClient CreateWithProxy(string token)
         {
             // Local override: when the machine can reach Telegram directly (e.g. a TUN-mode VPN),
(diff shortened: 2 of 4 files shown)


MohKardan:
Both review findings applied in f558657, and both were real — thanks. Neither touched production code.

- `RecordingProxy` caught only what `AcceptTcpClientAsync` threw, so a client disconnecting mid-exchange would fault the accept task and `Dispose`'s `Wait` would rethrow it as an `AggregateException` at teardown, burying the actual assertion. The whole loop body is inside the catch now.
- The unreachable destination was a released ephemeral port — precisely the range the OS hands out next. If something claimed it, the "direct" request would succeed and the assertion would then wait out the client's 120 s long-poll timeout. It is `127.0.0.1:1` now, which nothing binds, plus a 10 s cancellation so an unexpected connection fails in seconds instead of stalling a build for two minutes.

Re-verified after the change: `--no-incremental` build 0 warnings / 0 errors, `dotnet test TallaEgg.sln` 718/718, the ten new tests stable over three consecutive runs, and both bypass tests still fail against the pre-fix handler — including the socket-level one, which is the point of it.

Issue #99, opened 2026-08-08
Bot has no centralized error handling or on-disk logging (companion to #88)


MohKardan:
## Summary

Issue #88 covers centralized exception handling for the 5 HTTP APIs via ASP.NET Core's
`IExceptionHandler` / `UseExceptionHandler`. That mechanism does not apply to
`TallaEgg.TelegramBot.Infrastructure`: the bot is not an HTTP server. It's a `BackgroundService`
(`TelegramBotHostedService`) that long-polls Telegram (`StartReceiving`, with webhook mode
explicitly disabled via `DeleteWebhook(dropPendingUpdates: true)`) and dispatches updates through
`HandleUpdateAsync`. There is no ASP.NET request pipeline for `UseExceptionHandler` to attach to,
so this needs its own design, tracked separately from #88.

## Current state

- `HandleUpdateAsync` (`TelegramBotHostedService.cs:173-196`) already wraps update dispatch in one
  top-level try/catch, but it only logs — it never replies to the user. Today, an unhandled
  exception there means the user sees silence, not an error message.
- The ~65 scattered `catch (Exception ex)` blocks the #88 audit counted under `TelegramBot/` are
  mostly deeper in command/handler code, where each one logs and separately sends the user a
  generic Persian error message — because only they have the `chatId`/`botClient` needed to reply.
  A single top-level handler can't do that without a design change.
- **Logging never reaches disk today.** `TelegramBot/.../Program.cs` only wires up
  `Host.CreateDefaultBuilder`'s default providers (console/debug) — no Serilog, no file sink.
  Contrast with all 5 API services, which already write to `logs/{service}-api-.log` via Serilog
  (`WriteTo.File(..., rollingInterval: RollingInterval.Day)`). If the bot process isn't attached
  to a visible console (e.g. running as a background/Windows service), an exception leaves zero
  trace anywhere.

## Why it matters

- Users hit silent failures with no feedback in the chat.
- Debugging today depends on someone watching a live console at the exact moment of failure —
  nothing persists past that.
- Both human developers and AI coding agents debugging this repo need a queryable, on-disk trail:
  structured log entries with enough context (timestamp, chat/user id, command or handler name,
  exception type + full stack trace) to localize a failure fast — the same way the API services'
  `logs/*.log` files already allow today.

## Proposed scope

- [ ] Add Serilog (or equivalent) to the bot with a file sink, matching the API services' existing
[...]

Issue #70, opened 2026-07-30
[Deployment] Process supervision: keep the services alive across crashes and reboots


MohKardan:
Fourth of five deployment issues (KR1). Depends on #68, #33 and #69.

## Summary

KR1 asks for the bot online for **7 consecutive days with no manual intervention**. Today there is no mechanism that restarts a process — not after a crash, not after a reboot. Five or six processes started by hand meet the target only if nothing goes wrong for a week, which is not a plan.

## Why this is its own issue rather than a footnote

The other deployment issues are about getting the product to run **once**. This one is the entire difference between "it ran" and "it stayed up", and it is what the KR is actually measured on:

> **How it is measured**: server uptime log + a daily manual test of one transaction

A crash at 3am on day 4 resets the count to zero. So this must be in place before the 7 days start being counted, not added after the first outage.

## What is needed

A supervisor that starts each process at boot, restarts it if it exits, and writes its output somewhere durable:

- **Linux**: one `systemd` unit per service with `Restart=always` and `RestartSec`. Logs go to `journald` for free.
- **Windows**: `sc.exe` or NSSM, one service per process, recovery set to restart.

Either is a few hours' work. No container runtime or orchestrator is needed for six processes on one host — see the note on Docker below.

## Start-order matters here

`Orders.Api` and `Users.Api` call `Wallet.Api` on startup paths, and all four run `MigrateAsync()` against the database as the first thing they do. A supervisor that launches everything simultaneously will produce startup failures that are pure noise. `Restart=always` absorbs it, but it will fill the logs and make a real failure hard to see on the day it happens. Prefer explicit ordering (`After=` on Linux) or a short delay.

## Also worth doing while here

**Log rotation.** Serilog is configured with `WriteTo.File("logs/wallet-api-.log", rollingInterval: RollingInterval.Day)` — it rolls daily but nothing ever deletes old files. Over a long-running deployment that grows without bound. `retainedFileCountLimit` is a one-line fix.

**A liveness signal.** The daily manual transaction test the KR describes is fine for the review, but knowing the bot has stopped responding *before* a reviewer discovers it is worth more. Simplest version that needs no new infrastructure: the existing `TelegramLoggerService` already sends messages to a Telegram chat — have the supervisor or the bot report a restart through it. Note that its token is currently hardcoded to an unrelated dev bot; that is covered in #33.

## On Docker

[...]

PR #202, opened 2026-09-02
[Hotfix] Log the proxy decision instead of printing it to a console the service does not have

MohKardan:
**Branch Name**: `fix/proxy-diagnostics-reach-the-log`
**Linked Issue**: none — follow-up to #199 / #201, raised in that PR's review and requested inline
rather than as an issue

---

## Description

`ProxyBotClient` reported which of its four connection paths it took with `Console.WriteLine`. The
bot is installed with `sc.exe create` (#70), and a native Windows service has no console: on the
server all four lines were written to a handle nobody holds.

Those are the lines that say whether the bot is proxying. After #199 they are also the lines an
operator has been explicitly told to read carefully — and they were unreadable on the one machine
where the failure they describe actually happens. Fixing the message's honesty and leaving it
undeliverable was half a fix.

---

## Type of Change
- [x] Hotfix (urgent bug fix)

---

## Changes Made

- `ProxyBotClient.CreateWithProxy` takes an `ILogger`; `Program.cs` passes
  `ILogger<ProxyBotClient>` from the container, matching how `OrderApiClient`, `UsersApiClient` and
  the rest of that registration block already get theirs.
- The decision line is logged as a single property — `LogInformation("{TelegramConnection}", message)`
  — not a template with holes. The sentence is the artefact here: it is worded for an operator to
  act on, and `ProxyBotClientTests` asserts it whole. Splitting it into placeholders would only
  make those assertions weaker, and this is a message whose exact wording caused #199.
- The exception path logs at **Warning**, not Information — it silently changes how the bot reaches
  Telegram — and passes the exception object rather than just `ex.Message`, so the file sink keeps
  the stack trace. On the server there is no console to re-run it in.
- `ProxyBotClient.ResolveHandler` splits handler resolution from client construction, so the
  `catch` covers only what its message blames. It previously wrapped
  `new TelegramBotClient(token, …)` as well, which throws `ArgumentException` on a malformed token
  — a configuration typo logged "could not configure the proxy" and then threw the same exception
  again out of the DI factory from the fallback.
- `ProxyBotClientTests` pins all three connection messages **verbatim**. The runbook quotes them
  and tells an operator which action each one calls for, so a silent reword is precisely the #199
  failure mode; substring assertions would not have caught it.
- `NetworkTroubleshooting.md`: the section is "Reading the line it logs" now, documents the
  fallback warning as the fourth path (and that it is *not* a bypass), and gives the log file's
  location per launch method.

### Scope

This file only. `NetworkTest`, `ProxyTest`, `SimpleHttpTest`, `TestBotToken` and `OfflineTestMode`
still use `Console.WriteLine` — 44 calls between them — and are still invisible under the service.
That is a real gap and it is not this change; the doc now states it plainly instead of leaving it
to be rediscovered.

---

### Unit Tests
- [x] Three new cases pinning the exact wording of each connection message

```
Build succeeded.
    0 Warning(s)
    0 Error(s)

dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 721, Skipped: 0, Total: 721
```

### End-to-end

Run on 2026-09-02 on a machine with a live WinInet proxy at `127.0.0.1:10808`. The claim is that
the line reaches the **file** sink, so the file is what is checked — not the console.

**Where that file actually is.** Serilog's path here is relative, and it resolves against the
process's working directory rather than the binary's folder — verified by running the built exe
with a working directory different from its own, which put the new `logs\` at the working
directory. `sc.exe create` sets no working directory, so under the SCM the file is
`C:\Windows\System32\logs\`. The sink path is deliberately **not** changed here: all four services
share this pattern and `WINDOWS_DEPLOYMENT.md` documents the current locations, making that a
deployment decision rather than part of a logging fix. What this PR changes is that the line
reaches a file at all, instead of a console the service does not have.

**`BOT_DIRECT_CONNECTION=1`** — console and file, same line:

```
console:  [17:11:43 INF] 🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)
logs/telegrambot-20260902.log:
          2026-09-02 17:11:43.951 +03:30 [INF] 🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)
```

**Flag unset**, the path the server actually takes:

```
logs/telegrambot-20260902.log:
          2026-09-02 17:12:30.500 +03:30 [INF] 🔧 Using proxy: http://127.0.0.1:10808/
          2026-09-02 17:12:49.990 +03:30 [INF] Bot is now running and listening for messages...
```

Before this change the log file contained neither line. The bot still starts, connects through the
proxy and receives updates — the behaviour #201 verified is untouched, only the destination of the
report changed.

### Environment
- [x] Tested locally (Windows 10, .NET 9)

---

## Security Checklist

- [x] No sensitive data in logs — the logged line names the proxy endpoint and nothing else; the
      token was never part of any of these messages and still is not
- [x] Code compiles without warnings (`TreatWarningsAsErrors`, 0 warnings)

---

## Code Quality

- [x] Follows `AGENT.md` — "Use `ILogger<T>` for logging"
- [x] No new dependencies — `Microsoft.Extensions.Logging` is already referenced

---

## Documentation

- [x] `NetworkTroubleshooting.md` updated

---

## Breaking Changes?

- [x] No breaking changes

`CreateWithProxy` gains a required parameter, and its one caller is updated in the same commit.

---

## Deployment Notes

**Pre-Deployment**: none.

**Post-Deployment Verification**: `logs/telegrambot-<date>.log` should now open with
`[INF] 🔧 Using proxy: …` on the server. Its absence is itself the signal — either the build did
not deploy, or startup failed before the client was constructed.

**Rollback**: revert the commit. The lines go back to being written to a console that does not
exist.

---

## Related Issues

- Follows #201 (issue #199), which made these messages true; this makes them reachable
- Depends on the Serilog file sink added in #99, and the service install in #70

---

## Final Checklist

- [x] All tests pass locally
- [x] Documentation updated

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/NetworkTroubleshooting.md
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/NetworkTroubleshooting.md
@@ -24,15 +24,8 @@ with it, because `TelegramBotHostedService` derives its polling-recovery gap fro
 
 ### Reading the line it logs
 
-Exactly one line is logged, and it means what it says. Until #199 two of the three did not, so
-output from a build older than that fix cannot be read this way.
-
-It goes through `ILogger`, so it reaches the console during a local `dotnet run` **and**
-`logs/telegrambot-<date>.log`. The second half is the point on a server: the bot is installed with
-`sc.exe create` (#70) and a native Windows service has no console, so while these were
-`Console.WriteLine` they went nowhere on the one machine where a proxy problem is worth
-diagnosing. Everything else in this project still printing with `Console.WriteLine` — the startup
-diagnostics in particular — is still invisible there.
+One line is logged, and it means what it says. Until #199 two of the four did not, so output from
+a build older than that fix cannot be read this way.
 
 - **`🔗 Direct connection (proxy bypassed via BOT_DIRECT_CONNECTION=1)`** — the handler has
   `UseProxy = false`, so nothing the bot sends is proxied: not the system proxy, not one
@@ -47,6 +40,32 @@ diagnostics in particular — is still invisible there.
   itself, or with null. Both mean no proxy. The first used to surface as
   `🔧 Using proxy: https://api.telegram.org/`, which says *no proxy* and reads as its opposite;
   the second took the proxy branch outright and printed `🔧 Using proxy: ` with nothing after it.
+- **`Could not resolve the system proxy; falling back to the default handler…`** — logged at
+  **Warning**, with the exception attached, and the only one of the four with no emoji. Reading the
+  system proxy threw. The client is built on the stock handler, which still consults
+  `HttpClient.DefaultProxy`, so this is not a bypass: treat it as "proxy state unknown".
+
+### Where the line goes
+
+Through `ILogger`, so both the console and Serilog's rolling file. That matters because the bot is
+installed with `sc.exe create` (#70) and a native Windows service has no console — while these
+were `Console.WriteLine` they went nowhere at all under the SCM.
+
+**Know which file, though.** The sink path in `Program.cs` is *relative*, and Serilog resolves it
+against the process's **working directory**, not the folder the binary sits in:
+
+| How it was started | Where `telegrambot-<date>.log` lands |
+|---|---|
+| `dotnet run --project …` | the project folder's `logs\` |
+| The published exe, launched from a shell | `logs\` under whatever that shell's directory was |
+| A service from `sc.exe create` | `C:\Windows\System32\logs\` — `sc.exe` sets no working directory |
+
+So on a server, look under the working directory of the service, not next to the exe.
+`docs/operations/WINDOWS_DEPLOYMENT.md` describes the log as `logs\telegrambot-.log` relative to
+the publish folder, which holds only when something set the working directory there.
+
+Everything else in this project still printing with `Console.WriteLine` — the startup diagnostics
+in particular — remains invisible under the service regardless.
 
 ## The failure this file exists for
 
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/NetworkTroubleshooting.md
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/NetworkTroubleshooting.md
@@ -22,10 +22,17 @@ for it. Every row gets it now, the exception fallback included: that row used to
 with it, because `TelegramBotHostedService` derives its polling-recovery gap from
 `_botClient.Timeout` (#199).
 
-### Reading the console output
+### Reading the line it logs
 
-One line is printed, and it means what it says. Until #199 two of the three did not, so output
-from a build older than that fix cannot be read this way.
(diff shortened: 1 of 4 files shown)


MohKardan:
All four findings were right; fixed in 12eba6f. The first contradicted a claim in this PR's own description, so I have corrected the description too.

**Relative log path.** Confirmed rather than reasoned about: I ran the built exe with a working directory different from its own folder, and the new `logs\` appeared at the *working directory*. `sc.exe create` sets none, so under the SCM the file is `C:\Windows\System32\logs\telegrambot-<date>.log`. The doc now gives the location per launch method instead of one that only holds under `dotnet run`.

I did **not** change the sink path. All four services share this relative-path pattern and `WINDOWS_DEPLOYMENT.md` documents the current locations, so making the bot alone absolute is a deployment decision, not something to smuggle into a logging fix. The PR's benefit stands either way — a file that exists somewhere beats a console handle nobody holds — but the description overstated where, and that is now fixed.

**"Asserted whole in tests" was false.** Every assertion was a substring, so a reword would have passed silently — in the one file whose whole point (#199) is that these sentences must fail a test when they stop being true. All three are pinned verbatim now, with a comment saying why exact and not `Contains`.

**The warning blamed a cause its try block could not support.** Worse than misleading: the try also wrapped `new TelegramBotClient(token, …)`, which throws `ArgumentException` on a malformed token, so a config typo logged "could not configure the proxy" and then threw the same exception again out of the DI factory from the fallback. Handler resolution moved into `ResolveHandler` and client construction sits outside the try, so the catch covers only proxy resolution and a bad token fails once, plainly.

**Three bullets against four paths.** The fallback warning is documented now, including the part that matters — it is not a bypass, because the stock handler still consults `DefaultProxy`.

`--no-incremental` build 0 warnings / 0 errors, `dotnet test` 721/721 (three new theory cases), and the bot re-run with the flag still logs the line and still goes direct after the restructure.

Separately, and outside this PR: `ResolveSharedConfigPath` walks up from `Directory.GetCurrentDirectory()`, which under the SCM is `C:\Windows\System32`. If that holds, no service started by `install-services.ps1` can find `config/appsettings.global.json` at all. I have not touched it and am not claiming it verified — it needs a real service install to confirm, and it is a deployment question rather than a logging one.

MohKardan:
Recorded properly: this is now a comment on #105, the open production launch checklist, rather than living only in this PR's thread — https://github.com/MohKardan/TallaEgg/issues/105#issuecomment-5513752426

It also got sharper on the way there. The bot is the **only** one of the four deployed services that resolves its shared config from the working directory; `Users.Api/Program.cs:501` and its siblings walk up from `environment.ContentRootPath`, which `UseWindowsService()` points at the publish folder. And I confirmed by running it that the bot dies in `HostBuilder.Build()` with `FileNotFoundException` when the working directory has no `config/` ancestor — so if a service's working directory really is `C:\Windows\System32`, `TallaEggBot` never starts. That last link is the one I have not observed, and #105 carries the one command that settles it.

PR #204, opened 2026-09-02
[Fix] Remove the unreachable create-default-wallets endpoint — issue #190

### LINKED ISSUE #190: POST /api/user/{userId}/create-default-wallets is broken two ways and has no caller
## What is wrong

`src/User/Users.Api/Program.cs:466` maps a public endpoint that cannot succeed. Two independent defects:

```csharp
using var httpClient = new HttpClient();
httpClient.BaseAddress = new Uri("https://localhost:7001/");
var response = await httpClient.PostAsync($"api/wallet/create-default/{userId}", null);
```

1. **Hardcoded, wrong port.** `Wallet.Api` listens on `http://localhost:60933` (`config/appsettings.global.json`, `Services:Wallet.Api:Urls`). Port 7001 has never been used by this system. The line bypasses configuration entirely, which is the opposite of the rule in `AGENT.md` — all service URLs come from the shared config file.

2. **Wrong HTTP verb.** Even against the right port it would fail: `src/Wallet/Wallet.Api/Program.cs:309` maps that route with `MapGet`, not `MapPost`, so the call gets 405.

Every request therefore falls into the `else` branch and returns HTTP 500 with `ApiResponse.Error("خطا در ایجاد کیف پول‌های پیش‌فرض.")`.

## Why nobody noticed

The real registration path does not use this endpoint. `UserService.CreateDefaultWalletsAsync` (private, called from `RegisterUserAsync`) resolves the named `"WalletAPI"` client, which reads its `BaseAddress` from configuration, and calls `GetAsync` — right port, right verb. Default wallets do get created on registration.

`grep -rn "create-default-wallets"` across the repository returns exactly one hit: the `MapPost` that defines it. **Nothing calls it** — not the bot, not another service, not a script, not a test.

## Options

**A — delete the endpoint.** It has no consumer, and the behaviour it was meant to expose already happens automatically at registration. This is the recommended resolution: the repository's four APIs are loopback-bound and called only by the bot on the same host (README, "Ports and bind addresses"), so there is no external consumer that could break.

**B — fix it.** Inject the configured `"WalletAPI"` client and call `GetAsync`, exactly as `UserService` does.

## Related, in the same file

`src/User/Users.Api/Program.cs:114` gives the `WalletAPI` client the fallback `?? "https://localhost:60932/"` — also a port this system does not use. Configuration always supplies the value, so the fallback is dead, but it contradicts the documented rule that missing configuration should fail at startup rather than fall back to a default.

## Provenance

Found while verifying documentation claims for #189. `DEFAULT_WALLETS_FEATURE.md` (deleted there) documented "ports 7000/7001"; that number was not stale documentation, it matched this live defect.

MohKardan:
## Description

`POST /api/user/{userId}/create-default-wallets` in `Users.Api` could never succeed, and nothing
called it. It is removed. The dead wallet-URL fallback in the same file is replaced with a
startup guard, so a missing key stops the service instead of silently defaulting.

## Issue/Task Reference

Closes #190

## Type of Change

- [ ] Hotfix (urgent bug fix)

Bug fix, but not urgent: the endpoint has never worked, so nothing is currently regressing.

All three were confirmed against the tree, not taken from the issue:

| Claim | Confirmed at |
| --- | --- |
| Hardcoded `https://localhost:7001/` | `src/User/Users.Api/Program.cs:479` (before this PR) |
| Target route is `MapGet`, so `PostAsync` gets 405 | `src/Wallet/Wallet.Api/Program.cs:309` |
| No consumer | `grep -rn "create-default-wallets" .` returned only the mapping at line 466 |

Port 7001 appears in neither the live shared config nor
`config/appsettings.global.example.json`. The scheme was also wrong: no service in this system
binds HTTPS, so the call failed in the TLS handshake before the verb ever mattered, and with no
`try` around it every request returned 500.

## Why removal rather than repair

The behaviour the endpoint was meant to expose already happens on its own. The real path:

```
POST /api/user/register                   Users.Api:227
POST /api/user/register-with-invitation   Users.Api:406
        ↓ both
UserService.RegisterUserAsync             UserService.cs:53, :163
        ↓
UserService.CreateDefaultWalletsAsync     UserService.cs:202  (private)
        ↓ _httpClientFactory.CreateClient("WalletAPI"), BaseAddress from configuration
GET api/wallet/create-default/{userId}    UserService.cs:209
        ↓
MapGet /api/wallet/create-default/{userId}  Wallet.Api:309
        ↓
WalletService.CreateDefaultWalletsAsync   → Toman, MAUA, CREDIT_MAUA
```

Right port, right verb, address from configuration. Removing the broken endpoint loses nothing.
The four APIs are loopback-bound and called only by the bot on the same host, so there is no
external consumer that could break.

## Changes Made

- Removed the `create-default-wallets` mapping and its handler from `src/User/Users.Api/Program.cs`
- Added `ConfigurationGuard.RequireUri`, alongside the existing `RequireConnectionString`. It
  returns a parsed `Uri`, so a value that is present but unusable — a bare port, a host with no
  scheme, a leftover placeholder — fails the same way an absent one does
- `Users.Api` now reads `WalletApiUrl` through it, replacing `?? "https://localhost:60932/"` —
  another port this system does not use
- Both the read and the parse sit outside the `AddHttpClient` configure delegate, which does not
  run until the first client is created. Inside it, a bad address surfaced as a failed
  registration long after startup — and `UserService.CreateDefaultWalletsAsync` catches
  everything and discards the result, so it surfaced as nothing at all. Outside it, the service
  refuses to start, which is what `AGENT.md` requires
- Eight new `RequireUri` tests covering presence, blankness, four malformed shapes, and the
  message contents

### Build

```
```

### Unit tests

```
dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 732, Skipped: 0, Total: 732
```

### The registration path still works — the only thing this PR could break

A real registration through `POST /api/user/register` against the running stack, then the wallet
database queried directly for that user id:

```
CREDIT_MAUA  balance=0.00000000  locked=0.00000000
IRT          balance=0.00000000  locked=0.00000000
MAUA         balance=0.00000000  locked=0.00000000
```

All three default wallets, created within the same second as the user row. The probe user was in
the simulator's `TelegramId < 0` range and has since been cleared by `DataReset`.

The removed endpoint now answers 404, as expected.

### End to end

```
=== Done in 00:00:18.3. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:
  BTC/IRT: 17 filled
  MAUA/IRT: 17 filled
  SEKE_BAHAR/IRT: 16 filled
```

Registration is phase 1 of that run, so all 20 users took the same path.

### The startup guard, both directions

With the key present, all three services come up (`All services up after 2s.`). With
`Services:Users.Api:WalletApiUrl` removed from a scratch copy of the shared config, `Users.Api`
stops before binding a port and exits non-zero:

```
Unhandled exception. System.InvalidOperationException: Configuration value 'WalletApiUrl' is
missing. Add it under this service's own section, Services:{ApplicationName}, in
appsettings.global.json. The service will not start without it.
   at TallaEgg.Core.ConfigurationGuard.RequireUri(...) in ConfigurationGuard.cs
   at Program.<Main>$(String[] args) in src/User/Users.Api/Program.cs:line 114
```

Present-but-unusable values are rejected at startup too — the case self-review caught, since
`new Uri(...)` had been left inside the configure delegate:

```
--WalletApiUrl=localhost:60933  →  "is not an absolute http(s) URL: 'localhost:60933'"
--WalletApiUrl=60933            →  "is not an absolute http(s) URL: '60933'"
```

`localhost:60933` is the one worth naming: it parses as an absolute URI whose scheme is
`localhost`, so only a scheme check catches it. No fallback, no silent start against the wrong
host, no failure deferred to the first registration.

## Security Checklist

- [x] No secrets/tokens in the diff — `config/appsettings.global.json` is untouched and uncommitted

## Reported, deliberately not fixed

**`Wallet.Api/Program.cs:309` is a `MapGet` that creates wallets** — a safe verb that mutates
state. Risk today is low: the APIs bind loopback only and sit behind an API key. But with this
PR merged it becomes the only remaining path to default-wallet creation, which makes it worth
correcting on its own terms rather than as a side effect here. **Should we open an issue for it?**

**Four more configuration fallbacks on ports this system does not use**, left alone under scope
discipline:

| Location | Fallback | Note |
| --- | --- | --- |
| `src/Order/Orders.Api/Program.cs:120` | `http://localhost:60933` | right port, still a fallback |
| `src/TallaEgg/TallaEgg.Infrastructure/Clients/UsersApiClient.cs:26` | `http://localhost:5001/api` | 5001 is not used here |
| `src/TallaEgg/TallaEgg.Infrastructure/Clients/WalletApiClient.cs:43, :51` | `http://localhost:60933/api` | right port, still a fallback |
| `TelegramBot/.../Clients/OrderApiClient.cs:26` | `http://localhost:5135/api` | 5135 is `TallaEgg.Api`, which maps no endpoints; orders are on 5140 |

`RequireUri` now exists for whoever picks these up. Worth a follow-up issue.

**Two more from self-review** (`/code-review 204`, answered in full in a comment below): the
configuration guards throw before Serilog is configured, so under `sc.exe` the message reaches no
log an operator is told to read; and `UserService.CreateDefaultWalletsAsync` catches every
exception, reports it with `Console.WriteLine` while holding an injected logger, and returns a
bool both callers discard — so a registration whose wallet creation fails is silent. Both are
pre-existing and outside this issue. Worth issues of their own.

## Breaking Changes?

- [x] No breaking changes

Removing an endpoint is a contract change in principle, but this one returned 500 to every
caller and had none.

## Deployment Notes

No migrations, no new configuration. `Services:Users.Api:WalletApiUrl` was already required in
practice and is present in both the live config and the example file; it is now required in fact.
A deployment missing it will fail fast at startup with a message naming the key — the intended
behaviour, but worth knowing before a rollout.

**Rollback**: revert the commit. Nothing depends on the removed endpoint.

### DIFF
--- a/src/TallaEgg/TallaEgg.Core/ConfigurationGuard.cs
+++ b/src/TallaEgg/TallaEgg.Core/ConfigurationGuard.cs
@@ -54,21 +54,30 @@ public static class ConfigurationGuard
     }
 
     /// <summary>
-    /// Returns the named configuration value, or throws naming what is missing and where it belongs.
+    /// Returns the named configuration value as an absolute http(s) URI, or throws naming what
+    /// is wrong and where it belongs.
     /// </summary>
     /// <remarks>
-    /// The same reasoning as <see cref="RequireConnectionString"/>, applied to the settings a
-    /// service needs before it can talk to anything: the address of another service, for one.
-    /// A fallback literal there is worse than no value at all, because the service starts and
-    /// then dials a host nobody configured — issue #190 found
+    /// The same reasoning as <see cref="RequireConnectionString"/>, applied to the address of
+    /// another service. A fallback literal there is worse than no value at all: the service
+    /// starts and then dials a host nobody configured — issue #190 found
     /// <c>GetValue&lt;string&gt;("WalletApiUrl") ?? "https://localhost:60932/"</c> pointing at a
     /// port this system has never used, kept alive only by configuration always supplying the
     /// real one.
+    ///
+    /// It returns a <see cref="Uri"/> rather than the string so that the parse happens here,
+    /// at startup, and not wherever the caller eventually constructs one. A value that is
+    /// present but not a usable address — a bare port, a host with no scheme, a leftover
+    /// placeholder — is a configuration mistake in exactly the way an absent one is, and both
+    /// have to stop the service before it accepts a request. <c>HttpClient</c> would otherwise
+    /// raise it on the first call, far from the cause.
     /// </remarks>
     /// <param name="configuration">Configuration the service was built with.</param>
     /// <param name="key">Configuration key, e.g. <c>WalletApiUrl</c>.</param>
-    /// <exception cref="InvalidOperationException">The value is absent, empty, or whitespace.</exception>
-    public static string RequireValue(IConfiguration configuration, string key)
+    /// <exception cref="InvalidOperationException">
+    /// The value is absent, empty, whitespace, or not an absolute http(s) URI.
+    /// </exception>
+    public static Uri RequireUri(IConfiguration configuration, string key)
     {
         ArgumentNullException.ThrowIfNull(configuration);
         ArgumentException.ThrowIfNullOrWhiteSpace(key);
@@ -79,10 +88,22 @@ public static class ConfigurationGuard
         if (string.IsNullOrWhiteSpace(value))
         {
             throw new InvalidOperationException(
-                $"Configuration value '{key}' is missing. Add it under the service's section in " +
-                $"{SHARED_CONFIG_FILE_NAME}. The service will not start without it.");
+                $"Configuration value '{key}' is missing. Add it under this service's own " +
+                $"section, Services:{{ApplicationName}}, in {SHARED_CONFIG_FILE_NAME}. The " +
+                $"service will not start without it.");
         }
 
-        return value;
+        // Uri.TryCreate alone is not enough: "localhost:60933" parses as absolute with the
+        // scheme "localhost", and HttpClient rejects it only when a request is finally made.
+        if (!Uri.TryCreate(value, UriKind.Absolute, out var uri)
+            || (uri.Scheme != Uri.UriSchemeHttp && uri.Scheme != Uri.UriSchemeHttps))
+        {
+            throw new InvalidOperationException(
+                $"Configuration value '{key}' is not an absolute http(s) URL: '{value}'. Fix it " +
+                $"under this service's own section, Services:{{ApplicationName}}, in " +
+                $"{SHARED_CONFIG_FILE_NAME}. The service will not start without a usable address.");
+        }
+
+        return uri;
     }
 }
--- a/src/TallaEgg/TallaEgg.Core/ConfigurationGuard.cs
+++ b/src/TallaEgg/TallaEgg.Core/ConfigurationGuard.cs
@@ -52,4 +52,37 @@ public static class ConfigurationGuard
 
         return value;
     }
+
+    /// <summary>
+    /// Returns the named configuration value, or throws naming what is missing and where it belongs.
+    /// </summary>
+    /// <remarks>
+    /// The same reasoning as <see cref="RequireConnectionString"/>, applied to the settings a
+    /// service needs before it can talk to anything: the address of another service, for one.
+    /// A fallback literal there is worse than no value at all, because the service starts and
+    /// then dials a host nobody configured — issue #190 found
+    /// <c>GetValue&lt;string&gt;("WalletApiUrl") ?? "https://localhost:60932/"</c> pointing at a
+    /// port this system has never used, kept alive only by configuration always supplying the
+    /// real one.
+    /// </remarks>
+    /// <param name="configuration">Configuration the service was built with.</param>
+    /// <param name="key">Configuration key, e.g. <c>WalletApiUrl</c>.</param>
+    /// <exception cref="InvalidOperationException">The value is absent, empty, or whitespace.</exception>
+    public static string RequireValue(IConfiguration configuration, string key)
+    {
+        ArgumentNullException.ThrowIfNull(configuration);
+        ArgumentException.ThrowIfNullOrWhiteSpace(key);
+
+        var value = configuration[key];
+
+        // Whitespace counts as missing, for the same reason it does for a connection string.
+        if (string.IsNullOrWhiteSpace(value))
+        {
+            throw new InvalidOperationException(
+                $"Configuration value '{key}' is missing. Add it under the service's section in " +
+                $"{SHARED_CONFIG_FILE_NAME}. The service will not start without it.");
+        }
+
+        return value;
+    }
 }
--- a/src/User/Users.Api/Program.cs
+++ b/src/User/Users.Api/Program.cs
@@ -108,13 +108,15 @@ builder.Services.AddScoped<UserService>();
 builder.Services.AddScoped<UserMapper>();
 builder.Services.AddTallaEggErrorHandling();
 
-// HttpClient for calling the Wallet API. The address is read here rather than inside the
-// configure delegate, which would not run until the first client was created: a missing key
-// has to stop the service coming up, not surface later as a failed registration.
-var walletApiUrl = ConfigurationGuard.RequireValue(builder.Configuration, "WalletApiUrl");
+// HttpClient for calling the Wallet API. The address is read and parsed here rather than
+// inside the configure delegate, which would not run until the first client was created. A bad
+// address has to stop the service coming up: registration creates the default wallets through
+// this client and swallows what it throws (UserService.CreateDefaultWalletsAsync), so deferring
+// the failure means users registered with no wallets and nothing said about it.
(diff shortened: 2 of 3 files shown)


MohKardan:
## Self-review response (`/code-review 204`)

Eight findings. Three fixed in 7f40489, five answered below.

### Fixed

**1. `new Uri(...)` left inside the configure delegate.** Correct, and the more serious of the
two because the comment I added claimed the opposite. The read was hoisted; the parse was not,
so only an *absent* key stopped the service. A present-but-unusable one still threw on the first
`CreateClient`, inside `UserService.CreateDefaultWalletsAsync`, whose `catch (Exception)` writes
to the console and returns a bool both call sites discard — a user committed with no wallets and
nothing recorded. Verified against the running service:

```
--WalletApiUrl=localhost:60933  →  InvalidOperationException at startup, no port bound
--WalletApiUrl=60933            →  InvalidOperationException at startup, no port bound
--WalletApiUrl=                 →  InvalidOperationException at startup, no port bound
```

**2. The guard validated presence only.** Taken with finding 1: `RequireValue` is now
`RequireUri`, returning a parsed `Uri` rather than a string, so the parse happens once at startup
instead of at each call site. `Uri.TryCreate` alone is not sufficient — `localhost:60933` parses
as absolute with scheme `localhost` — so the scheme is checked too. Putting this in the shared
guard rather than at the Users.Api call site also makes it unit-testable, which is what partly
answers finding 5.

**6. The message did not name the section.** Fixed, and the reasoning holds: the shared config
defines `WalletApiUrl` in three sections with two incompatible shapes — `Users.Api` takes the
bare root, `Orders.Api` and the bot take a `/api` suffix — so copying the nearest one yields
`/api/api/...`, a 404, swallowed exactly as described. The message now names
`Services:{ApplicationName}`.

Tests: 726 → 732. `RequireUri_WhenTheValueIsNotAnAbsoluteHttpUrl_Throws` covers the bare port,
the scheme-less host, a relative path, a leftover placeholder, and a non-http scheme.

### Not fixed, deliberately

**3. The guard throws before Serilog is configured, so under `sc.exe` the message reaches no
log.** Real, and it lands on the pre-existing `RequireConnectionString` at `Program.cs:80`
identically — this PR did not introduce it and moving Serilog's configuration ahead of the
guards is a change to startup ordering in five services. It is also not fully solvable in place:
the "config file not found" case throws before any configuration exists to build a logger from.
Worth an issue, alongside the working-directory half of the same problem already tracked on #105.

**4. No remediation path for a registration whose wallet creation failed.** Agreed that the need
is real, and agreed the endpoint removed here was not serving it — it returned 500 to every
caller. The substantive defect is in `UserService.CreateDefaultWalletsAsync`: it catches
everything, reports via `Console.WriteLine` while holding an injected `_logger` used on the line
above, and returns a bool both callers discard. That is a behavioral change in the registration
path, in a different file, and #190 asked for neither. Worth an issue.

**5. No test pins the positional change.** Accurate. The guard's contract is now well covered,
but nothing asserts that the call sits outside the `AddHttpClient` delegate — move it back and
all 732 still pass. Pinning it needs a host-level startup test, and this solution has no
`WebApplicationFactory` harness to hang one on; adding the first one is its own piece of work.
The behaviour is hand-verified above and in the PR body. Recording the limitation rather than
papering over it.

**7. `RequireConnectionString` is still read inside the `AddDbContext` delegate.** Correct, and a
good catch that it works today only because the migration scope resolves the context at startup.
But it is pre-existing code that #190 did not ask about, and `STANDARDS.md` is explicit that
adjacent code is flagged, not fixed inline. Belongs with the follow-up that migrates the other
call sites.

**8. `RequireUri` duplicates `RequireConnectionString`'s guard clauses.** True — about six lines.
Collapsing them means rewriting the existing method, which is the same scope rule as finding 7.
With two methods the drift risk is small, and the follow-up issue that moves the four remaining
fallback sites onto this helper is the right moment to extract the common core, when there is
more than one caller to justify the shape.

### Follow-up issues this suggests

1. `Wallet.Api:309` — a `MapGet` that creates wallets (already raised in the PR body)
2. Four remaining configuration fallbacks on unused ports (already raised in the PR body)
3. Configuration guards throw before logging is configured (finding 3)
4. `CreateDefaultWalletsAsync` swallows failures and discards its result (finding 4)

Happy to open all four, or any subset — say which.

MohKardan:
Follow-ups from this PR are now filed, grouped by domain rather than one per finding:

- **#205** — service addresses fall back to hardcoded literals instead of failing at startup. Collects the five remaining fallback literals, the `Orders.Api` → `UsersApiUrl` scan that resolves by enumeration order, the guards evaluated inside DI delegates, and the guards throwing before Serilog is configured. All one mechanism, one fix.
- **#206** — default wallet creation at registration: the `MapGet` that mutates, the caller that swallows every failure and reports it via `Console.WriteLine` while holding an injected logger, and the remediation path this PR's removal leaves unowned. All one flow.

Findings 3, 4, 5, 7 and 8 from the self-review above are covered across those two.

PR #207, opened 2026-09-02
[Feat] Fail CI when a document links to a repository path that does not exist — issue #196

### LINKED ISSUE #196: CI: fail when a tracked document references a repository path that does not exist
## The problem

Five consecutive PRs — #189, #192, #193, #194, #195 — have deleted tracked text that describes code the repository no longer has. Between them: 13 stale documents, a folder of architecture diagrams, four probe scripts for a deleted API, a generated code map, and six unrunnable root files.

Every one of them was found by a hand-run `grep`, and **every one of those greps leaked**:

| PR | How it was verified | What it missed |
|---|---|---|
| #189 | `grep --include=*.cs` | Four `.ps1`/`.txt` files calling the deleted endpoint — the PR body claimed they were already gone |
| #193 | Fixed those two extensions by naming four files, then verified across `.md .cs .ps1 .yml .csproj` | `.dgml`, `.bat`, `.cmd`, `.http`, `.sql`, `.json` — six more files of the same class |
| #195 | Extension-agnostic at last (`git ls-files -z \| xargs -0 grep`) | Nothing in its own scope, but it searches for *deleted filenames*, so by construction it cannot find docs describing deleted **code** |

Each PR diagnoses why the previous one leaked and then leaks somewhere else. That is a sign the fix belongs in CI, not in a better one-off command.

## What the check should do

Fail the build when a tracked text file references a repository path that does not exist.

Two reference forms cover almost all of the observed drift:

1. **Markdown links** — `[text](docs/process/STANDARDS.md)`, relative to the containing file. This is what caught the one break currently in the tree (`docs/pull-requests/PR_TASK-003_QUARANTINE_STUBS.md` → `docs/process/SPRINT_PLAN.md`).
2. **Backticked repo paths** — `` `src/Order/Orders.Api/Program.cs` ``, `` `run.bat` ``, `` `scripts/windows-services/publish-all.ps1` ``. Restrict to strings that look like paths (contain `/` or a known root-level filename, and end in a real extension) so prose is not flagged.

A prototype of form 1 already exists as a throwaway script used to verify #189–#195; it walks every `.md`, resolves each relative link, and reports misses. It found exactly one break across 42 files. Promoting that to a committed script plus a workflow step is most of the work.

## Scope

- Run over every tracked file, discovered with `git ls-files` — **never an `--include=*.ext` list**. That is the specific mistake this issue exists to prevent.
- Add as a step in `.github/workflows/build-and-test.yml`, or a small sibling workflow. It needs no SDK, no database, and no build; it should take seconds.
- Exit non-zero with the offending file, line and target.

## Known exclusions

- `docs/audit/AUDIT_*.md`, `docs/audit/METHODOLOGY_v*.md`, `docs/pull-requests/*.md` and `governance/**` are archives. They are historical records and are deliberately never edited, so they will always reference things that have since moved. Either exclude those paths or allow an inline opt-out marker.
- `CHANGELOG.md` names files in dated entries as a record of what happened at that version (`Created API_DOCUMENTATION.md`, `Updated run.bat`). Those are prose, not links, and stay accurate as history — form 2 must not flag them. Excluding `CHANGELOG.md` outright is the simplest answer.

## What this does not solve

A document can be false without naming a missing path — `AGENT.md` listed three admin commands that never existed, and `README.md` advertised an endpoint by URL rather than by file path. Neither would be caught. This closes the *dangling path* class, which is the one that has recurred five times, not documentation drift in general.

## Acceptance

- A PR that deletes a tracked file still referenced by a link in a non-archive document fails CI.
- The check runs on every push and pull request, and passes on `main` as it stands today, once the one known break in `docs/pull-requests/` is either fixed or excluded as an archive.

MohKardan:
**Branch Name**: `chore/ci-check-dangling-doc-paths`
**Linked Issue**: #196

---

## Description

Adds `scripts/check-doc-paths.sh` and runs it as the first step of `build-and-test.yml`. It fails the build when a tracked text file contains a Markdown link to a repository path that does not exist, reporting file, line and target.

Five consecutive PRs (#189, #192, #193, #194, #195) deleted text describing code the repository no longer had, and every one of them was verified by a hand-run `grep` that leaked. Three leaked identically — the grep named the extensions to search and an extension was forgotten each time (first `.ps1`/`.txt`, then `.dgml`, `.bat`, `.cmd`, `.http`, `.sql`, `.json`). So the script names no extension anywhere: files come from `git ls-files`, and a file is skipped only because of the bytes inside it.

---

## Type of Change
- [x] Feature (new capability) — CI check
- [ ] Hotfix / Refactor / Performance / Documentation / Security

---

## Changes Made

- **`scripts/check-doc-paths.sh`** (new). Walks every tracked file, resolves each Markdown link relative to the file containing it, and exits non-zero with `file:line: target -> resolved`.
  - File list is `git ls-files -z`. No extension list, anywhere.
  - Binary files are identified by content — one `git grep --perl-regexp '[\x00-\x08\x0B\x0E-\x1F]'` classifies the whole repo. Skipped files are **printed**, not silently dropped, so nothing goes unchecked quietly.
  - The known-path set is built from git, not the filesystem, so a developer machine and the runner agree: an untracked local file cannot make a broken reference look fine.
  - Handles relative and root-relative targets, `.`/`..`, directory targets, `<angle brackets>`, `"Title"` suffixes, `#fragment`/`?query`, external URL schemes, template placeholders, and percent-escapes.
  - Fenced code blocks are skipped: a fence turns inline markup back into literal text, so a link inside one is an example, not a live reference.
- **`.github/workflows/build-and-test.yml`**: runs the script as the first step, before `setup-dotnet`.
- **`.gitattributes`**: `*.sh text eol=lf`, so a Windows checkout does not give bash a CRLF shebang.
- **`docs/process/STANDARDS.md`**: §8 CI now describes the check, per §9 ("updated when adding a tool").

---

### Form 2 (backticked repo paths) is deliberately **not** implemented

I prototyped it and measured. Restricted as the issue suggested (path-shaped charset, real-looking extension, not a directory, resolved both relative to the containing file and to the repo root), it produced:

```
TOTAL CANDIDATE FAILURES: 113   (across 18 files)
  ...of which 38 still fail when restricted to candidates containing "/"
```

**None of the 113 was a real break.** They fall into these classes:

| Class | Examples |
|---|---|
| Bare filenames used as shorthand | `` `Program.cs` ``, `` `install-services.ps1` ``, `` `WalletService.cs` `` |
| Naming *examples*, not references | `` `WalletServiceTests.cs` ``, `` `ADR-001-microservices-with-database-per-service.md` ``, `` `release/v1.0.0` `` |
| Paths the doc explicitly says do not exist yet | `docs/process/INDEX.md` §"Planned (not yet created)" — `` `docs/design/API_CONTRACT.md` ``, `` `docs/operations/RUNBOOK.md` `` |
| History of renamed files, correct as written | `docs/audit/README.md` — `` `docs/operations/AUDIT_FINDINGS.md` `` (this is the `CHANGELOG.md` class the issue predicted) |
| Project/assembly names, not paths | `` `Affiliate.Api` ``, `` `Orders.Core` ``, `` `TallaEgg.Api` `` |
| Hostnames and executables | `` `api.telegram.org` ``, `` `dotnet.exe` ``, `` `sc.exe` `` |
| Elided paths | `` `TelegramBot/.../BotHandler.cs` `` |
| Extensions alone | `` `.md` ``, `` `.html` `` |
| C# expressions in code comments | `` `u.Id` ``, `` `Orders.Amount` `` |
| Deliberately untracked config | `` `config/appsettings.global.json` `` |

113 false positives and zero true positives is the check people learn to ignore. Form 1 alone is silent and correct, and it is what caught the only real break. Form 2 can be revisited if a break ever escapes form 1.

### `CHANGELOG.md` — no exclusion added

`CHANGELOG.md` was deleted in #198 and no longer exists. Its exclusion rationale was specifically that form 2 must not flag dated prose entries (`Created API_DOCUMENTATION.md`), and form 2 is not implemented — so the exclusion would be doubly dead configuration. If the file ever returns *and* form 2 is ever added, the reasoning still holds and the exclusion should be added then, not now.

### The known break — excluded, not fixed

`docs/pull-requests/PR_TASK-003_QUARANTINE_STUBS.md:159` links to `docs/process/SPRINT_PLAN.md`, deleted in #197. It is an archived PR record and `docs/pull-requests/` is never edited, so it is excluded rather than corrected.

Running with the exclusions removed shows the exclusion is load-bearing, and that there were **three** such breaks, not one:

```
docs/audit/AUDIT_2026-07.md:11: ../process/SPRINT_PLAN.md -> docs/process/SPRINT_PLAN.md (no such path in the repository)
docs/audit/AUDIT_2026-07.md:147: ../process/SPRINT_PLAN.md -> docs/process/SPRINT_PLAN.md (no such path in the repository)
docs/pull-requests/PR_TASK-003_QUARANTINE_STUBS.md:159: docs/process/SPRINT_PLAN.md -> docs/pull-requests/docs/process/SPRINT_PLAN.md (no such path in the repository)
```

Line 266 of that file (`- Related to SPRINT_PLAN.md (Sprint 1, Week 1)`) is bare prose with no link and no backticks, so neither form would ever have caught it.

### Exclusion by path, not by inline marker

"This file is frozen" is a property of the file, and an inline opt-out marker could only be added by editing the very files the rule says never to edit. `AUDIT_*` and `METHODOLOGY_v*` are matched **without** an extension on purpose — `docs/audit/AUDIT_2026-07.html` is the same archived report as its `.md`, and matching only `.md` would have let it through. That is the same class of mistake as an `--include` list.

### A step in `build-and-test.yml`, not a sibling workflow

The `main` ruleset requires the check named `test`. A sibling workflow would create a check nothing requires — green or red, nobody would be stopped by it — and would need a repository-settings change I should not make myself. As a step in the existing `test` job it is required from the first run, and **no repository settings need to change.**

It runs *before* `setup-dotnet` because it needs no SDK, no database and no build, so a broken link fails in about a second instead of after a full Release build.

### Bash rather than PowerShell

The repo's other scripts are `.ps1`, but CI runs `ubuntu-latest` and this machine has no `pwsh` — a PowerShell script could only be tested against Windows PowerShell 5.1 locally and would run on PowerShell 7/Linux in CI, so local proof would not transfer. Bash is the same interpreter in both places, and ships with Git for Windows.

---

### (a) Passes on `main` as it stands

```
$ bash scripts/check-doc-paths.sh
check-doc-paths: skipped 2 file(s) whose bytes are not text:
  docs/business/business-proposal-v1.6.0.pdf
  src/Order/Orders.Infrastructure/Configurations/TradeConfiguration.cs
check-doc-paths: 380 tracked text file(s) checked, every reference resolves.
EXIT=0
```

`TradeConfiguration.cs` is UTF-16LE — a real text file that every NUL-byte heuristic (git's included) calls binary. It is listed rather than silently dropped; a `.cs` file cannot carry a Markdown link, so nothing is lost, but the gap is visible instead of hidden.

The check also caught **itself** on the first run: an explanatory comment in the script contained a literal Markdown link. Rewording it was the fix. That is proof the check covers every tracked text file, not only `.md`.

### (b) It actually fails — with file, line and target

Broken link added to a live (non-archive) document:

```
$ printf '\nSee [the deployment runbook](../operations/DOES_NOT_EXIST.md) for details.\n' >> docs/process/WORKFLOW.md
$ bash scripts/check-doc-paths.sh
docs/process/WORKFLOW.md:91: ../operations/DOES_NOT_EXIST.md -> docs/operations/DOES_NOT_EXIST.md (no such path in the repository)
check-doc-paths: 1 broken reference(s) across 380 tracked text file(s).
Fix the reference, or -- if the file is an archive that is never edited -- add its
path to EXCLUDED_GLOBS in scripts/check-doc-paths.sh.
EXIT=1
```

Reverted afterwards.

### (c) The exclusions work

The identical broken link, appended to an archived file instead:

```
$ printf '\nSee [the deployment runbook](../operations/DOES_NOT_EXIST.md) for details.\n' >> docs/pull-requests/PR_TASK-003_QUARANTINE_STUBS.md
$ bash scripts/check-doc-paths.sh
check-doc-paths: 380 tracked text file(s) checked, every reference resolves.
EXIT=0
```

Reverted afterwards.

### Edge cases spot-checked

Appended to `docs/process/WORKFLOW.md`, run, reverted — only the one intended failure was reported:

| Input | Result |
|---|---|
| `[a dir that exists](../../scripts/windows-services)` | passes (directories resolve) |
| `[a dir that does not](../../scripts/nope)` | **flagged** |
| `[anchor only](#section)` | skipped |
| `[external](https://example.com/x.md)` | skipped |
| `[with title](../../AGENT.md "The guide")` | passes (title stripped) |
| link inside a ```` ```markdown ```` fence | skipped |
| identical link outside the fence | **flagged** |

### Unit tests

```
$ dotnet build TallaEgg.sln
Build succeeded.
    0 Warning(s)
    0 Error(s)

$ dotnet test TallaEgg.sln --no-build
Passed!  - Failed: 0, Passed: 732, Skipped: 0, Total: 732, Duration: 9 s
```

### End-to-end
- [x] N/A — no runtime behaviour changes. This adds a CI script and touches no service, handler or migration.

### Environment
- [x] Tested locally (Windows, Git Bash). Runtime ~8s locally; faster on the runner, which does not pay Cygwin's fork cost.

---

## Security Checklist
- [x] No hardcoded secrets, passwords, API keys or tokens
- [x] `config/appsettings.global.json` untouched and still untracked
- [x] No personal names, emails, handles or paths introduced

---

## Code Quality
- [x] Comments in English, explaining "why"
- [x] No new dependencies — `git`, `bash`, nothing else
- [x] Scope kept narrow: the check, its CI step, the `.gitattributes` line it needs to run on Windows, and the `STANDARDS.md` §8 entry that §9 requires

The check found no broken links in live documents, so no document fixes were needed in this PR.

---

## Breaking Changes?
- [x] No breaking changes

---

## What this does not solve

Carried over from the issue, unchanged: a document can be false without naming a missing path. `AGENT.md` once listed three admin commands that never existed; `README.md` advertised an endpoint by URL. Neither would be caught. This closes the *dangling path* class — the one that has now recurred five times — not documentation drift in general.

Known, accepted gaps, all visible rather than silent:
- Reference-style links (`[text][ref]` with a `[ref]:` definition) and HTML `href=`/`src=` are not parsed. Verified there are none in live files today.
- Links inside fenced code blocks are skipped by design.
- UTF-16 and binary files are skipped, and printed on every run.

---

## Deployment Notes

**Pre-Deployment**: none. No migration, no configuration, no service change.

**Rollback Plan**: revert the commit. The only effect is that CI stops running the check.

---

## Related Issues
- Closes #196
- Follows #189, #192, #193, #194, #195 (the five leaking hand-run greps)
- The excluded break points at `SPRINT_PLAN.md`, deleted in #197
- `CHANGELOG.md`, whose exclusion the issue discussed, was deleted in #198

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,6 @@
 # Auto detect text files and perform LF normalization
 * text=auto
+
+# Shell scripts must keep LF endings: bash reads a CRLF shebang line literally and fails with
+# "bad interpreter", so a Windows checkout would break scripts/check-doc-paths.sh (issue #196).
+*.sh text eol=lf
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -34,6 +34,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # Before the SDK, because it needs none and finishes in about a second. Five consecutive
+      # PRs (#189–#195) deleted text describing code that was already gone, and every one of
+      # them was verified by a hand-run grep that leaked; #196 moved that verification here.
+      #
+      # A step in this job rather than a sibling workflow: the main ruleset requires the check
+      # named `test`, so a second workflow would add a check nothing requires — green or red,
+      # nobody would be stopped by it.
+      - name: Check documents for dangling repository paths
+        run: bash scripts/check-doc-paths.sh
+
       - uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '9.0.x'
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -34,16 +34,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # Before the SDK, because it needs none and finishes in about a second. Five consecutive
-      # PRs (#189–#195) deleted text describing code that was already gone, and every one of
-      # them was verified by a hand-run grep that leaked; #196 moved that verification here.
-      #
-      # A step in this job rather than a sibling workflow: the main ruleset requires the check
-      # named `test`, so a second workflow would add a check nothing requires — green or red,
-      # nobody would be stopped by it.
-      - name: Check documents for dangling repository paths
-        run: bash scripts/check-doc-paths.sh
-
       - uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '9.0.x'
@@ -86,3 +76,25 @@ jobs:
           name: test-results
           path: test-results/
           retention-days: 14
+
+      # Five consecutive PRs (#189–#195) deleted text describing code that was already gone, and
+      # every one of them was verified by a hand-run grep that leaked; #196 moved that
+      # verification here. Needs no SDK and takes seconds.
+      #
+      # A step in this job rather than a sibling workflow: the main ruleset requires the check
+      # named `test`, so a second workflow would add a check nothing requires — green or red,
+      # nobody would be stopped by it.
+      #
+      # Last, with `if: always()`, rather than first. It is cheap enough to run anywhere, but a
+      # PR whose only fault is a stale link should still get its test results: running this ahead
+      # of the build would skip Restore/Build/Test and leave the author knowing nothing about
+      # whether their actual change works. `always()` keeps it a gate even when the tests failed.
+      #
+      # --self-test first: this check is 300 lines of hand-rolled Markdown parsing, and the whole
+      # premise of #196 is that verifying by hand leaks. So the parser has fixtures, and they run
+      # here rather than only on the author's machine.
+      - name: Check documents for dangling repository paths
+        if: always()
+        run: |
+          bash scripts/check-doc-paths.sh --self-test
+          bash scripts/check-doc-paths.sh
--- a/docs/process/STANDARDS.md
+++ b/docs/process/STANDARDS.md
@@ -342,10 +342,13 @@ Example (BAD):
 - `.github/workflows/build-and-test.yml` runs `dotnet build` and `dotnet test TallaEgg.sln` on
   every pull request, and on pushes to `main`. It is the `test` check the `main` ruleset requires.
   The admin role bypasses every rule, so treat it as a gate you keep, not one that holds you.
-  Its first step is `scripts/check-doc-paths.sh`, which fails the build when a tracked text file
+  It also runs `scripts/check-doc-paths.sh`, which fails the build when a tracked text file
   contains a Markdown link to a repository path that does not exist. Run it locally the same way
-  — it needs no SDK. Archives under `docs/audit/`, `docs/pull-requests/` and `governance/` are
-  exempt; if you add another frozen archive, add its path to `EXCLUDED_GLOBS` in the script.
+  — it needs no SDK — and `scripts/check-doc-paths.sh --self-test` to exercise its parser.
+  Exempt as frozen archives: `docs/audit/AUDIT_*`, `docs/audit/METHODOLOGY_v*`,
+  `docs/pull-requests/*` and `governance/*`. Note `docs/audit/README.md` is *not* exempt — it is
+  a living index. If you add another frozen archive, add its path to `EXCLUDED_GLOBS` in the
+  script.
 - `.github/workflows/manual-test-run.yml` stands Users/Wallet/Orders and the bot up against a
   throwaway SQL Server so a human can drive it over real Telegram. It asserts nothing and needs
   the `TELEGRAM_BOT_TOKEN` and `OWNER_TELEGRAM_ID` secrets.
--- a/docs/process/STANDARDS.md
+++ b/docs/process/STANDARDS.md
@@ -342,6 +342,10 @@ Example (BAD):
 - `.github/workflows/build-and-test.yml` runs `dotnet build` and `dotnet test TallaEgg.sln` on
   every pull request, and on pushes to `main`. It is the `test` check the `main` ruleset requires.
   The admin role bypasses every rule, so treat it as a gate you keep, not one that holds you.
+  Its first step is `scripts/check-doc-paths.sh`, which fails the build when a tracked text file
+  contains a Markdown link to a repository path that does not exist. Run it locally the same way
+  — it needs no SDK. Archives under `docs/audit/`, `docs/pull-requests/` and `governance/` are
+  exempt; if you add another frozen archive, add its path to `EXCLUDED_GLOBS` in the script.
 - `.github/workflows/manual-test-run.yml` stands Users/Wallet/Orders and the bot up against a
   throwaway SQL Server so a human can drive it over real Telegram. It asserts nothing and needs
   the `TELEGRAM_BOT_TOKEN` and `OWNER_TELEGRAM_ID` secrets.
--- a/scripts/check-doc-paths.sh
+++ b/scripts/check-doc-paths.sh
@@ -1,0 +1,215 @@
+#!/usr/bin/env bash
+#
+# Fails when a tracked text file references a repository path that does not exist.
+#
+# Five consecutive pull requests (#189, #192, #193, #194, #195) deleted text describing code
+# the repository no longer had, and each was verified by a hand-run grep that leaked. Three of
+# them leaked the same way: the grep named the extensions to search, and an extension was
+# forgotten every time -- first .ps1 and .txt, then .dgml, .bat, .cmd, .http, .sql and .json.
+#
+# So this script names no extension anywhere. The file list is `git ls-files`, and a file is
+# skipped only because of what is inside it, never because of what it is called -- and the
+# skipped ones are printed, so nothing goes unchecked quietly. If you are about to add an
+# extension list to this file, you are re-introducing the bug it exists to prevent (issue #196).
+#
+# Usage:  scripts/check-doc-paths.sh
+# Exit:   0 every reference resolves, 1 at least one does not, 2 the script could not run.
+
+set -uo pipefail
+
+repo_root=$(git rev-parse --show-toplevel 2>/dev/null) || {
+    printf 'check-doc-paths: not inside a git repository\n' >&2
+    exit 2
+}
+cd "$repo_root" || exit 2
+
+# Archives: historical records that are deliberately never edited, so they will always name
+# things that have since moved. Whole paths are excluded rather than individual lines marked,
+# because "this file is frozen" is a property of the file, and an inline opt-out marker could
+# only be added by editing the very files the rule says not to edit.
+#
+# AUDIT_* and METHODOLOGY_v* carry no extension on purpose: docs/audit/AUDIT_2026-07.html is
+# the same archived report as AUDIT_2026-07.md, and matching only .md would have let it through.
+EXCLUDED_GLOBS=(
+    'docs/audit/AUDIT_*'
+    'docs/audit/METHODOLOGY_v*'
+    'docs/pull-requests/*'
+    'governance/*'
+)
+
+is_excluded() {
+    local path=$1 glob
+    for glob in "${EXCLUDED_GLOBS[@]}"; do
+        # shellcheck disable=SC2053 -- the right-hand side is a glob on purpose
+        [[ $path == $glob ]] && return 0
+    done
+    return 1
+}
+
+# Files holding a C0 control byte that text does not use. That is a property of the bytes, so a
+# new binary format needs no change here, and one `git grep` classifies the whole repository.
+# Tab, newline, carriage return and form feed are excluded from the set as legitimate in text.
+declare -A IS_BINARY=()
+while IFS= read -r path; do
+    [[ -n $path ]] && IS_BINARY[$path]=1
+done < <(git grep --files-with-matches --perl-regexp '[\x00-\x08\x0B\x0E-\x1F]' -- . 2>/dev/null)
+
+# Every tracked file plus every directory that contains one, so a link to a folder resolves too.
+# Built from git rather than from the filesystem so the answer is the same on a developer
+# machine and on a runner: an untracked local file must not make a broken reference look fine.
+declare -A KNOWN_PATHS=()
+while IFS= read -r -d '' tracked; do
+    KNOWN_PATHS[$tracked]=1
+    dir=$tracked
+    while [[ $dir == */* ]]; do
+        dir=${dir%/*}
+        KNOWN_PATHS[$dir]=1
+    done
+done < <(git ls-files -z)
+
+if ((${#KNOWN_PATHS[@]} == 0)); then
+    printf 'check-doc-paths: git ls-files returned nothing\n' >&2
+    exit 2
+fi
+
+# The helpers below assign to a global instead of printing, because a command substitution forks
+# a subshell and this runs once per link across the whole repository.
+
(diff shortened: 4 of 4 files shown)


MohKardan:
## Self-review response (`/code-review 207`, extra-high effort)

15 findings. **11 fixed** in e4e02f1, **1 partly**, **3 declined with reasons**.

The review's central point landed: this PR's premise is that hand-run verification leaks, and the checker was itself verified only by hand. Four of the parser defects were reachable from a ten-line fixture and none of my manual experiments found them. `--self-test` now ships with the script and runs in CI.

---

### Fixed — defects that could let a broken link pass, or turn the check off

**1. Fence tracking desynchronised on nested or mismatched fences.** Confirmed. A single in/out toggle ignored the delimiter's character and length, so a ```` ```` ````-delimited block quoting a ``` ``` ``` block closed early (links in the outer block reported), and an odd delimiter count left `in_fence=1` for the rest of the file — silently disabling the gate for that file, which is precisely the failure class #196 exists to remove. Now follows CommonMark: the closing fence must use the same character and be at least as long. Six fixtures cover it.

**2. `printf '%b'` interpreted backslashes already in the destination.** Confirmed, and the worst of the set: `docs\c%20x.md` decoded to `docs`, because `\c` means "produce no further output". `docs` is in `KNOWN_PATHS`, so a broken link **passed**. Backslashes are now doubled before the substitution, and `%00` is rejected outright. Fixture asserts `docs\c%20x.md` → `docs\c x.md`; removing the escaping step reproduces `docs` and turns the self-test red.

**3. Globbing in `normalize_path`.** Confirmed — `normalize_path 'docs/*'` returned a `/`-joined listing of the repository root. `set -f` is now saved and restored around the split. Two fixtures (`*` and `[...]`).

**4. Inline code spans were scanned.** Confirmed. Now stripped before link extraction, which is also CommonMark-correct — a destination inside backticks renders as literal text. A link whose *label* is code keeps its destination, which is the dominant style in this repo's docs; there is a fixture for exactly that.

**5. `git grep` status discarded.** Fixed — non-{0,1} status now exits 2 with the status.

One correction to that finding's reasoning: it claimed the run "looks identical to a healthy one". It would not — with `IS_BINARY` empty the PDF is scanned as prose and emits 31 bogus failures, so CI goes red. The finding is still right that the failure mode was bad and the cause unreported; failing with a clear message beats failing with 31 lines of mojibake.

**7. Excluded files were invisible.** Fixed — the run now prints `excluded 28 archived file(s) by path (...)` alongside the binary-skip block. The header comment promised no silent skips and exclusions broke that promise.

**8. `STANDARDS.md` overstated the `docs/audit/` exemption.** Fixed. It now names the two prefixes and says explicitly that `docs/audit/README.md` is *not* exempt. Fair catch on a PR about documents being accurate.

**14. Step placement suppressed the test signal.** Accepted. Moved after `Test (Release)` with `if: always()`. The argument is right: running it first costs a lost test signal on every doc typo, running it last costs one build that was going to run anyway.

**15. `$0` in the remediation message; split streams.** Fixed — literal `scripts/check-doc-paths.sh`, and all failure output now goes to stderr.

**4 (list order) — unreadable tracked file counted as checked.** Fixed — now exits 2 rather than reporting a pass over a file it never read.

**11. UTF-16 files skipped.** Fixed properly rather than documented. The finding is right that this is reachable: PowerShell 5.1 writes UTF-16LE by default and that is almost certainly how `TradeConfiguration.cs` became UTF-16, so a doc written with `Out-File docs/foo.md` would have had every link ignored. Files with a UTF-16 BOM are now decoded through `iconv` instead of skipped. Scanned count went 380 → 381; only the PDF is skipped now.

**9 (partly).** Inline code spans fixed. HTML comments are *not* stripped — that needs multi-line state for a construct this repo's live docs do not contain. Listed as a known gap instead.

---

### Declined

**6. Narrow `governance/*`.** The observation is fair — `governance/proposals/proposal-template-*.md` exists to be copied and edited, and `governance/archive/` implies the rest is not archive. But #196 names `governance/**` in its exclusion list explicitly, and whether governance documents are frozen is the owner's call about their own repository, not something I should quietly re-scope inside an implementation PR. Verified the finding's factual claim: governance contributes **zero** failures today, so narrowing or dropping the glob would be a no-op right now and can be done whenever the owner wants, with no rework.

@MohKardan — worth a look. If you want `governance/*` narrowed to `governance/archive/*` and `governance/meeting-notes/*`, say so and it is a one-line change.

**10. Pre-filter with `git grep -l --fixed-strings ']('`.** Declined, and this one matters. It would be ~12x faster and the semantics are sound today — but it adds a *second* place that decides which files get scanned, which must stay in agreement with the link-extraction regex forever. That is the same shape as the `--include=*.ext` list that leaked three times: not the same mechanism, but the same defect — a filter that silently narrows the file set and drifts out of sync with what is actually being searched. The header comment now says so in as many words. 13s locally on Windows (Cygwin fork tax); the runner will be well under, and the step no longer gates the build.

**12. `ALLOWED_UNTRACKED_PATHS` for `config/appsettings.global.json`.** Declined for the same reason this PR declined the `CHANGELOG.md` exclusion: no document links to it today, so it would be dead configuration added for a hypothetical. The trap is real — if someone does link it, both remedies the script suggests are wrong — but the honest fix at that point is one line, and adding it now on speculation is the thing `STANDARDS.md` §3 warns against. Noting it here so the reasoning is on the record if it ever fires.

**13. Off-the-shelf checker instead of a hand-rolled parser.** Partly accepted: the test half is done, the dependency half is not. A Node/Python link checker would handle fences, code spans, reference-style links and HTML `href` correctly by construction — but it would put a package manager and a lockfile into a .NET repo's required check, for a repository of 40 Markdown files. The fixtures address the actual risk the finding identifies. If the parser needs a fifth edge case, that trade should be revisited.

---

```
$ bash scripts/check-doc-paths.sh --self-test
check-doc-paths --self-test: 30 assertions passed.

$ bash scripts/check-doc-paths.sh
check-doc-paths: skipped 1 file(s) whose bytes are not text:
  docs/business/business-proposal-v1.6.0.pdf
check-doc-paths: excluded 28 archived file(s) by path (docs/audit/AUDIT_* docs/audit/METHODOLOGY_v* docs/pull-requests/* governance/*).
check-doc-paths: 381 tracked text file(s) checked, every reference resolves.
EXIT=0
```

**The self-test earns its place** — each parser bug reintroduced one at a time, into a copy:

| Bug reintroduced | Self-test result |
|---|---|
| globbing in `normalize_path` | `FAIL: normalize: glob is not expanded` (29 passed, 1 failed) |
| backslash escaping removed | `FAIL: clean: backslash survives decode` — expected `docs\c x.md`, actual `docs` |
| code-span stripping disabled | 2 failures |
| fence char/length matching disabled | 4 failures |

Proofs (a), (b) and (c) from the PR body were re-run against the revised script and all still hold — same file, same line number, same exit codes.

```
$ dotnet build TallaEgg.sln   → Build succeeded. 0 Warning(s), 0 Error(s)
$ dotnet test TallaEgg.sln    → Passed! Failed: 0, Passed: 732, Skipped: 0, Total: 732
```

One more self-catch worth recording: after adding the fixtures, the check failed on **its own self-test data**, because fixture strings like `[x](y.md)` are real Markdown links to a real scanner. Fixtures now spell the destination as two adjacent literals so the file contains no link. Second time the check has caught its own source — which is the evidence that it reads every tracked text file, not just `.md`.

Issue #206, opened 2026-09-02
Default wallet creation at registration: a GET that mutates, and a failure path that records nothing


MohKardan:
Found while doing #190 and reviewing its PR (#204). Three defects in one flow — the creation of a
user's default wallets at registration — so one issue rather than three.

#190 removed `POST /api/user/{userId}/create-default-wallets`, which had never worked and had no
caller. That was correct, and it leaves this flow as the only path. It has problems of its own.

## The flow

```
POST /api/user/register                     Users.Api
        ↓
UserService.RegisterUserAsync               UserService.cs:53, :163
        ↓
UserService.CreateDefaultWalletsAsync       UserService.cs:202  (private)
        ↓ named "WalletAPI" client
GET api/wallet/create-default/{userId}      UserService.cs:209
        ↓
MapGet /api/wallet/create-default/{userId}  Wallet.Api/Program.cs:309
        ↓
WalletService.CreateDefaultWalletsAsync     → Toman, MAUA, CREDIT_MAUA
```

It works. All three problems below are about what happens when it does not.

## 1. A `MapGet` that creates wallets

`src/Wallet/Wallet.Api/Program.cs:309` maps wallet creation to **GET**. A safe verb that mutates
state.

Risk today is low and this is not urgent: the APIs bind loopback only and sit behind an API key,
so there is no browser, crawler, or proxy in a position to prefetch it. `CreateWalletAsync` is
also idempotent, so a repeat call is harmless. But GET is the verb every intermediary in the world
assumes it may retry or prefetch freely, and this one is now the only route into wallet creation.

Changing it is a breaking change for exactly one caller — `UserService.cs:209` — which is in this
repository. Do both in one commit.

## 2. The caller swallows every failure, and reports it where nobody is listening

`src/User/Users.Application/UserService.cs:202-219`:

```csharp
catch (Exception ex)
{
    // Log the error but don't fail user creation
    Console.WriteLine($"خطا در ایجاد کیف پول‌های پیش‌فرض برای کاربر {userId}: {ex.Message}");
    return false;
}
```

Three things:

- **`Console.WriteLine`, with an injected `ILogger` in scope** — `_logger` is used on the line
  above for the success path. Under `sc.exe` (`UseWindowsService()`, #70) a service has no
  console, so this failure is written to a handle nobody holds. Same class of problem as #202.
- **`catch (Exception)`** — swallows the transport failure, a non-2xx response, and a
  misconfigured address alike.
[...]

Issue #205, opened 2026-09-02
Service addresses fall back to hardcoded literals instead of failing at startup, in six places


MohKardan:
Found while doing #190 and reviewing its PR (#204). #190 fixed one instance; these are the rest
of the same pattern, collected into one issue because they share a mechanism and a fix.

`AGENT.md` states the rule: *"Missing configuration fails at startup rather than falling back to
a default."* Six places do the opposite. None is a live defect today — configuration currently
supplies every value — which is exactly why they have survived: the fallback branch is never
taken, so nothing marks these lines as wrong, and no audit has flagged them.

They are not dead code. Each line runs at every startup. Only the right-hand side of the `??`
is unreached, and only for as long as the configuration stays complete.

## 1. Fallback literals on ports this system does not use

| Location | Fallback | Note |
|---|---|---|
| `src/Order/Orders.Api/Program.cs:120` | `http://localhost:60933` | correct port, still a fallback |
| `src/TallaEgg/TallaEgg.Infrastructure/Clients/UsersApiClient.cs:26` | `http://localhost:5001/api` | 5001 is not used anywhere in this system |
| `src/TallaEgg/TallaEgg.Infrastructure/Clients/WalletApiClient.cs:43` | `http://localhost:60933/api` | string constructor |
| `src/TallaEgg/TallaEgg.Infrastructure/Clients/WalletApiClient.cs:51` | `http://localhost:60933/api` | `IConfiguration` constructor |
| `TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/OrderApiClient.cs:26` | `http://localhost:5135/api` | 5135 is `TallaEgg.Api`, which maps zero endpoints; orders are on 5140 |

The last two are the interesting ones. If either fallback were ever taken, the caller would
silently talk to a service that answers nothing — 5001 to nothing at all, 5135 to a host that
starts, binds, and maps no routes. That is the failure mode `ConfigurationGuard` was introduced
for in #68: not a crash, but a service running against the wrong thing with nothing in the logs.

`ConfigurationGuard.RequireUri` was added in #204 for exactly this and is the intended
replacement. It returns a parsed `Uri`, so a value that is present but unusable — a bare port, a
host with no scheme, a leftover placeholder — fails the same way an absent one does.

## 2. `Orders.Api` resolves its Users address by scanning the whole config file

This one is worse than a dead fallback, because the live branch is the accidental one.

[...]

PR #208, opened 2026-09-03
[Fix] Wallet creation is a POST, and its failures reach the log — issue #206

MohKardan:
## Description

Registration creates a user's default wallets — Toman, MAUA, CREDIT_MAUA — by calling
`Wallet.Api`. #190 removed the only other route into wallet creation, which left this path both
the only one and, in two ways, the least observable one. The endpoint moves from GET to POST,
and the caller now records its failures through the injected `ILogger` instead of writing them
to a console a Windows service does not have.

## Issue/Task Reference

Part of #206 — parts 1 and 2. Part 3 needed a decision rather than code; see
"Part 3" below.

## Type of Change

Bug fix, not urgent: the flow works today, and everything here is about what happens when it
does not.

Confirmed against the tree, not taken from the issue:

| Claim | Confirmed at |
| --- | --- |
| Wallet creation mapped to GET | `src/Wallet/Wallet.Api/Program.cs:309` (before this PR) |
| Exactly one caller, in this repository | `grep -rn "create-default"` returned three hits: the mapping, and the log line and the call in `UserService` |
| `Console.WriteLine` with `_logger` in scope | `UserService.cs:207` logs through `_logger`, `:216` through the console |
| The `bool` is discarded by both call sites | `UserService.cs:54`, `:164` |

One item the issue did not state, confirmed here and worth naming because it is the bigger half
of the problem: **a non-2xx response never reached the `catch` at all.** `return
response.IsSuccessStatusCode` treats a completed request as a completed request — a 404 or a 500
from `Wallet.Api` produced a discarded `false` and no log line whatsoever, whereas an exception
at least reached `Console.WriteLine`. The refusal is also the likelier of the two failures.

Two of the three things the issue's `catch (Exception)` bullet names cannot happen here any
more, which is why the catch stays broad (see below): a non-2xx is now handled before it, and a
misconfigured address stops the service at startup since #204's `ConfigurationGuard.RequireUri`.

## Changes Made

- `src/Wallet/Wallet.Api/Program.cs` — `MapGet` → `MapPost` for
  `/api/wallet/create-default/{userId}`, with a comment saying why. The route and the handler are
  otherwise untouched
- `src/User/Users.Application/UserService.cs` — `GetAsync` → `PostAsync`, in the same commit, as
  the issue asks. Both halves of a breaking change, and the only caller is in this repository
- The non-2xx branch logs at `Error` with the status code and the response body, capped at 500
  characters. (For this particular endpoint that body is currently always the same generic
  message — see #210 — but the status code alone separates 401 from 405 from 500)
- The `catch` logs at `Error` through `_logger` with the exception object, replacing
  `Console.WriteLine` with `ex.Message`. The type matters — it is what tells a refused connection
  from a timeout — and only the exception object carries it
- The `catch` deliberately stays broad. Nothing is rethrown, so narrowing it would turn some
  failures into failed registrations, which is exactly the behaviour the method exists to avoid.
  A comment says so, and the log entry now makes the type visible instead of flattening it away
- Return type `Task<bool>` → `Task`. Both call sites discarded the bool, and with the method
  recording its own failures there is nothing left for a caller to act on
- Nine tests in `tests/TallaEgg.AllServices.Tests/DefaultWalletCreationFailureTests.cs`, and the
  `Users.Application` project reference they need

Registration still succeeds when wallet creation fails. That was deliberate, the issue agrees it
is defensible, and it is unchanged — two of the new tests pin it.

`1ae7b78` applies the self-review findings on this PR's own code: the Error messages no longer
overclaim ("may be missing or incomplete", since a timeout or a partial creation can leave a
user with some wallets), the body is truncated, the response is disposed, and three test
weaknesses are closed. Full accounting in the review comment below.

## Part 3 — no remediation path, by decision

The issue's third part asked whether an operator needs a supported way to repair a user whose
registration-time wallet creation failed, and if so whether that should be an admin bot command
or an endpoint. Asked and answered by the product owner: **no repair surface**. Nothing is built
for it here.

Two things make that the cheap answer rather than a deferral:

- **A wallet-less user largely self-heals.** Since the lazy-creation work behind
  `WalletLazyCreationTests`, every *write* path creates a missing wallet for a valid currency on
  demand — `WalletService.IncreaseBalanceAsync` and `DecreaseBalanceAsync`,
  `WalletRepository.LockBalanceAsync`/`UnlockBalanceAsync` and the settlement path. The surface
  that genuinely breaks is a pure *read*: `WalletService.GetBalanceAsync` throws
  "کیف پول پیدا نشد" when there is no row.
- **The Error entries in part 2 are the "cheap way to find them" the issue asked about.** They
  name the user id, so `logs/users-api-*.log` now answers "which users need looking at", which
  before this PR nothing anywhere did. `WalletRepository.cs:280` still cannot tell a missing
  wallet from a lazily-created one, and that is unchanged — but it no longer has to be the place
  you find out.

Since part 3 closes by decision rather than by code, this PR says "part of #206". **Nothing is
left to build**; #206 can be closed once this merges.

### Build

```
```

### Unit tests

```
dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 741, Skipped: 0, Total: 741
```

Nine new, covering: the exception path logs `Error` carrying the exception; the non-2xx path
logs `Error` carrying status and body; both still register the user; a huge body is truncated
before it is logged; the request is a POST at the expected path, through the configured
`"WalletAPI"` client; a success logs no `Error`; and — from the serving side — `Wallet.Api` maps
the route as `MapPost` and not `MapGet`.

### The new tests were shown to fail

A check that has never been red is not a check. With the two source files stashed and the tests
left in place, the four that pin new behaviour go red and the three that pin the deliberate
unchanged behaviour stay green. (Run at `12b450e`, before self-review added the truncation and
client-name tests; those two cover code that did not exist at all beforehand.)

```
Failed  RegisterUserAsync_CreatesDefaultWallets_WithPostNotGet
        Assert.Equal() Failure: Values differ   Expected: POST   Actual: GET
Failed  RegisterUserAsync_WalletServiceUnreachable_LogsErrorWithTheException
        Assert.Single() Failure: The collection did not contain any matching items
        Collection: [Entry { Level = Information, ... }]
Failed  RegisterUserAsync_WalletServiceReturnsNon2xx_LogsErrorWithStatusAndBody
        Assert.Single() Failure: The collection did not contain any matching items
Failed  WalletApi_MapsDefaultWalletCreation_AsPostNotGet
        Assert.Contains() Failure: Sub-string not found

Failed! - Failed: 4, Passed: 3, Skipped: 0, Total: 7
```

### The normal path still works — the only thing this PR could break

A real registration through `POST /api/user/register` against the running stack, then the wallet
database queried directly for that user id:

```
CREDIT_MAUA  balance=0.00000000  locked=0.00000000  created=2026-09-03 06:00:48
IRT          balance=0.00000000  locked=0.00000000  created=2026-09-03 06:00:47
MAUA         balance=0.00000000  locked=0.00000000  created=2026-09-03 06:00:48
```

All three default wallets, created within a second of the user row, over POST. The probe user was
in the simulator's `TelegramId < 0` range and has since been cleared by `DataReset`.

### Failure is now visible — both paths, against the running stack

**Wallet.Api down** (the exception path). Registration succeeds, and the failure lands in
`src/User/Users.Api/logs/users-api-20260903.log` — the rolling file sink, which is what an
operator reads under `sc.exe` and is exactly what `Console.WriteLine` never reached:

```
2026-09-03 09:31:06.683 +03:30 [ERR] Default wallet creation for user
"e2853c8d-5d2a-4dbe-9b83-f8be0746a717" failed. Registration continues; this user's default
wallets may be missing or incomplete.
System.Net.Http.HttpRequestException: No connection could be made because the target machine
actively refused it. (localhost:60933)
 ---> System.Net.Sockets.SocketException (10061): No connection could be made because the
target machine actively refused it.
```

`registration success = True`, and the user has 0 wallet rows — the failure was real, not
simulated at the logging layer. (Rows checked on the first pass of this scenario, before the
review fixes reworded the message; the second pass re-verified the log line.)

**Answering non-2xx** (the path that skips the catch). The real service stopped and a stub bound
to port 60933 answering 500 to everything, so the request completes and never throws:

```
2026-09-03 09:31:22.465 +03:30 [ERR] Default wallet creation for user
"330feb78-db05-49e0-98e0-de4f6b7ce620" was refused: HTTP 500. Response:
{"success":false,"message":"stub: wallet service refused","data":null}. Registration
continues; this user's default wallets may be missing or incomplete.
```

`registration success = True` here too. Before this PR that registration produced no log line at
all.

### End to end

```
=== Done in 00:00:45.4. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:
  BTC/IRT: 17 filled
  MAUA/IRT: 17 filled
  SEKE_BAHAR/IRT: 16 filled
```

Registration is phase 1, so all 20 users took this path over POST. The log holds exactly the two
deliberate failures above and nothing from the run itself.

### CI doc-path check (#207)

```
scripts/check-doc-paths.sh
check-doc-paths: 381 tracked text file(s) checked, every reference resolves.
```

### Environment

## Security Checklist

- [x] No secrets/tokens in the diff — `config/appsettings.global.json` is untouched and uncommitted

One thing worth a reviewer's eye: the new non-2xx entry logs the response body. It carries the
wallet service's own `BusinessRuleException` message, which is Persian text written for a human
and holds no credential or personal data. In Development an unhandled 500 could put a stack trace
there instead, which is diagnostic rather than sensitive.

## Code Quality

- [x] Comments in English and explain the "why". The existing Persian `Information` line is left
      as it was, per the rule that existing Persian strings are not translated
- [x] No unnecessary dependencies — one `ProjectReference` (`Users.Application`) that the new
      tests need

## Documentation

No `.md` changes. Nothing in `docs/` documents this route or the private method; `AGENT.md`'s
port table and configuration rules are unaffected. Swagger picks the verb up from the mapping.

## Breaking Changes?

- [x] No breaking changes

The verb change is a contract change in principle. In practice the route has exactly one caller,
it is in this repository, and it changes in the same commit. The four APIs bind loopback and are
called only by the bot on the same host, so there is no external consumer. A deployment that ships
one service without the other would see wallet creation fail with 405 — and, thanks to part 2,
would now say so in the log instead of silently registering users without wallets.

## Deployment Notes

No migrations, no new configuration. Deploy `Users.Api` and `Wallet.Api` together — see Breaking
Changes.

**Post-deployment verification**: register a user and confirm three wallet rows appear, or watch
`logs/users-api-*.log` for `Default wallet creation ... was refused: HTTP 405`, which is what a
half-deployed pair looks like.

**Rollback**: revert the commit. Both sides move back together, so there is no intermediate state.

## Related Issues

- Part of #206
- Follows #190 / #204, which removed the other route and made this one the only one
- **#209** — opened from this PR's self-review, and the more serious defect in this path:
  `Users.Api`'s `"WalletAPI"` client sends no `X-API-Key`, so this call gets 401 on any
  Production deployment and default wallets have never been created there. Pre-existing, not
  caused or fixed here — but this PR is what makes it visible in the log, and it should land soon
  after
- **#210** — also from self-review: the endpoint's `catch (BusinessRuleException)` is unreachable
  because the service wraps everything in `InvalidOperationException`, and the response is built
  from entities the repository may have discarded
- Same class of problem as #202 (output written where no console exists)
- #205 (the remaining configuration fallbacks) is untouched, deliberately — separate issue

### DIFF
--- a/src/User/Users.Application/UserService.cs
+++ b/src/User/Users.Application/UserService.cs
@@ -215,16 +215,16 @@ public class UserService
             using var httpClient = _httpClientFactory.CreateClient("WalletAPI");
             _logger.Log(LogLevel.Information,"درخواست برای ساخت کیف پول های پیش فرض" + " " + httpClient.BaseAddress + $"api/wallet/create-default/{userId}");
 
-            var response = await httpClient.PostAsync($"api/wallet/create-default/{userId}", null);
+            using var response = await httpClient.PostAsync($"api/wallet/create-default/{userId}", null);
 
             if (!response.IsSuccessStatusCode)
             {
                 // A non-2xx never reaches the catch below — it is a completed request, not an
                 // exception — and it is the likelier of the two failures, so it needs its own
-                // Error entry. The body carries the wallet service's own reason for refusing.
-                var body = await response.Content.ReadAsStringAsync();
+                // Error entry. The body carries whatever reason the wallet service gave.
+                var body = Truncated(await response.Content.ReadAsStringAsync());
                 _logger.LogError(
-                    "Default wallet creation for user {UserId} was refused: HTTP {StatusCode}. Response: {ResponseBody}. Registration continues; this user has no wallets until one is created on first use.",
+                    "Default wallet creation for user {UserId} was refused: HTTP {StatusCode}. Response: {ResponseBody}. Registration continues; this user's default wallets may be missing or incomplete.",
                     userId, (int)response.StatusCode, body);
             }
         }
@@ -235,8 +235,20 @@ public class UserService
             // must not happen. The exception is logged, so its type is visible in the log
             // rather than flattened away here.
             _logger.LogError(ex,
-                "Default wallet creation for user {UserId} failed. Registration continues; this user has no wallets until one is created on first use.",
+                "Default wallet creation for user {UserId} failed. Registration continues; this user's default wallets may be missing or incomplete.",
                 userId);
         }
     }
+
+    /// <summary>
+    /// Caps a response body before it is logged. An intermediary answering with an HTML error
+    /// page would otherwise be copied whole into the rolling log once per registration, and a
+    /// registration burst is exactly when this logging fires.
+    /// </summary>
+    private static string Truncated(string body) =>
+        body.Length <= MaxLoggedResponseBody
+            ? body
+            : string.Concat(body.AsSpan(0, MaxLoggedResponseBody), "… (truncated)");
+
+    private const int MaxLoggedResponseBody = 500;
 }
--- a/src/User/Users.Application/UserService.cs
+++ b/src/User/Users.Application/UserService.cs
@@ -197,24 +197,46 @@ public class UserService
     /// <summary>
     /// Creates the default wallets for a new user.
     /// </summary>
+    /// <remarks>
+    /// A failure here deliberately does not fail the registration: a user should not be turned
+    /// away because Wallet.Api happens to be restarting. It does have to be recorded, though.
+    /// Until issue #206 it was not — failures went to <c>Console.WriteLine</c>, and under
+    /// <c>sc.exe</c> a Windows service has no console, so a registration burst during a
+    /// Wallet.Api restart committed users with no wallets and said so nowhere. These Error
+    /// entries are the only trace such a user leaves: a missing wallet is created lazily the
+    /// first time it is written to, so afterwards nothing in the wallet database distinguishes
+    /// them from anyone else.
+    /// </remarks>
     /// <param name="userId">User id.</param>
-    /// <returns>Whether it succeeded.</returns>
-    private async Task<bool> CreateDefaultWalletsAsync(Guid userId)
+    private async Task CreateDefaultWalletsAsync(Guid userId)
     {
         try
         {
             using var httpClient = _httpClientFactory.CreateClient("WalletAPI");
             _logger.Log(LogLevel.Information,"درخواست برای ساخت کیف پول های پیش فرض" + " " + httpClient.BaseAddress + $"api/wallet/create-default/{userId}");
-            
-            var response = await httpClient.GetAsync($"api/wallet/create-default/{userId}");
-            
-            return response.IsSuccessStatusCode;
+
+            var response = await httpClient.PostAsync($"api/wallet/create-default/{userId}", null);
+
+            if (!response.IsSuccessStatusCode)
+            {
+                // A non-2xx never reaches the catch below — it is a completed request, not an
+                // exception — and it is the likelier of the two failures, so it needs its own
+                // Error entry. The body carries the wallet service's own reason for refusing.
+                var body = await response.Content.ReadAsStringAsync();
+                _logger.LogError(
+                    "Default wallet creation for user {UserId} was refused: HTTP {StatusCode}. Response: {ResponseBody}. Registration continues; this user has no wallets until one is created on first use.",
+                    userId, (int)response.StatusCode, body);
+            }
         }
         catch (Exception ex)
         {
-            // Log the error but don't fail user creation
-            Console.WriteLine($"خطا در ایجاد کیف پول‌های پیش‌فرض برای کاربر {userId}: {ex.Message}");
-            return false;
+            // Stays broad on purpose: nothing is rethrown, so narrowing it would turn some
+            // failures into failed registrations, which is exactly what the remarks above say
+            // must not happen. The exception is logged, so its type is visible in the log
+            // rather than flattened away here.
+            _logger.LogError(ex,
+                "Default wallet creation for user {UserId} failed. Registration continues; this user has no wallets until one is created on first use.",
+                userId);
         }
     }
 }
--- a/src/Wallet/Wallet.Api/Program.cs
+++ b/src/Wallet/Wallet.Api/Program.cs
@@ -302,8 +302,8 @@ app.MapGet("/api/wallet/transactions/{userId}", async (Guid userId, string? asse
 
 // Creates the wallets a new user starts with: Toman, gold, and the gold credit ledger.
 // POST, not GET: this creates rows, and GET is the verb every intermediary assumes it may
-// retry or prefetch freely. Idempotent — CreateWalletAsync returns the existing wallet — so a
-// repeat is harmless, but the verb should still say what the call does. Issue #206.
+// retry or prefetch freely. A repeat creates no duplicate rows — CreateWalletAsync keeps the
+// existing wallet — but the verb should still say what the call does. Issue #206.
 // userId: User id.
 // walletService: Wallet service.
 // Returns: The wallets that were created.
--- a/src/Wallet/Wallet.Api/Program.cs
+++ b/src/Wallet/Wallet.Api/Program.cs
@@ -301,12 +301,15 @@ app.MapGet("/api/wallet/transactions/{userId}", async (Guid userId, string? asse
 .WithTags("Transactions");
 
 // Creates the wallets a new user starts with: Toman, gold, and the gold credit ledger.
+// POST, not GET: this creates rows, and GET is the verb every intermediary assumes it may
+// retry or prefetch freely. Idempotent — CreateWalletAsync returns the existing wallet — so a
+// repeat is harmless, but the verb should still say what the call does. Issue #206.
 // userId: User id.
 // walletService: Wallet service.
 // Returns: The wallets that were created.
 // 200: Default wallets created.
 // 400: Wallet creation failed.
-app.MapGet("/api/wallet/create-default/{userId}", async (Guid userId, IWalletService walletService) =>
+app.MapPost("/api/wallet/create-default/{userId}", async (Guid userId, IWalletService walletService) =>
 {
     try
     {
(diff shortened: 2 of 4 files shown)


MohKardan:
## Self-review (`/code-review 208`) — all ten findings

Ten findings, every one verified against the tree before answering. Seven were on code this PR
introduces and are fixed in `1ae7b78`. Three are pre-existing defects in the path this PR touches
and are now issues.

### Fixed here

**1. The Error messages claimed more than is known** — "this user has no wallets until one is
created on first use". False in two real cases: a client timeout (`Timeout = 30s`) against a
`Wallet.Api` that commits at t=31s logs it for a user who has all three, and
`WalletService.CreateDefaultWalletsAsync` commits the three wallets in three separate
`SaveChangesAsync` calls with no transaction, so a failure on the second logs it for a user who
has one. Since this PR positions these entries as the way an operator learns which users to look
at, telling them the wrong thing is worse than telling them nothing. Both messages now end
"this user's default wallets may be missing or incomplete", which holds in every case.

**2. The response body was logged untruncated.** A reverse proxy or IIS answering 502 with a
multi-kilobyte HTML page would be copied whole into `logs/users-api-*.log`, once per registration
— and a registration burst against a sick wallet service is the exact scenario this logging
exists for, so the unbounded version is at its worst precisely when it is most needed. Capped at
500 characters with a `… (truncated)` marker, and a test covers it.

**3. `HttpResponseMessage` was not disposed** while the factory-owned `HttpClient` beside it was.
`using var response = …` added. The `using var httpClient` on the factory-created client is
left alone: it is the pre-existing line, and disposing a client from `IHttpClientFactory` is
harmless by design — the factory pools the handler, not the client. Worth tidying by whoever
next has business in this method.

**4. The test double ignored the client name.** `SingleClientFactory.CreateClient` accepted any
name, so the whole suite would have stayed green if `"WalletAPI"` drifted on one side — at which
point the real factory returns a default client with a null `BaseAddress`, `PostAsync` throws
`InvalidOperationException`, and the deliberately-broad catch turns that into one Error line per
registration and no wallets for anybody. The fake now records the names and
`RegisterUserAsync_CreatesDefaultWallets_ThroughTheConfiguredWalletClient` asserts it.

**5. `Assert.Contains("400", …)` could pass on the wrong status code** — the same message
interpolates a `Guid.NewGuid()`, and a 32-hex string contains "400" roughly one time in fifty.
That is flaky-green, which is worse than flaky-red. Now `Assert.Contains("HTTP 400", …)`.

**6. The source assertion on `Wallet.Api`'s mapping was brittle to formatting.** Now a regex
tolerating whitespace and line breaks. It still cannot survive the route being extracted to a
constant or mapped through `MapMethods`; the doc comment says so and says the assertion is meant
to be rewritten rather than deleted if that happens. Asserting over `EndpointDataSource` or a
`WebApplicationFactory` would be better, and is why nothing covers that route table today:
building `Wallet.Api` needs the shared configuration file and a database. A string that describes
the verb is worse than the route table and better than nothing.

**7. The endpoint comment called the call "idempotent" without qualification.** True of the
database, not of the response — see the issue below. Reworded to "a repeat creates no duplicate
rows".

### Not fixed here — pre-existing, now tracked

**8. `Users.Api`'s `"WalletAPI"` client sends no `X-API-Key`, so this call 401s in Production
— #209.** Confirmed: `Wallet.Api/Program.cs:90-104` sets `FallbackPolicy =
RequireAuthenticatedUser()` in Production, `ApiKeyAuthenticationHandler.cs:30` fails without the
header, `install-services.ps1:99` sets `ASPNETCORE_ENVIRONMENT=Production`, and
`Users.Api/Program.cs:117-121` configures only `BaseAddress` and `Timeout` while every other
inter-service client adds the header. Default wallet creation has therefore never worked on a
deployed system, and this PR neither causes nor fixes that.

It does change what happens next: before this PR a 401 was a discarded `false` with no log line,
so the first Production deployment after it merges will show `Default wallet creation for user …
was refused: HTTP 401` once per registration instead of nothing. Being findable is the point of
part 2 — but #209 should land soon after, not eventually. Left out of this PR because verifying
it needs a Production-environment run with a real API key, which is a different piece of work
from the one this branch has verified.

**9. The endpoint's `catch (BusinessRuleException)` is unreachable — #210.**
`WalletService.cs:274-277` wraps everything in `InvalidOperationException`, so a real failure is
a 500 with the fixed body `{"success":false,"message":"خطای داخلی سرور"}`, never the 400 the
comment at `Program.cs:311` documents. Worth naming against this PR specifically: the new
non-2xx entry logs the body on the stated grounds that it carries the wallet service's reason,
and for this endpoint that reason is currently always the same six words. The entry is still
strictly better than the nothing it replaces — the status code alone separates 401 from 500 from
405 — but the body will not earn its place until #210.

**10. `CreateDefaultWalletsAsync` maps its DTOs from the throwaway entities — #210.**
`irrResult`, `mauaResult` and `creditMauaResult` are assigned and never read;
`CreateWalletAsync` returns the existing row when there is one, but the mapper gets the discarded
new entity. Idempotent in the database, not in what it reports.

### Declined

**`FindRepositoryRoot` is a third copy, and `RecordingHandler` a third canned handler.** Fair,
and left as is. The two existing copies each carry a comment explaining that they are duplicated
deliberately rather than shared, on the grounds that extracting the helper means editing tests
the change has no business touching — the same reasoning applies here, and adding a shared
version in `Fakes/` without migrating the other two would leave four call sites and two idioms,
which is worse than three copies. `RecordingHandler` also does something the other two do not:
it captures requests, which is what finding 4's fix needs. Consolidating all of them is a real
cleanup and belongs in a change that is about the test project.

PR #213, opened 2026-09-03
[Refactor] Require service addresses instead of defaulting to literals — issue #205

### LINKED ISSUE #205: Service addresses fall back to hardcoded literals instead of failing at startup, in six places

MohKardan:
## Description

Five clients ended their service-address read with `?? "http://localhost:…"`, and `Orders.Api`
found its Users address by scanning every key in the merged configuration for one ending in
`UsersApiUrl`. Neither branch was ever taken wrongly, because configuration happened to supply
every value — which is exactly why they survived. All of it now goes through
`ConfigurationGuard`, so a missing key stops the host, as `AGENT.md` requires.

## Issue/Task Reference

Closes #205

## Type of Change

- [x] Refactor (code organization, no behavior change *when configuration is complete*)

Behaviour changes only on a host where a key is absent — which is the point.

## The open question in the issue, settled first

> `Orders.Api` registers `IWalletApiClient` twice […] which registration wins (and therefore
> whether the delegate at :118 runs at all) was not established.

**The `AddScoped` at `:129` wins, and the `AddHttpClient` configure delegate never ran.** Proved
by replicating the two registrations and resolving through a scope, not by reasoning:

```
lastDescriptorLifetime=Scoped;  resolved=WalletApiClient;  delegateRan=False
baseAddress=http://from-config.invalid/api      <- the value of configuration["WalletApiUrl"],
                                                   not the marker set inside the delegate
```

Three consequences:

1. **Last registration wins** for `GetRequiredService<IWalletApiClient>()`, and `:129` came after
   `:118`.
2. **It did not fail to resolve** — `WalletApiClient`'s constructor needs an `HttpClient`, and in
   .NET 9 `AddHttpClient()` itself registers a bare transient one (probed separately:
   `bareHttpClient=registered`). That is the *unnamed default* client, so the delegate registered
   under the name `IWalletApiClient` never applied to it.
3. **Line 120 was dead twice over.** Even had the delegate run, `WalletApiClient.cs:54` overwrites
   `BaseAddress` from `configuration["WalletApiUrl"]` immediately afterwards.

So the live wallet address in `Orders.Api` came from `WalletApiClient.cs:51`, not from
`Program.cs:120`. Both are guarded now, and the duplicate registration is removed — asked about
before doing it, per the issue's scope note, and confirmed.

## Section 2, confirmed and narrower than the issue assumed

Replicating `Orders.Api`'s configuration build over the example file:

```
resolved='http://localhost:5136/api'
enumerationOrder=[ Services:TallaEgg.TelegramBot.Infrastructure:UsersApiUrl=http://localhost:5136/api
                 | Services:TallaEgg.Api:UsersApiUrl=http://localhost:5136 ]
```

It landed on the *bot's* key. The other candidate is the bare root, and `UsersApiClient`
concatenates (`{_baseUrl}/users/list`) against routes mapped at `/api/users/list` — so had the
enumeration gone the other way, **every `Orders.Api` to `Users.Api` call would have 404'd**. Not a
cosmetic difference between two shapes: one works and one does not.

Demonstrated against the running stack, both directions — see Testing below.

## Changes Made

**The guards** (`src/TallaEgg/TallaEgg.Core/ConfigurationGuard.cs`)
- Extracted `ReadNonBlank`, the argument guards and blank check both public guards duplicated.
  Deliberately left alone in #204 under scope discipline; there is more than one caller now.
- Added `RequireAbsoluteHttpUri(string?, key)` for a value that reached the caller some other way.
  `RequireUri` is now `ReadNonBlank` + this, so the bot's options-bound path and the services'
  `IConfiguration` path produce the *same words* — pinned by a test.

**The five fallback literals**

| Location | Was | Now |
| --- | --- | --- |
| `Orders.Api/Program.cs:120` | `?? "http://localhost:60933"` | `RequireUri` at startup; the configure delegate is gone, since the client sets its own `BaseAddress` |
| `UsersApiClient.cs:26` | `?? "http://localhost:5001/api"` | `RequireUri` on its own key |
| `WalletApiClient.cs:43` (string ctor) | `?? "http://localhost:60933/api"` | `RequireAbsoluteHttpUri` |
| `WalletApiClient.cs:51` (config ctor) | `?? "http://localhost:60933/api"` | `RequireUri` |
| `OrderApiClient.cs:26` | `?? "http://localhost:5135/api"` | `RequireUri` |

**The scan** — `ResolveUsersApiBaseUrl` is deleted. `Services:Orders.Api:UsersApiUrl` added to the
example file. `Services:TallaEgg.Api:UsersApiUrl` removed from it: with the scan gone it has no
consumer at all — `TallaEgg.Api` never constructs a `UsersApiClient` and maps no endpoints — and
leaving a consumer-less key is what made this confusing in the first place.

**Lazy evaluation (section 3)** — every `ConfigurationGuard` call is out of the `AddDbContext` and
`AddHttpClient` delegates in all five services. `UsersApiClient` in `Orders.Api` is scoped, so its
constructor would not run until the first request either; `Orders.Api` therefore reads that key at
startup as well.

**Logging (section 4)** — Serilog moves to the top of each `Program.cs`, ahead of everything that
can throw, and `StartupLogging.ReportUnhandledExceptionsToLog()` reports a terminating exception
through it. The bot gets the same handler.

## Section 4 turned out to be fully solvable

The issue expected a remainder:

> Not fully solvable in place: the "shared config file not found" case throws before there is any
> configuration to build a logger from.

It is not built from configuration — `WriteTo.Console()` and `WriteTo.File("logs/…")` are fixed in
code — so it can be installed before the file is located. Run from a directory where the walk-up
finds no config:

```
[13:37:58 FTL] The service is terminating on an unhandled exception.
System.IO.FileNotFoundException: Shared configuration 'appsettings.global.json' not found relative to '…\nocfg'.
```

```
…\nocfg\logs\wallet-api-20260903.log      <- written
```

**What does remain**: the sink path is relative to the working directory, and `sc.exe` sets none —
so on an installed service the file lands beside `C:\Windows\System32`, not the service directory.
That is the working-directory half of the same problem, already tracked on #105. This PR makes the
message reach *a* file; #105 is what makes it reach the expected one.

### Build

```
```

### Unit tests

```
dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 772, Skipped: 0, Total: 772
```

741 before; 31 new. Ten existing tests failed on the first run because they built clients from
empty configuration — the guards doing their job. They now supply the real ports.

### The positional part *is* pinned, contrary to the issue's expectation

> Nothing pins the positional part of this. Move a guarded read back inside a configure delegate
> and every test still passes.

`StartupGuardPlacementTests` reads the five `Program.cs` files and asserts (a) every
`ConfigurationGuard.Require*` call sits at column zero — these are top-level statements, so
indentation means it is inside a lambda — (b) each service still reads at least one value through
the guard, and (c) the `Log.Logger` assignment precedes the first guard call — the static logger, not
`UseSerilog()`, is what a guard throwing before the host exists can reach.

Blunt, and it is source inspection rather than behaviour. But it is the difference between a
property that is checked and one that merely happens to hold. Verified by breaking it on purpose —
guard moved back inside `Wallet.Api`'s `AddDbContext` delegate:

```
Failed StartupGuardPlacementTests.EveryConfigurationGuardCall_RunsAtStartupRatherThanInsideAConfigureDelegate
       (relativePath: "src/Wallet/Wallet.Api/Program.cs")
   Assert.Empty() Failure: Collection was not empty
```

`ServiceAddressConfigurationTests` pins the client side, including the one that matters most: a
`UsersApiUrl` defined only in *another service's* section must now be no key at all.

### (a) With complete configuration, nothing changes

All five services and the bot come up. The bot logs all four addresses from its own section and
proceeds to its network diagnostics — no guard fires:

```
[13:48:31 INF] Order API URL: http://localhost:5140/api
[13:48:31 INF] Users API URL: http://localhost:5136/api
[13:48:31 INF] Affiliate API URL: http://localhost:60812/api
[13:48:31 INF] Wallet API URL: http://localhost:60933/api
```

End to end:

```
=== Done in 00:00:33.2. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:
  BTC/IRT: 17 filled
  MAUA/IRT: 17 filled
  SEKE_BAHAR/IRT: 16 filled
```

Those 50 settlements are also the proof that removing the duplicate registration is safe:
settlement resolves `IWalletApiClient` from the outbox processor's scope, which now goes through
the typed-client registration that was previously shadowed.

### (b) The guard fires, at startup

No config had to be edited for this: `Services:Orders.Api:UsersApiUrl` did not exist yet, so
running `Orders.Api` before the key was added gave, before any port was bound:

```
[13:37:21 FTL] The service is terminating on an unhandled exception.
System.InvalidOperationException: Configuration value 'UsersApiUrl' is missing. Add it under
this service's own section, Services:{ApplicationName}, in appsettings.global.json. The service
will not start without it.
   at Program.<Main>$(String[] args) in src/Order/Orders.Api/Program.cs:line 135
```

and the same in `src/Order/Orders.Api/logs/orders-api-20260903.log` — the `[FTL]` line is
Serilog's, which is section 4 working. The key was then added and the service starts normally.

### (c) `Orders.Api` reads Users from its own key, and the two shapes are not interchangeable

`POST /api/orders` reaches `OrderService.cs:117`, `_usersApiClient.GetUserByIdAsync`. Same request,
same user, two configurations:

| `Services:Orders.Api:UsersApiUrl` | Orders.Api log |
| --- | --- |
| `http://localhost:5136/api` (its own key) | `[INF] Validating balance for user …` — no warning; the lookup succeeded |
| `http://localhost:5136` (overridden to the root shape) | `[WRN] Users API returned 404 while fetching user by id …` |

The override reaching the client proves it reads *that* key. The 404 proves the old scan was a coin
flip between working and 404'ing every `Orders.Api` to `Users.Api` call, decided by section-name
enumeration order. Probe orders were placed on simulated users only (negative `TelegramId`) and
cancelled afterwards.

### Doc-link check (#207)

```
scripts/check-doc-paths.sh
check-doc-paths: 382 tracked text file(s) checked, every reference resolves.
```

## Security Checklist

- [x] No secrets in the diff — `config/appsettings.global.json` is untouched and uncommitted; only
      the example file changed, and no value from the live file appears in this PR

## Reported, deliberately not fixed

**`APIKeyConstant.RequireTallaEggApiKey()` has the same lazy shape**, in all four API services
(`Users.Api:105`, `Wallet.Api:102`, `Orders.Api:109`, `Affiliate.Api:91`): it is called inside the
`AddScheme` options delegate, which does not run until authentication first executes — so in
Production a missing `TALLAEGG_API_KEY` surfaces at the first request, not at startup. Left alone
because it is not one of the six the issue names, it reads an environment variable rather than the
shared file, and hoisting it moves a Production-only auth read earlier, which is a behaviour change
in the auth path. **Worth an issue of its own.**

**`UsersApiClient` and `OrderApiClient` discard the `HttpClient` they are given** and construct
their own in the constructor body, so the injected one — and the handler pooling behind it — is
wasted. Pre-existing, unrelated to configuration, untouched here.

## Breaking Changes?

- [x] No breaking changes to any API contract.

Two things a reader should not have to discover for themselves:

- **Configuration is stricter**, which is the intent. See Deployment Notes.
- **`IWalletApiClient` in `Orders.Api` goes from Scoped to Transient.** Removing the shadowing
  `AddScoped` leaves the typed-client registration, which `AddHttpClient<TClient, TImplementation>`
  registers as transient. Within one request, `OrderService`, `OrderCollateralReconciler` and
  `QuoteFillService` previously shared one instance and now each get their own. Harmless today —
  `WalletApiClient` holds no per-request state — but any future per-request field would no longer
  be shared across the three.

## Deployment Notes

**Newly required keys.** A host missing one of these will now fail to start, by design, with a
message naming the key:

| Key | Note |
| --- | --- |
| `Services:Orders.Api:UsersApiUrl` | **New key** — must be added before deploying. Value: `http://localhost:5136/api`, with the `/api` suffix |
| `Services:Orders.Api:WalletApiUrl` | Already in the example file, but until now a host could omit it and the client would default |
| `Services:TallaEgg.TelegramBot.Infrastructure:` `OrderApiUrl` / `UsersApiUrl` / `WalletApiUrl` | Effectively required already — `TelegramBotHostedService.ValidateConfiguration` refused a blank one |
| `Services:Users.Api:WalletApiUrl` | Required since #204; unchanged |

No migrations. `config/appsettings.global.example.json` carries all of them.

**Rollback**: revert the commit. The added configuration key is harmless if left in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ Create your own copy from the template below.
     },
     "Orders.Api": {
       "Urls": [ "http://localhost:5140" ],
+      "UsersApiUrl": "http://localhost:5136/api",
       "WalletApiUrl": "http://localhost:60933/api",
       "Matching": {
         "RequireMarketMakerCounterparty": true,
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/OrderApiClient.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/OrderApiClient.cs
@@ -29,9 +29,11 @@ public class OrderApiClient : IOrderApiClient
         // branch meant talking to something that answers nothing, with nothing in the logs to
         // say so (issue #205).
         //
-        // TrimEnd because the paths below are concatenated, not resolved: Uri normalises an
-        // empty path to "/", which would otherwise produce "host//orders".
-        _baseUrl = ConfigurationGuard.RequireUri(configuration, "OrderApiUrl").ToString().TrimEnd('/');
+        // AbsoluteUri rather than ToString, which unescapes: a percent-escaped or non-ASCII
+        // host would come back decoded and be concatenated into a request against a different
+        // authority than was configured. TrimEnd because the paths below are concatenated, not
+        // resolved: Uri normalises an empty path to "/", which would produce "host//orders".
+        _baseUrl = ConfigurationGuard.RequireUri(configuration, "OrderApiUrl").AbsoluteUri.TrimEnd('/');
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
         var handler = new HttpClientHandler();
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/OrderApiClient.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/OrderApiClient.cs
@@ -23,7 +23,15 @@ public class OrderApiClient : IOrderApiClient
     public OrderApiClient(HttpClient httpClient, IConfiguration configuration, ILogger<OrderApiClient> logger)
     {
         _httpClient = httpClient;
-        _baseUrl = configuration["OrderApiUrl"] ?? "http://localhost:5135/api";
+
+        // Guarded rather than defaulted: the old fallback pointed at 5135, which is TallaEgg.Api
+        // — a host that starts, binds and maps no routes. Orders are on 5140, so taking that
+        // branch meant talking to something that answers nothing, with nothing in the logs to
+        // say so (issue #205).
+        //
+        // TrimEnd because the paths below are concatenated, not resolved: Uri normalises an
+        // empty path to "/", which would otherwise produce "host//orders".
+        _baseUrl = ConfigurationGuard.RequireUri(configuration, "OrderApiUrl").ToString().TrimEnd('/');
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
         var handler = new HttpClientHandler();
--- a/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Program.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Infrastructure/Program.cs
@@ -36,6 +36,10 @@ public class Program
             .WriteTo.File("logs/telegrambot-.log", rollingInterval: RollingInterval.Day, retainedFileCountLimit: 30)
             .CreateLogger();
 
+        // Configuration guards throw before the host exists, and the bot runs as a Windows
+        // service with no console to print to — same as the five APIs (issue #205).
+        TallaEgg.Core.StartupLogging.ReportUnhandledExceptionsToLog();
+
         using var host = CreateHostBuilder(args).Build();
         await host.RunAsync();
     }
--- a/TelegramBot/TallaEgg.TelegramBot.Simulator/Program.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Simulator/Program.cs
@@ -34,8 +34,9 @@ if (!serviceSection.Exists())
 // Some of the bot's own clients (OrderApiClient, UsersApiClient) read flat keys like
 // "OrderApiUrl" straight off IConfiguration rather than through TelegramBotOptions — the real
 // bot's Program.cs flattens Services:{ApplicationName}:* into root-level keys for exactly
-// this reason. Without this step those clients silently fall back to their compiled-in
-// defaults instead of the configured URLs.
+// this reason. Without this step those clients find no key at all and refuse to be built;
+// before issue #205 they were worse than that, falling back to compiled-in defaults so the
+// simulator ran against addresses nobody had configured.
 var prefix = $"Services:{BotApplicationName}:";
 var flattened = serviceSection.AsEnumerable(true)
     .Where(pair => pair.Value is not null)
--- a/config/appsettings.global.example.json
+++ b/config/appsettings.global.example.json
@@ -61,7 +61,6 @@
             "Urls": [
                 "http://localhost:5135"
             ],
-            "UsersApiUrl": "http://localhost:5136",
             "Cors": {
                 "AllowedOrigins": []
             }
@@ -79,6 +78,7 @@
             "Urls": [
                 "http://localhost:5140"
             ],
+            "UsersApiUrl": "http://localhost:5136/api",
             "WalletApiUrl": "http://localhost:60933/api",
             "Cors": {
                 "AllowedOrigins": []
--- a/src/Affiliate/Affiliate.Api/Program.cs
+++ b/src/Affiliate/Affiliate.Api/Program.cs
@@ -15,9 +15,10 @@ using TallaEgg.Core.ErrorHandling;
 var builder = WebApplication.CreateBuilder(args);
 
 // Serilog before anything that can throw. This configuration reads nothing from the shared file
-// — the sinks are fixed here — so it can be installed ahead of the file being located, which is
-// what lets a configuration failure reach the rolling log rather than a console no Windows
-// service has (issue #205).
+// — the sinks are fixed here — so it can be installed ahead of the file being located, and a
+// configuration failure reaches the rolling log rather than only stderr (issue #205). This
+// service is not one of the four installed with sc.exe, so it does have a console; the ordering
+// matches its siblings, where that is the whole point.
 Log.Logger = new LoggerConfiguration()
     .WriteTo.Console()
     .WriteTo.File("logs/affiliate-api-.log", rollingInterval: RollingInterval.Day, retainedFileCountLimit: 30)
--- a/src/Affiliate/Affiliate.Api/Program.cs
+++ b/src/Affiliate/Affiliate.Api/Program.cs
@@ -14,6 +14,18 @@ using TallaEgg.Core.ErrorHandling;
 
 var builder = WebApplication.CreateBuilder(args);
 
+// Serilog before anything that can throw. This configuration reads nothing from the shared file
+// — the sinks are fixed here — so it can be installed ahead of the file being located, which is
+// what lets a configuration failure reach the rolling log rather than a console no Windows
+// service has (issue #205).
+Log.Logger = new LoggerConfiguration()
+    .WriteTo.Console()
+    .WriteTo.File("logs/affiliate-api-.log", rollingInterval: RollingInterval.Day, retainedFileCountLimit: 30)
+    .CreateLogger();
+
+builder.Host.UseSerilog();
+StartupLogging.ReportUnhandledExceptionsToLog();
+
 const string sharedConfigFileName = "appsettings.global.json";
 var sharedConfigPath = ResolveSharedConfigPath(builder.Environment, sharedConfigFileName);
 builder.Configuration.AddJsonFile(sharedConfigPath, optional: false, reloadOnChange: true);
@@ -61,8 +73,13 @@ if (string.IsNullOrWhiteSpace(builder.Configuration[WebHostDefaults.ServerUrlsKe
 }
 
 // تنظیم اتصال به دیتابیس SQL Server
+// SQL Server connection. Read here, not inside the options delegate below: that delegate does
+// not run until DbContextOptions<T> is first resolved, so a missing connection string failed
+// startup only because the migration block further down happens to resolve the context (#205).
+var affiliateConnectionString = ConfigurationGuard.RequireConnectionString(builder.Configuration, "AffiliateDb");
+
 builder.Services.AddDbContext<AffiliateDbContext>(options =>
-    options.UseSqlServer(ConfigurationGuard.RequireConnectionString(builder.Configuration, "AffiliateDb"),
+    options.UseSqlServer(affiliateConnectionString,
         b => b.MigrationsAssembly("Affiliate.Api")));
 
 // فقط در production محافظت فعال شود
@@ -98,14 +115,6 @@ builder.Services.AddTallaEggCors(builder.Configuration);
 builder.Services.AddScoped<IAffiliateRepository, AffiliateRepository>();
 builder.Services.AddScoped<AffiliateService>();
 builder.Services.AddTallaEggErrorHandling();
-
-// پیکربندی Serilog برای لاگ‌نویسی روی فایل و کنسول
-Log.Logger = new LoggerConfiguration()
-    .WriteTo.Console()
-    .WriteTo.File("logs/affiliate-api-.log", rollingInterval: RollingInterval.Day, retainedFileCountLimit: 30)
-    .CreateLogger();
-
-builder.Host.UseSerilog();
 
 var app = builder.Build();
 
--- a/src/Order/Orders.Api/Program.cs
+++ b/src/Order/Orders.Api/Program.cs
@@ -136,13 +136,13 @@ ConfigurationGuard.RequireUri(builder.Configuration, "UsersApiUrl");
 builder.Services.AddScoped<UsersApiClient>();
 
 
-// Add Wallet API Client. The address is read and parsed out here rather than inside the
-// configure delegate, which does not run until the first client is created (issue #205).
-var walletApiBaseAddress = ConfigurationGuard.RequireUri(builder.Configuration, "WalletApiUrl");
-builder.Services.AddHttpClient<TallaEgg.Infrastructure.Clients.IWalletApiClient, TallaEgg.Infrastructure.Clients.WalletApiClient>(client =>
-{
-    client.BaseAddress = walletApiBaseAddress;
-});
+// Add Wallet API Client. No configure delegate: WalletApiClient's constructor reads
+// WalletApiUrl and assigns BaseAddress itself, so anything set here would be overwritten a
+// moment later — which is what made the old fallback on this line unreachable twice over.
+// The read below is what the delegate could not be: the typed client is transient, so its
+// constructor runs on first resolution, and only reading here fails a missing key at startup.
+ConfigurationGuard.RequireUri(builder.Configuration, "WalletApiUrl");
+builder.Services.AddHttpClient<TallaEgg.Infrastructure.Clients.IWalletApiClient, TallaEgg.Infrastructure.Clients.WalletApiClient>();
 
 // CORS — issue #31: a whitelist read from configuration, not AllowAnyOrigin.
 builder.Services.AddTallaEggCors(builder.Configuration);
--- a/src/Order/Orders.Api/Program.cs
+++ b/src/Order/Orders.Api/Program.cs
@@ -30,6 +30,19 @@ var builder = WebApplication.CreateBuilder(args);
 // no third-party supervisor needed (issue #70).
 builder.Host.UseWindowsService();
 
+// Serilog before anything that can throw. This configuration reads nothing from the shared file
+// — the sinks are fixed here — so it can be installed ahead of the file being located, which is
+// what lets a configuration failure reach the rolling log rather than a console no Windows
+// service has (issue #205).
+Log.Logger = new LoggerConfiguration()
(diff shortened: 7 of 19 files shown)


MohKardan:
## Self-review response (`/code-review 213`)

Thirteen findings. Eight fixed in `ee5f7ea`, five answered below.

### Fixed

**1. `README.md:119` — the copy-paste template lacked the now-mandatory `Services:Orders.Api:UsersApiUrl`.**
The most important finding here, and one I missed: `docs/operations/WINDOWS_DEPLOYMENT.md:32` points
operators at that template, so a host built from it would have refused to start with the very guard
this PR adds. Added.

**3. `Orders.Api/Program.cs:144` — `client.BaseAddress = walletApiBaseAddress` was dead.**
Correct, and the awkward part is that the PR body proves it dead two paragraphs above the line it
keeps: `WalletApiClient`'s constructor reads `WalletApiUrl` and assigns `BaseAddress` itself. The
configure delegate is gone; `AddHttpClient<IWalletApiClient, WalletApiClient>()` now takes no
delegate, and the startup read stays as the thing the delegate could never be.

**4. `StartupGuardPlacementTests.cs:93` — anchored on `UseSerilog()` rather than the `Log.Logger` assignment.**
The sharpest finding of the set. `StartupLogging.ReportUnhandledExceptionsToLog()` writes through
Serilog's *static* `Log`; `UseSerilog()` only redirects `ILogger<T>` resolved from the host, which
does not exist when a guard throws. The test now anchors on `Log.Logger = new LoggerConfiguration()`
and is renamed `TheStaticSerilogLogger_IsAssignedBeforeTheFirstGuardCanThrow`.

Verified against exactly the scenario described — `Log.Logger` moved below the guard in
`Users.Api`, `UseSerilog()` left at line 36 above it:

```
Failed StartupGuardPlacementTests.TheStaticSerilogLogger_IsAssignedBeforeTheFirstGuardCanThrow
       (relativePath: "src/User/Users.Api/Program.cs")
   Assert.InRange() Failure: Value not in range
```

The old anchor passed that same tree. Good catch.

**8. `StartupGuardPlacementTests.cs:52` — the column-zero rule would reject a guard legitimately nested in `if (IsProduction())`.**
True; it cannot tell a lambda body from an ordinary nested block. I have kept the rule rather than
attempting to parse C# with a regex, and said so in the test's own documentation: when a guard
genuinely belongs inside a conditional startup block, either hoist the read above the block or relax
the rule deliberately — do not quietly indent around it. A lint that fails loudly with a stated
escape hatch seemed better than one that tries to be clever.

**9. `StartupLogging.cs:20` — asserts all callers run under `UseWindowsService()`; two do not.**
Confirmed against `scripts/windows-services/install-services.ps1:59-62`: only Wallet, Users, Orders
and the bot are installed. `Affiliate.Api` and `TallaEgg.Api` call neither `UseWindowsService()` nor
get installed, and I had copied the "no console a Windows service has" comment into both. Reworded in
those two, and `StartupLogging`'s own comment now says "the four deployed hosts" rather than "the
services".

**10. `Simulator/Program.cs:37` — comment describes behaviour this PR deleted.**
Right, and it is the exact failure mode this issue is about, preserved in prose: a maintainer would
go looking for a silent wrong-URL problem when what they have is a loud missing-key exception.
Rewritten.

**11. `Uri.ToString()` vs `AbsoluteUri`.**
Correct — `ToString()` unescapes. Both concatenated base URLs (`UsersApiClient`, `OrderApiClient`)
now use `AbsoluteUri`, with the reason in a comment. The `TrimEnd('/')` stays for the separate reason
already documented.

**12. UTF-8 BOM added to four files.**
Real, and self-inflicted: my editing helper wrote `utf-8-sig` unconditionally, so
`ConfigurationGuard.cs`, `Wallet.Api/Program.cs` and two test files went from `75 73 69` to
`EF BB BF`. Reverted; those four are back to no BOM, and the two new test files match their
siblings.

### Not fixed, with reasons

**2 and 7. `UsersApiClient` and `OrderApiClient` discard the injected `HttpClient`.**
Confirmed and genuinely bad — `UsersApiClient` is `AddScoped` in `Orders.Api`, so it is one new
`HttpClient` + `HttpClientHandler` per request, none disposed, bypassing the factory's pooling
entirely. The finding is right about the consequence.

Not fixed here under scope discipline: it is pre-existing, has nothing to do with configuration, and
`STANDARDS.md` is explicit that a change should not carry a repair of adjacent code nobody asked
for. It was already in this PR's "Reported, deliberately not fixed" section before the review; the
review has made the case sharper, so **this should get an issue of its own — shall I open one?** The
socket-per-request behaviour in `Orders.Api` is the part that deserves a number.

**5. `Orders.Api:135` discards the parsed `UsersApiUrl` and the client reads it again.**
Accurate description; I am keeping it, and after fixing finding 3 both of `Orders.Api`'s guards are
now symmetric statement-calls, so at least the pattern is consistent and commented.

The alternative — hoist the `Uri` and pass it to the client — means changing `UsersApiClient`'s
constructor signature and every call site in the bot, the simulator and the tests. That is a wider
change than #205 asks for, and it would not remove the duplication so much as move it: the bot
reaches the same client through `TelegramBotOptions`, so the key would still be read in two shapes.

On the `reloadOnChange` scenario: editing the shared file at runtime to a blank value would indeed
give a 500 rather than a refused start. That is true of every configuration value this system reads
lazily and is not made worse by this PR — a fair note for a future issue about validating on reload,
not something to solve inside this one.

**6. Removing `AddScoped` changes the lifetime from Scoped to Transient.**
Correct, and the PR body understated it — "no breaking changes" was about the API contract, but the
lifetime change deserved naming rather than being left for a reader to discover. Now stated
explicitly in the PR description.

Keeping the transient typed client rather than restoring a scoped registration: that is the standard
shape for `AddHttpClient<TClient, TImplementation>`, `WalletApiClient` holds no per-request state,
and the alternative is to keep the registration that was shadowing the typed client in the first
place. The 50 settlements in the smoke run exercise the new resolution path end to end. The finding
is right that a future per-request field would no longer be shared — worth knowing, and now on the
record.

**13. `RequireConnectionString`'s `ArgumentException` ParamName changed from `name` to `key`.**
Technically true. Reachable only by calling `RequireConnectionString(config, "")` — passing a blank
*name*, a programmer error — and every call site in the solution passes a literal. The operator-facing
message is unaffected: a missing connection string still reports `Connection string 'OrdersDb' is
missing…`, naming the string, not a parameter. Adding a redundant argument guard to `RequireConnectionString`
to preserve a ParamName on an unreachable path would undo part of the extraction the issue asked for,
so I have left it.

```
scripts/check-doc-paths.sh
check-doc-paths: 385 tracked text file(s) checked, every reference resolves.

dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 772, Skipped: 0, Total: 772
```

Smoke re-run, because finding 3 changed how `Orders.Api` registers its wallet client:

```
=== Done in 00:00:36.0. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:
  BTC/IRT: 17 filled
  MAUA/IRT: 17 filled
  SEKE_BAHAR/IRT: 16 filled
```

MohKardan:
Both items from the "Reported, deliberately not fixed" section above are now tracked together in #214 — they are two separate defects, but they sit in the same wiring and were found by the same walk through it, so they share one issue with independent sections rather than two.

One correction to what this PR said about the second: I claimed the API-key guard "surfaces at the first request, not at startup" without having run it. Running it turned up something worth recording — `README.md:190` promises "each one throws at startup if it is missing", and it does not. The service binds its port and then returns 500 to every request. Details and the transcript are in #214.

Issue #214, opened 2026-09-03
DI wiring: two clients discard their injected HttpClient, and the API-key guard fires per request instead of at startup


MohKardan:
Found while doing #205 and reviewing its PR (#213). **Two distinct defects**, collected into one
issue because they sit in the same wiring, were found by the same walk through it, and share a
family: in both, what the DI container does at run time differs from what the code reads like.
They do **not** share a fix — each section below stands alone and can be picked up separately.

Neither is a security hole, and neither is urgent.

---

## 1. Two API clients throw away the `HttpClient` they are given

`UsersApiClient` and `OrderApiClient` both assign the injected client and then overwrite it with
one they build themselves:

```csharp
// src/TallaEgg/TallaEgg.Infrastructure/Clients/UsersApiClient.cs:25 and :46
_httpClient = httpClient;          // the factory's client, with its pooled handler
// ... 20 lines later ...
_httpClient = new HttpClient(handler);   // and here it is discarded
```

`TelegramBot/TallaEgg.TelegramBot.Infrastructure/Clients/OrderApiClient.cs:25` and `:44` are the
same two lines. Nothing disposes either object, and the `HttpClientHandler` built alongside is a
fresh connection pool each time.

**Why it matters more in `Orders.Api` than anywhere else.** The bot and the simulator register both
clients as singletons (`Program.cs:114`, `:119`; `Simulator/Program.cs:71`, `:75`), so there the
waste is one abandoned client per process — untidy, bounded. But `Orders.Api` registers
`UsersApiClient` **scoped** (`src/Order/Orders.Api/Program.cs:136`), so every inbound request
constructs one, which means **a new `HttpClient` and a new socket pool per request**, none of them
disposed. Sockets accumulate in `TIME_WAIT`; under sustained load that ends in ephemeral-port
exhaustion, and `IHttpClientFactory`'s handler rotation — the thing that picks up DNS changes — is
bypassed entirely.

There is no evidence of this biting yet. Traffic today is one bot on the same host, and the
simulator's runs are short.

**Untouched deliberately.** #205 was about configuration, and `STANDARDS.md` is explicit that a
change should not carry a repair of adjacent code nobody asked for — so #213 rewrote the
surrounding lines of both constructors and left these two alone, reporting them instead.

**Likely fix**: delete the `_httpClient = new HttpClient(handler)` line and keep the injected
client. The `#if DEBUG` certificate-validation callback on the discarded handler is the thing to
[...]

Issue #211, opened 2026-09-03
Serilog writes every service's log to C:\Windows\System32\logs\ when running as a Windows service


MohKardan:
## What happens

Observed on the production VM on 2026-09-03, during the first deployment that has ever installed
these as real Windows services. After `install-services.ps1` created the four services, every log
file appeared under `C:\Windows\System32\logs\`:

```
orders-api-20260903.log
telegrambot-20260903.log
users-api-20260903.log
wallet-api-20260903.log
```

Neither `C:\TallaEgg\logs\` nor `C:\TallaEgg\publish\<Service>\logs\` exists on that machine.

## Why

The Serilog file sink path is relative in all six programs (line numbers on `main`):

| File | Line | Path |
| --- | --- | --- |
| `src/Affiliate/Affiliate.Api/Program.cs` | 105 | `"logs/affiliate-api-.log"` |
| `src/Order/Orders.Api/Program.cs` | 220 | `"logs/orders-api-.log"` |
| `src/TallaEgg/TallaEgg.Api/Program.cs` | 89 | `"logs/tallaegg-api-.log"` |
| `src/User/Users.Api/Program.cs` | 144 | `"logs/users-api-.log"` |
| `src/Wallet/Wallet.Api/Program.cs` | 77 | `"logs/wallet-api-.log"` |
| `TelegramBot/TallaEgg.TelegramBot.Infrastructure/Program.cs` | 36 | `"logs/telegrambot-.log"` |

Serilog resolves a relative sink path against the process working directory. `sc.exe create` has
no option to set one, so the SCM starts each service in `C:\Windows\System32`.
`UseWindowsService()` sets `ContentRootPath` to `AppContext.BaseDirectory` but does **not** change
`Environment.CurrentDirectory`, so it does not help the sink.

## Why it matters

- `docs/operations/WINDOWS_DEPLOYMENT.md:76` states that each service writes its own
  `logs\<service>-.log` and tells the operator to `Get-Content -Wait` it. That instruction does not
  work on a deployed machine — the files are not there.
- KR1 is measured on a server uptime log. The log exists, but not where the runbook says to look.
- Writing into a Windows system directory is only possible because the services run as
  `LocalSystem`, and it mixes product output into `System32`.
- `README.md:291` describes the `logs/` directory next to each project, which is accurate for
  `dotnet run` and misleading for a deployment.

This is misplacement, not unbounded growth: `retainedFileCountLimit: 30` still applies.

[...]

PR #216, opened 2026-09-03
[Fix] Users.Api sends X-API-Key when creating default wallets — issue #209

### LINKED ISSUE #209: Users.Api's WalletAPI client sends no X-API-Key, so default wallet creation 401s in Production
Found while reviewing #208 (issue #206). Pre-existing, unrelated to the verb change in that PR,
and more consequential than either defect #206 describes.

## What is wrong

`Users.Api`'s `"WalletAPI"` named client does not send the `X-API-Key` header:

```csharp
// src/User/Users.Api/Program.cs:116-121
var walletApiBaseAddress = ConfigurationGuard.RequireUri(builder.Configuration, "WalletApiUrl");
builder.Services.AddHttpClient("WalletAPI", client =>
{
    client.BaseAddress = walletApiBaseAddress;
    client.Timeout = TimeSpan.FromSeconds(30);
});
```

`Wallet.Api` requires it in Production:

- `src/Wallet/Wallet.Api/Program.cs:90-104` — in `IsProduction()`, registers the `ApiKey` scheme
  and sets `FallbackPolicy = RequireAuthenticatedUser()`, so every endpoint is protected by
  default
- `src/Wallet/Wallet.Api/Program.cs:145-151` — `UseAuthentication()` in Production, then
  `UseAuthorization()`
- `src/TallaEgg/TallaEgg.Core/ApiKeyAuthenticationHandler.cs:30` —
  `AuthenticateResult.Fail("Missing API Key")` when the header is absent

And Production is what a real deployment runs:
`scripts/windows-services/install-services.ps1:99` sets `ASPNETCORE_ENVIRONMENT=Production`.

So `GET`/`POST api/wallet/create-default/{userId}` — the only route into default-wallet creation
since #190 — gets **401 for every registration on a deployed system**. Nobody gets Toman, MAUA or
CREDIT_MAUA at registration there.

Every other inter-service client sends the key:

| Client | Adds the header at |
| --- | --- |
| `TallaEgg.Infrastructure/Clients/WalletApiClient.cs` | `:45`, `:56` |
| `TallaEgg.Infrastructure/Clients/UsersApiClient.cs` | `:35` |
| `TelegramBot/.../Clients/OrderApiClient.cs` | `:35` |
| `TelegramBot/.../Clients/AffiliateApiClient.cs` | `:28` |

`Users.Api`'s raw named client is the only one that does not, and it is the only one registered
through `AddHttpClient(string, …)` rather than as a typed client whose constructor sets the
header. `Orders.Api:118` uses `AddHttpClient<IWalletApiClient, WalletApiClient>`, so it inherits
the header from the client class and is unaffected.

## Why nobody noticed

Three things hid it, and #208 removes the third:

1. **Development skips the whole auth block.** `dotnet run` and `driver.ps1` both run
   Development, where `Wallet.Api` registers no authentication at all — every local run, every
   simulator run and every manual test passes.
2. **The 401 is a non-2xx, not an exception.** Before #208, `UserService.CreateDefaultWalletsAsync`
   returned `response.IsSuccessStatusCode` and both callers discarded it, so a 401 produced no
   log line anywhere.
3. **A missing wallet is created lazily on first write.** Since the lazy-creation work behind
   `WalletLazyCreationTests`, a deposit, a balance lock or a settlement creates the row on
   demand, so the symptom is not "trading is broken" — it is that a user's balance view fails
   (`WalletService.GetBalanceAsync` throws "کیف پول پیدا نشد") until something writes to them.

After #208 merges this stops being silent: the failure logs at `Error` with the status, so a
Production deployment will show `Default wallet creation for user … was refused: HTTP 401` once
per registration. That is how it should be found, but it is not a fix.

## Fix

One line, matching what every other client does:

```csharp
builder.Services.AddHttpClient("WalletAPI", client =>
{
    client.BaseAddress = walletApiBaseAddress;
    client.Timeout = TimeSpan.FromSeconds(30);
    client.DefaultRequestHeaders.Add("X-API-Key", APIKeyConstant.TallaEggApiKey);
});
```

Worth checking whether `APIKeyConstant.RequireTallaEggApiKey()` should be used here so a missing
`TALLAEGG_API_KEY` fails at startup, the way #204 made `WalletApiUrl` fail.

## Verification this needs

Development proves nothing here — that is the whole reason the defect survived. It has to be
exercised with `ASPNETCORE_ENVIRONMENT=Production` and a `TALLAEGG_API_KEY` set on both services:
register a user, and confirm three wallet rows rather than a 401 in `logs/users-api-*.log`.

Not verified in this state yet: everything above is read from the code and from
`install-services.ps1`, and the local runs in #208 were all Development.

MohKardan:
## Description

`Users.Api` creates a new user's default wallets by calling `Wallet.Api` through the `"WalletAPI"`
named client. That client sent no `X-API-Key`, and `Wallet.Api` requires one in Production, so
every registration on a deployed system was refused with 401 and the user got no wallet rows.
This adds the header every other inter-service client already sends.

## Issue/Task Reference

Closes #209

## Type of Change

- [x] Bug fix, not urgent

Not urgent for the reason the evidence below makes precise: the deployed product looks healthy,
because lazy creation produces the missing rows as soon as anyone touches the user.

---

## Step 0 — what actually happened on the server

The issue was written from a static read of the code and said so. This PR confirms it three ways
before changing anything.

**(a) The client sends no key** — `src/User/Users.Api/Program.cs:133-137` on `main` set
`BaseAddress` and `Timeout` and nothing else.

**(b) `Wallet.Api` requires it in Production only** — `src/Wallet/Wallet.Api/Program.cs:97-115`
registers the `ApiKey` scheme and `FallbackPolicy = RequireAuthenticatedUser()` inside
`IsProduction()`; the `else` branch adds authorization *with no authentication in front of it*.
`src/Wallet/Wallet.Api/Program.cs:152-156` calls `UseAuthentication()` in Production only. That is
why no local run, simulator run or manual test has ever seen this.

**(c) The server log** — recorded by the product owner on the deployment in the #209 comment:
the `[ERR] ... was refused: HTTP 401` line from #208 fires on every registration, `Wallet.Api`
answers `401` with "Missing API Key", and the wallet rows appear later, when an operator grants
credit. So the 401 did happen and lazy creation masked it.

## Verification: Production, before and after

Development proves nothing here, so both runs used `ASPNETCORE_ENVIRONMENT=Production` and a
`TALLAEGG_API_KEY` on both services, registering one user through `POST /api/user/register` and
reading the wallet table **before any deposit**. Both excerpts below are from the code as merged —
the "after" run was repeated on the second commit.

One thing worth recording for whoever reproduces this next: the first attempt showed no 401 at
all and looked like the issue was wrong. `dotnet run` applies `Properties/launchSettings.json`,
which pins `ASPNETCORE_ENVIRONMENT=Development` over the ambient variable, so that run was
Development wearing a Production label. `--no-launch-profile` is what makes the condition real.

| | Before the fix | After the fix |
| --- | --- | --- |
| HTTP response to `POST api/wallet/create-default/{id}` | `401` | `200` |
| `users-api-*.log` | `[ERR] Default wallet creation for user "..." was refused: HTTP 401.` | no error line |
| Wallet rows, before any deposit | `ROW COUNT = 0` | `IRT`, `MAUA`, `CREDIT_MAUA` — `ROW COUNT = 3` |

Before:

```
[INF] Sending HTTP request POST http://localhost:60933/api/wallet/create-default/2454ad26-...
[INF] Received HTTP response headers after 154.5894ms - 401
[ERR] Default wallet creation for user "2454ad26-..." was refused: HTTP 401. Response: .
      Registration continues; this user's default wallets may be missing or incomplete.

=== wallet rows, BEFORE any deposit ===
ROW COUNT = 0
```

After:

```
[INF] Sending HTTP request POST http://localhost:60933/api/wallet/create-default/2a219922-...
[INF] Received HTTP response headers after 706.6875ms - 200

=== wallet rows, BEFORE any deposit ===
CREDIT_MAUA | balance=0.00000000
IRT         | balance=0.00000000
MAUA        | balance=0.00000000
ROW COUNT = 3
```

"Before any deposit" is the whole point: wallets are created lazily, so a successful trade — or
the mere existence of wallet rows afterwards — would not have distinguished a fixed registration
path from a broken one masked by the first write.

## Changes Made

- `src/User/Users.Api/Program.cs` — the `"WalletAPI"` named client now sets `X-API-Key`, resolves
  the key at startup (Production requires the real one; elsewhere the placeholder), and declares
  `RedactLoggedHeaders` so the key cannot reach a log
- `tests/TallaEgg.AllServices.Tests/DefaultWalletCreationFailureTests.cs` — one regression test
  pinning the header inside that registration

The second commit applies this PR's own self-review; the reasoning for each of the eight findings
is in the review comment below. Two things it corrected are worth having in the description
rather than only in a comment, because both were wrong here first:

### The key is resolved per environment, not unconditionally

The issue asked whether `RequireTallaEggApiKey()` belonged here. This description originally
argued no, on the grounds that requiring it would stop every clone and CI job. That objection is
real but it only applies outside Production:

```csharp
var walletApiKey = builder.Environment.IsProduction()
    ? APIKeyConstant.RequireTallaEggApiKey()
    : APIKeyConstant.TallaEggApiKey;
```

Without this, `APIKeyConstant.TallaEggApiKey` substitutes the public dev placeholder whenever
`TALLAEGG_API_KEY` is unset — so a Production host that never received the variable would send a
placeholder to a `Wallet.Api` enforcing the real key and reproduce #209 exactly, with the header
present so the code reads as fixed. Users.Api's own `RequireTallaEggApiKey()` at `:104` does not
prevent that: it sits in an `AddScheme` options delegate, which runs on first authenticated
request rather than at startup. Verified — Production with no key now refuses to start:

```
[FTL] The service is terminating on an unhandled exception.
System.InvalidOperationException: Environment variable 'TALLAEGG_API_KEY' is not set.
```

Development with no key set is unaffected; the smoke run recorded below had none in its
environment.

The read is hoisted next to `walletApiBaseAddress` rather than left inside the configure delegate,
for the reason the comment above it already gives about the address (#205): the delegate does not
run until the first client is created, so a guard placed there fails one registration at a time
inside a broad catch instead of stopping the service.

### There is a regression test, and the reasoning that said otherwise was wrong

This description originally claimed no test was practical without booting the host. That was
wrong. `WalletApi_MapsDefaultWalletCreation_AsPostNotGet`, in the same test file, already reads
`Program.cs` as source text and asserts against it, and `StartupGuardPlacementTests` does the same
across all five services for exactly this reason. The new test follows that pattern, matches the
header *inside* the `"WalletAPI"` registration, and requires an identifier rather than a string
literal so an inline key fails it too (#33). Confirmed it fails when the header line is removed —
without it, deleting that one line leaves all 790 tests green, which is how the line came to be
missing for the life of the deployment.

### Unit Tests

```
Build succeeded. 0 Warning(s) 0 Error(s)

dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 790, Skipped: 0, Total: 790
```

### End-to-end

Production, by hand — the before/after table above.

Development, unchanged, via `driver.ps1` — run with no `TALLAEGG_API_KEY` in the environment,
which is the clone/CI condition:

```
=== Done in 00:00:23.4. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:  BTC/IRT: 17 filled  MAUA/IRT: 17 filled  SEKE_BAHAR/IRT: 16 filled
```

Auto-quote flags restored for both symbols the driver touched. Date/time run: 2026-09-03.

- [x] `scripts/check-doc-paths.sh` — 386 tracked text files checked, every reference resolves

## Security Checklist

- [x] No hardcoded secrets, passwords, API keys, or tokens — the key is read from
      `TALLAEGG_API_KEY` through `APIKeyConstant`, and no key value appears in this diff, the
      commits, this description or any log quoted here
- [x] The key cannot reach a log — `RedactLoggedHeaders(["X-API-Key"])`. This is the only
      key-bearing client in the solution built through `IHttpClientFactory`, whose logging
      handlers write request headers at Trace and redact nothing by default
- [x] `config/appsettings.global.json` untouched and untracked
- [x] Authentication/authorization checks in place — this restores one that was being skipped

## Breaking Changes?

- [x] No breaking changes

A request that was being rejected in Production starts being accepted. Development is unaffected:
`Wallet.Api` registers no authentication there, and an unread header changes nothing.

## Deployment Notes

No migration, no new configuration. **`TALLAEGG_API_KEY` must be present in `Users.Api`'s
environment**, and after this change Production `Users.Api` will not start without it — it fails
with a message naming the variable, in `logs/users-api-*.log`. `install-services.ps1:99` already
sets it for every installed service, so a deployment installed that way needs nothing new; one
where the variable was set by hand should be checked before restarting.

**Post-deployment verification**: register one user and confirm no
`Default wallet creation ... was refused` line in `logs/users-api-*.log`. Since #215 those logs
sit beside the binary in the publish folder, not `System32`. Do not read a successful trade as
proof — that was exactly the trap this issue was hiding behind.

**Rollback**: revert this commit and restart `Users.Api`. Behaviour returns to registrations
logging a 401 with wallets created lazily on first write; no data written under this change needs
undoing.

## Related Issues

- Follows #206 / #208 — that PR made this failure path log at `Error`, which is what let the
  owner confirm the 401 on the server rather than guessing
- Follows #190, which left `create-default` as the only route into wallet creation
- Related to #214 (clients that discard the injected `HttpClient`) — same file, deliberately
  untouched here
- #215 moved these logs out of `System32`, which is why the log evidence above was readable

### DIFF
--- a/src/User/Users.Api/Program.cs
+++ b/src/User/Users.Api/Program.cs
@@ -134,6 +134,14 @@ builder.Services.AddHttpClient("WalletAPI", client =>
 {
     client.BaseAddress = walletApiBaseAddress;
     client.Timeout = TimeSpan.FromSeconds(30);
+
+    // Wallet.Api requires X-API-Key in Production, so without this every registration on a
+    // deployed system got 401 and created no wallets. It stayed hidden because Development
+    // registers no authentication at all, and because a missing wallet is created lazily on
+    // first write — the first deposit or trade produced the rows registration had failed to
+    // (issue #209). The typed clients set the same header in their constructors; a named
+    // client has no constructor to set it in, so it goes here.
+    client.DefaultRequestHeaders.Add("X-API-Key", APIKeyConstant.TallaEggApiKey);
 });
 
 // CORS — issue #31: a whitelist read from configuration, not AllowAnyOrigin.
--- a/src/User/Users.Api/Program.cs
+++ b/src/User/Users.Api/Program.cs
@@ -130,6 +130,19 @@ builder.Services.AddTallaEggErrorHandling();
 // this client and swallows what it throws (UserService.CreateDefaultWalletsAsync), so deferring
 // the failure means users registered with no wallets and nothing said about it.
 var walletApiBaseAddress = ConfigurationGuard.RequireUri(builder.Configuration, "WalletApiUrl");
+
+// Read here rather than inside the configure delegate below, for the reason the address is:
+// the delegate does not run until the first client is created, so a missing key would surface
+// one registration at a time inside the broad catch in UserService.CreateDefaultWalletsAsync
+// instead of stopping the service (issue #205). Production demands the real key — sending the
+// placeholder to a Wallet.Api that enforces the real one reproduces issue #209 exactly, with
+// the header present so the code reads as fixed. Outside Production the placeholder is right:
+// no service registers API-key authentication there, and requiring the variable would stop
+// every clone and CI job that has no reason to hold it.
+var walletApiKey = builder.Environment.IsProduction()
+    ? APIKeyConstant.RequireTallaEggApiKey()
+    : APIKeyConstant.TallaEggApiKey;
+
 builder.Services.AddHttpClient("WalletAPI", client =>
 {
     client.BaseAddress = walletApiBaseAddress;
@@ -141,8 +154,13 @@ builder.Services.AddHttpClient("WalletAPI", client =>
     // first write — the first deposit or trade produced the rows registration had failed to
     // (issue #209). The typed clients set the same header in their constructors; a named
     // client has no constructor to set it in, so it goes here.
-    client.DefaultRequestHeaders.Add("X-API-Key", APIKeyConstant.TallaEggApiKey);
-});
+    client.DefaultRequestHeaders.Add("X-API-Key", walletApiKey);
+})
+// The only key-bearing client in the solution built through IHttpClientFactory, whose logging
+// handlers write request headers at Trace and redact nothing by default. The other four are
+// raw HttpClients and never reach those handlers, so this is the one place the shared key
+// could be written to a log by lowering a level.
+.RedactLoggedHeaders(["X-API-Key"]);
 
 // CORS — issue #31: a whitelist read from configuration, not AllowAnyOrigin.
 builder.Services.AddTallaEggCors(builder.Configuration);
--- a/tests/TallaEgg.AllServices.Tests/DefaultWalletCreationFailureTests.cs
+++ b/tests/TallaEgg.AllServices.Tests/DefaultWalletCreationFailureTests.cs
@@ -277,6 +277,33 @@ public class DefaultWalletCreationFailureTests
     }
 
     /// <summary>
+    /// The credential, from the calling side. Wallet.Api requires <c>X-API-Key</c> in Production
+    /// and registers no authentication at all in Development, so a sender that omits it passes
+    /// every local run, every simulator run and every test in this project, and 401s on every
+    /// registration once deployed — where lazy wallet creation then hides the result behind the
+    /// first deposit or trade (issue #209). It was missing for the life of the deployment.
+    ///
+    /// <para>
+    /// Asserted against the source for the same reason as the test above: the registration lives
+    /// in Users.Api's top-level statements, and composing it for real needs the shared
+    /// configuration file and a database. The header is matched inside the <c>"WalletAPI"</c>
+    /// registration rather than anywhere in the file, so moving it to a different client does not
+    /// keep this green, and the value has to be an identifier — a string literal here would be a
+    /// key checked into a public repository (issue #33).
+    /// </para>
+    /// </summary>
+    [Fact]
+    public void UsersApi_WalletApiNamedClient_SendsTheSharedApiKeyHeader()
+    {
+        var program = File.ReadAllText(Path.Combine(
+            FindRepositoryRoot(), "src", "User", "Users.Api", "Program.cs"));
+
+        Assert.Matches(
+            @"AddHttpClient\(\s*""WalletAPI""[\s\S]*?DefaultRequestHeaders\.Add\(\s*""X-API-Key"",\s*\w+\s*\)",
+            program);
+    }
+
+    /// <summary>
     /// Walks up from the test assembly to the directory holding <c>TallaEgg.sln</c>. The same
     /// anchor <c>ConfigurationPrecedenceTests</c> and <c>SolutionMembershipTests</c> use, and
     /// duplicated for the reason they duplicate it: sharing it would mean editing tests this


MohKardan:
## Self-review — eight findings, four taken

`/code-review 216`. Applied in a393b53. The four taken are all about the line this PR adds; the
four declined are about the five call sites it does not.

### 1. Taken — the placeholder was substituted silently when the key was absent

`APIKeyConstant.TallaEggApiKey` falls back to the public dev placeholder whenever
`TALLAEGG_API_KEY` is unset or misspelled. A Production host that never received the variable —
installed by hand rather than through `install-services.ps1` — would send that placeholder to a
`Wallet.Api` enforcing the real key, and get 401 on every registration with wallets created
lazily afterwards. That is issue #209 exactly, except now the header is present so the code reads
as fixed. The worst possible outcome for this PR.

Users.Api's own `RequireTallaEggApiKey()` at `:104` does not save it: that call sits in an
`AddScheme` options delegate, which runs on first authenticated request, not at startup.

Resolved per environment instead, which is the question the issue asked be checked and answers
the objection my description raised against requiring it:

```csharp
var walletApiKey = builder.Environment.IsProduction()
    ? APIKeyConstant.RequireTallaEggApiKey()
    : APIKeyConstant.TallaEggApiKey;
```

Verified both ways. Production with no key:

```
[FTL] The service is terminating on an unhandled exception.
System.InvalidOperationException: Environment variable 'TALLAEGG_API_KEY' is not set.
```

It never binds a port, and the reason reaches the log file rather than a console no Windows
service has (#205, #211). Development with no key set still starts and runs the full smoke — that
is the run recorded in the description, and it had no `TALLAEGG_API_KEY` in its environment.

### 2. Taken — the key was read inside the configure delegate

The comment twelve lines above says why the address must not be read there: the delegate does not
run until the first client is created, so a guard inside it fails one registration at a time
inside the broad catch in `CreateDefaultWalletsAsync` rather than stopping the service (#205). My
line put a configuration read straight back into it, which also meant an
`Environment.GetEnvironmentVariable` per registration. Hoisted next to `walletApiBaseAddress`,
which finding 1 required anyway.

`StartupGuardPlacementTests` does not catch this — its rule matches `ConfigurationGuard.Require`
only, so an `APIKeyConstant` read inside a delegate is invisible to it. Noting it rather than
widening the rule here.

### 3. Taken — the key was loggable

`IHttpClientFactory`'s logging handlers write full request headers at Trace and redact nothing by
default. This is the solution's only key-bearing client that goes through them; the other four
build raw `HttpClient`s and never reach those handlers. So lowering Serilog to Verbose to debug
this very path — which is how the evidence in this PR was gathered — would have written the live
key to `logs/users-api-*.log`. One chained `.RedactLoggedHeaders(["X-API-Key"])` closes it
permanently.

(No key value appears in the excerpts quoted in the description: Serilog is at Information there
and never logged a header. The exposure was the next person's, not this run's.)

### 4. Taken — the "no test is practical" reasoning in my description was wrong

I claimed asserting on this would need a booted host. It does not.
`WalletApi_MapsDefaultWalletCreation_AsPostNotGet`, twenty lines above where the new test now
sits, reads `Wallet.Api/Program.cs` as text and regex-asserts on it, and
`StartupGuardPlacementTests` does the same across all five `Program.cs` files for exactly this
reason — "there is no host-level startup harness to hang the assertion on". The established
pattern was already there and I talked myself out of using it.

`UsersApi_WalletApiNamedClient_SendsTheSharedApiKeyHeader` matches the header *inside* the
`"WalletAPI"` registration, so moving it elsewhere does not keep it green, and requires an
identifier rather than a string literal, so a key pasted inline fails it too (#33). Confirmed it
fails with the line removed:

```
Failed! - Failed: 1, Passed: 0
Regex: "AddHttpClient\(\s*"WalletAPI"[\s\S]*?Default"...
```

Without it, deleting that one line leaves all 790 tests green — which is how it came to be
missing for the life of the deployment.

### 5–7. Declined — real, but they are #209's shape rather than #209

Three findings ask to fix the class rather than the instance:

- **A shared `X-API-Key` constant.** The literal appears 8 times (5 senders, 2 in
  `ApiKeyAuthenticationHandler`, 1 here). A typo on any sender is invisible in Development and
  surfaces only as a Production 401 — the exact failure mode of this issue.
- **An `AddTallaEggServiceClient` helper**, alongside the existing `AddTallaEggCors` and
  `AddTallaEggErrorHandling`, so the base address, the header and the redaction come as one call.
  This is the depth at which the next client added cannot repeat #209.
- **Validating or trimming the key value.** A trailing CR/LF from a piped file or the `sc.exe`
  registry `MultiString` makes `DefaultRequestHeaders.Add` throw `FormatException`, which lands
  in the broad catch and logs the generic "failed" line — indistinguishable from Wallet.Api being
  down. A trailing space is worse: `Add` trims it, so the sender's key silently differs from the
  receiver's untrimmed comparison and 401s forever.

All three touch the four clients this issue is not about, and the third is an exposure all five
share equally — fixing it only here would create exactly the asymmetry that let #209 happen.
`STANDARDS.md` is explicit that adjacent code stays untouched until a task asks for it, and #214
is already open over these same files. Flagging for the owner to open as its own issue rather
than folding it in.

### 8. Declined — the README statement is not made false by this change

`README.md:189` says every service "throws at startup if it is missing". That was already only
approximately true, on the receiving side: each service's `RequireTallaEggApiKey()` sits in an
`AddScheme` options delegate that runs on first authenticated request. This PR does not weaken it
— after finding 1, Users.Api's sending path is now the one place where it is *literally* true, as
the `[FTL]` above shows. The pre-existing imprecision belongs with whoever fixes the receiving
side, not here.

Worth recording that the same README section, four lines further down, already documents the
`launchSettings.json` trap that made my first reproduction attempt look like the issue was wrong.

Issue #183, opened 2026-09-01
Trade settlement is the one wallet write with no concurrency retry, so a lost race costs 10s and can strand a trade


MohKardan:
Found while fixing #174 (2026-09-01). Not caused by it, but made more likely by it.

## What happens

`WalletRepository.SettleTradeAsync` is the only wallet write with no concurrency retry.
`LockBalanceAsync`, `UnlockBalanceAsync` and `IncreaseBalanceForTradeAsync` all run inside
`WithConcurrencyRetryAsync`; settlement does not. When it loses an optimistic-concurrency race the
`DbUpdateConcurrencyException` escapes to the outbox, which logs a full stack trace and reschedules
the message.

Nothing is lost — the outbox is the safety net and settlement is idempotent on the trade id — but
each loss costs **at least ten seconds** of settlement delay on a trade that is already recorded and
whose collateral is already locked, plus an `Error settling trade` stack trace in the wallet log and
an `Outbox message … failed attempt N` in the orders log. Both read like faults; neither is.

The tail risk is the real concern. `MaxRetries = 5`, so a message that loses five times is marked
`Failed` and logs `SETTLEMENT STUCK` — trade recorded, collateral locked, no settlement, and an
operator has to `POST /api/outbox/{id}/redrive` by hand.

## Why

Same hot row as #174: the market maker is the counterparty to every quote fill, so every settlement
writes its wallet rows. `SettleTradeAsync` writes four wallet rows, four transaction rows and the
settlement row in a single `SaveChanges` inside one transaction, so its window for losing the race
is wider than a lock's.

#174's fix raised the rate. Letting a fill retry for two seconds instead of sixty milliseconds means
the fill wins the row more often, which pushes the collision onto the settlement side. Measured
across the same five simulator runs, before and after that change:

| | settlement conflicts |
|---|---|
| before #182 | 5 |
| after #182 | 31 |

That is the intended direction — the fill path has no safety net and refuses a customer, settlement
has one and self-heals — so this is a trade made on purpose, not a regression. But it is worth
closing the gap on the side that now absorbs more.

Odds of actually stranding a trade look low: roughly 31 conflicts over ~300 settlement attempts, so
about 10% per attempt, and the outbox spaces its five attempts 10/20/40/80s apart — far longer than
a settlement batch, so the attempts are near-independent. Five consecutive losses is on the order of
1e-5. Observed in those runs: 0 outbox failures, all 60 trades settled.

## Where

[...]

PR #222, opened 2026-09-04
[Fix] CreateDefaultWalletsAsync reports the stored rows, and stops flattening every failure — issue #210

### LINKED ISSUE #210: CreateDefaultWalletsAsync: the endpoint's 400 branch is unreachable, and its response is built from discarded entities
Found while reviewing #208 (issue #206). Both are pre-existing, both are in
`WalletService.CreateDefaultWalletsAsync` and the endpoint that calls it, and neither is in
#206's scope.

## 1. The endpoint's `catch (BusinessRuleException)` is unreachable

`src/Wallet/Wallet.Application/WalletService.cs:274-277` wraps everything:

```csharp
catch (Exception ex)
{
    throw new InvalidOperationException("خطا در ایجاد کیف پول‌های پیش‌فرض", ex);
}
```

So the handler's filter at `src/Wallet/Wallet.Api/Program.cs:319` never matches:

```csharp
catch (BusinessRuleException ex)
{
    return Results.BadRequest(ApiResponse<IEnumerable<WalletDTO>>.Fail(ex.Message));
}
```

A SQL timeout, a connection failure, anything at all, escapes as `InvalidOperationException` to
the global handler and comes back as **500 with a fixed body**, `{"success":false,"message":"خطای
داخلی سرور"}` — carrying no information about what went wrong.

Two consequences:

- The `// 400: Wallet creation failed.` line documenting the endpoint (`Program.cs:311`) describes
  a status it cannot return.
- #208 makes `Users.Api` log the response body on a refusal, on the stated grounds that it
  "carries whatever reason the wallet service gave". For this endpoint that reason is always the
  same six words. The wrapping `InvalidOperationException`'s `InnerException` holds what actually
  happened, and it is discarded at the boundary.

## 2. The response is built from entities the repository may have discarded

Same method, `WalletService.cs:248-271`:

```csharp
var irrResult = await _walletRepository.CreateWalletAsync(irrWallet);
wallets.Add(_walletMapper.MapRequired(irrWallet));
```

`irrResult`, `mauaResult` and `creditMauaResult` are assigned and never read — three dead locals.
`WalletRepository.CreateWalletAsync:47-51` returns the **existing** row when the user already has
that wallet, but the mapper is handed the throwaway new entity instead of the return value.

So a second call for a user whose MAUA wallet already holds a balance answers 200 with a wallet
`Id` that exists in no table and `Balance` 0. Harmless today only because the single caller
discards the body. It makes the operation idempotent in the database but not in what it reports,
which is a trap for the next caller.

Fix is to use the return values:

```csharp
var irrResult = await _walletRepository.CreateWalletAsync(irrWallet);
wallets.Add(_walletMapper.MapRequired(irrResult));
```

## 3. Related, worth deciding at the same time

The three wallets are created in three separate `SaveChangesAsync` calls with no transaction, so
a failure on the second leaves the first committed — a user with an IRT wallet and no MAUA or
CREDIT_MAUA. #208's log message was reworded to say the wallets "may be missing or incomplete"
rather than "are missing" precisely because of this. Whether partial creation is worth a
transaction is a real question and not obviously yes: the wallets are created lazily on first
write anyway (`WalletLazyCreationTests`), so a partial set self-heals the same way an empty one
does.

## Scope note

Not fixed in #208 under scope discipline — that PR changes the endpoint's verb and its caller's
logging, and touching the service's exception handling or its DTO mapping would be a different
change with different verification.

MohKardan:
## Description

`WalletService.CreateDefaultWalletsAsync` seeds a new user's Toman, MAUA and CREDIT_MAUA
wallets and is called once per registration. It described the wrong rows and it flattened
every failure into one type and one message. Both are fixed; the issue's third part is
answered by decision rather than by code.

## Issue/Task Reference

Closes #210. Found while reviewing #208 (issue #206); neither defect was in that PR's scope.

## Type of Change

- [ ] Hotfix (urgent bug fix)
- [x] Bug fix, not urgent

The endpoint has exactly one caller and it discards the response body, so nothing is broken
today. Everything here is about what the call *reports* — to the next caller, and to the log.

Confirmed against the tree, not taken from the issue. The issue's line numbers had gone stale.

| Claim | Confirmed at |
| --- | --- |
| Service wraps everything in `InvalidOperationException` | `WalletService.cs:276-279` before this PR |
| Endpoint's `catch (BusinessRuleException)` therefore unreachable | `Wallet.Api/Program.cs:346` before this PR (the issue says `:319`; #208 moved it) |
| `// 400: Wallet creation failed.` documents an impossible status | `Wallet.Api/Program.cs:338` before this PR |
| Three dead locals, mapper handed the argument not the return value | `WalletService.cs:248-272` before this PR |
| `CreateWalletAsync` returns the **existing** row | `WalletRepository.cs:41-51`, and again in the duplicate-race catch at `:62-72` |
| Only caller discards the body | `UserService.CreateDefaultWalletsAsync` — `await`ed, nothing read |

**One correction to the issue.** It says the endpoint answers "with a wallet `Id` that exists
in no table". `WalletDTO` has no `Id` — it carries `Asset`, `Balance`, `LockedBalance` and
`UpdatedAt` only. The defect is real but its visible shape is different: the response reported
`Balance` 0 for a wallet that holds funds, with an `UpdatedAt` stamped at the moment of the
call. Since the DTO has no id, `UpdatedAt` is what says *which row* was described, so the tests
assert it alongside the balances.

## Changes Made

- `src/Wallet/Wallet.Application/WalletService.cs` — all three `MapRequired` calls take the
  repository's return value instead of the entity passed to it. The three locals stop being
  dead, which is the whole fix; a comment says why the distinction matters
- Same file — the `catch (Exception ex) => throw new InvalidOperationException(...)` wrapper is
  removed. #134's precedent: catch blocks that duplicate `GlobalExceptionHandler` were deleted
  across nine `Orders.Api` endpoints, and this one is worse than a duplicate — it replaces the
  exception's identity with a fixed string on the way past
- `src/Wallet/Wallet.Api/Program.cs` — the endpoint's `catch (BusinessRuleException)` and the
  `// 400:` line are removed, and the `Returns:` line no longer claims every wallet was created
- The XML doc's `<returns>` now says what the method returns: the rows created **or** the rows
  that already existed
- Five tests in `tests/TallaEgg.AllServices.Tests/DefaultWalletCreationResponseTests.cs`

### Why the endpoint's catch was removed rather than left in place

Removing the service's wrapper does not by itself make the 400 branch reachable. Walking the
path after the fix, what can escape `CreateDefaultWalletsAsync` is `ArgumentException` (from
`WalletEntity.Create`, on `Guid.Empty`), whatever the database raises, and nothing else.
`BusinessRuleException` is not among them. Keeping the filter would leave exactly the defect
the issue reports — a branch that cannot run and a comment promising a status that cannot be
returned — with only the reason changed.

The argument the other way is real and worth naming for the reviewer: seven sibling endpoints
in the same file carry that catch, and the endpoint is arguably defending against
`IWalletService` the interface, not `WalletService` the implementation. It is dropped here
because a guard for something that cannot happen is not defence, it is documentation that is
wrong. If a domain rule is ever added to this path, the catch comes back with the rule.

Not touched, deliberately: `Guid.Empty` still produces a 500 rather than a 400. Turning bad
input into a domain rule would be new behavior, and the issue does not ask for it.

## Part 3 — no transaction, by decision

The issue's third part asks whether the three wallets, created through three separate
`SaveChangesAsync` calls with no transaction, need one — and says up front that the answer is
not obviously yes. Asked and answered by the product owner: **no transaction**. Nothing is
built for it here, and the reasoning is recorded rather than deferred:

- **A transaction would make the failure mode slightly worse, not better.** A partial set and
  an empty set both self-heal, because every *write* path creates a missing wallet on demand
  (`WalletLazyCreationTests`). The one surface that does not self-heal is a pure read —
  `WalletService.GetBalanceAsync` throws when there is no row. With a partial set some reads
  still work; with the empty set a rollback would leave, none do at all.
- **Nothing depends on the invariant.** No query, guard or report asserts "all three or none".
- **No value is at stake.** All three rows are created with `Balance` and `LockedBalance` at
  zero. Atomicity earns its keep where partial application leaves money in a wrong state; here
  it leaves three empty ledgers.
- **The cost lands on sensitive code.** The service holds no `DbContext`, so the transaction
  would have to move into `WalletRepository` as a new `IWalletRepository` member — next to the
  constraint that file already documents: an operation must not leave a transaction open when
  it throws, because the `ChangeTracker.Clear()` between concurrency retries is not sound
  inside an open `IDbContextTransaction` (#183).
- **`CreateWalletAsync` has create-or-return-existing semantics**, including a duplicate-race
  catch that re-reads. That combination inside one transaction is more fragile, not less.

The strongest case for "yes" is that if lazy creation were ever narrowed, a partial set would
become a durable defect with no repair surface — #206 decided against building one. That is a
reason to keep `WalletLazyCreationTests` green, which it is, not a reason to add a transaction
now. Since part 3 closes by decision, **nothing is left to build** and this PR says
"closes #210".

### Build

```
```

`TreatWarningsAsErrors` is on and `CS0219` (assigned but never used) is among the codes #134
promoted to errors. The three locals did not trip it before the fix — they were assigned — so
the build is not what would have caught this.

### Unit tests

```
dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 832, Skipped: 0, Total: 832
```

Five new, in `DefaultWalletCreationResponseTests`:

| Test | What it pins |
| --- | --- |
| `..._UserAlreadyHasAFundedWallet_ReportsTheStoredRow` | the defect itself: deposit 100 MAUA, lock 40, call again, and the response says 60/40 with the **stored** `UpdatedAt` |
| `..._CalledTwice_ReportsAllThreeDefaultWallets` | the deposit is reported against MAUA only, so the test above cannot pass by accident |
| `..._CalledTwice_CreatesNoDuplicateRows` | the database side was already idempotent and stays that way |
| `..._NewUser_ReturnsTheThreeWalletsItJustCreated` | the first-call path, where argument and return value are the same object, is unchanged |
| `..._RepositoryFails_PropagatesTheOriginalException` | an exact-type assertion: whatever fails underneath arrives as itself, with a null `InnerException` |

In-memory SQLite for the four that need a database, a hand-written `IWalletRepository` whose
other fourteen members throw for the fifth — no mocking library, per `STANDARDS.md`.

### The new tests were shown to fail

With the two source files stashed and the tests left in place, the two that pin new behaviour
go red and the three that pin unchanged behaviour stay green:

```
Failed  CreateDefaultWalletsAsync_UserAlreadyHasAFundedWallet_ReportsTheStoredRow
        Assert.Equal() Failure: Values differ   Expected: 60   Actual: 0
Failed  CreateDefaultWalletsAsync_RepositoryFails_PropagatesTheOriginalException
        Assert.Throws() Failure: Exception type was not an exact match
        Expected: StorageUnavailableException
        Actual:   System.InvalidOperationException
        ---- System.InvalidOperationException : [the six fixed words]
        -------- StorageUnavailableException : the storage layer is unavailable

Failed! - Failed: 2, Passed: 3, Skipped: 0, Total: 5
```

### Part 2 against the running service, over HTTP and SQL Server

`Wallet.Api` was brought up on a throwaway port against a throwaway database
(`--ConnectionStrings:WalletDb=...;Database=<scratch>` on the command line, which #159 made
outrank the shared file), so nothing touched the real wallet database. A user was created, given
100 MAUA, and `POST /api/wallet/create-default/{id}` called a second time. The same request
against a build with the fix stashed:

```
before:  {"asset":"MAUA","balance":0,     "updatedAt":"2026-09-04T10:09:38.28..."}   <- the moment of the call
after:   {"asset":"MAUA","balance":100.0, "updatedAt":"2026-09-04T10:07:36.32..."}   <- the stored row
```

The scratch database was dropped afterwards; `sys.databases` was checked to confirm only the
four real ones remain.

### Part 1 against a real failure

Same scratch database, taken offline mid-run with `ALTER DATABASE ... SET OFFLINE WITH ROLLBACK
IMMEDIATE`, then the same POST. Both builds, from `logs/wallet-api-*.log` (host identifiers
redacted):

```
before:
[ERR] Unhandled exception. TraceId=... Method=POST Path=/api/wallet/create-default/...
System.InvalidOperationException: خطا در ایجاد کیف پول‌های پیش‌فرض
 ---> System.InvalidOperationException: An exception has been raised that is likely due to a
      transient failure. Consider enabling transient error resiliency ...
 ---> Microsoft.Data.SqlClient.SqlException (0x80131904): Cannot open database "..." requested
      by the login. The login failed.

after:
[ERR] Unhandled exception. TraceId=... Method=POST Path=/api/wallet/create-default/...
System.InvalidOperationException: An exception has been raised that is likely due to a
transient failure. Consider enabling transient error resiliency ...
 ---> Microsoft.Data.SqlClient.SqlException (0x80131904): Cannot open database "..." requested
      by the login. The login failed.
```

Two things stated plainly rather than claimed away:

- **The HTTP response did not change.** 500 with `{"success":false,"message":"..."}` before and
  after — `GlobalExceptionHandler` writes that body regardless of what reached it. So the
  response body #208 logs in `Users.Api` is no more informative than it was; what improves is
  `Wallet.Api`'s own log, where the cause was always recoverable but only through
  `InnerException`.
- **EF Core does its own wrapping.** For this particular failure the top-level type is still
  `InvalidOperationException` — EF's, with EF's diagnostic message. The change is that the first
  line now names the actual failure mode instead of six fixed words, and there is one less level
  to unwrap. For an exception EF does not wrap, the type on that line is the real one.

### End to end

```
=== Done in 00:00:32.7. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:
  BTC/IRT: 17 filled
  MAUA/IRT: 17 filled
  SEKE_BAHAR/IRT: 16 filled
```

All 20 registrations went through this endpoint. One of them, queried directly afterwards:

```
Asset              Balance          CreatedAt
CREDIT_MAUA        120.00000000     2026-09-04 10:11:41.6301603
IRT                6020584830.0     2026-09-04 10:11:41.5991085
MAUA               121.54000000     2026-09-04 10:11:41.6145123
```

The three defaults, created within milliseconds of each other at registration. (The other four
rows on that user are lazily created ones for BTC and SEKE_BAHAR, from the trading phase.)

### CI doc-path check (#207)

```
scripts/check-doc-paths.sh
check-doc-paths: 392 tracked text file(s) checked, every reference resolves.
```

### Environment

## Security Checklist

- [x] No secrets/tokens in the diff — `config/appsettings.global.json` untouched and uncommitted

Worth a reviewer's eye: this PR removes a wrapper, which means a raw exception message now
reaches the log one level higher than before. It reached the log either way — the wrapper
preserved the inner exception and `GlobalExceptionHandler` logs the whole chain — so nothing
newly appears there. Nothing newly reaches a *client*: the response body is written by the
global handler and does not include the exception.

## Code Quality

- [x] The Persian success message users' registrations depend on is untouched. The Persian
      string this PR deletes was never user-facing — it lived in an exception message the
      global handler replaces before anything leaves the process
- [x] No new dependencies

## Documentation

No `.md` changes. Nothing in `docs/` documents this route or this method. `AGENT.md`'s port
table and configuration rules are unaffected. Swagger reads the shape from the mapping, which is
unchanged apart from the removed catch.

## Breaking Changes?

- [x] No breaking changes

The endpoint's status codes are unchanged in practice: the 400 that is removed could never be
returned. The response body's *content* changes for the repeat-call case — it now reports real
balances instead of zeros — which is the fix. The single caller discards the body.

## Deployment Notes

No migrations, no new configuration. `Wallet.Api` only; `Users.Api` is untouched.

**Post-deployment verification**: register a user and confirm three wallet rows appear, exactly
as for #208.

**Rollback**: revert the commit. There is no schema or contract change to unwind.

## Related Issues

- Opened from #208's self-review, alongside #209 (which merged as #216)
- Follows #134, whose precedent — do not re-handle what `GlobalExceptionHandler` handles — is
  what part 1 applies
- Part 3's reasoning rests on `WalletLazyCreationTests` and on #206's decision not to build a
  repair surface
- #214 (DI wiring) touches the same area and is deliberately not in scope; it merged as #221,
  which this branch is cut from

### DIFF
--- a/src/Wallet/Wallet.Api/Program.cs
+++ b/src/Wallet/Wallet.Api/Program.cs
@@ -331,22 +331,17 @@ app.MapGet("/api/wallet/transactions/{userId}", async (Guid userId, string? asse
 // POST, not GET: this creates rows, and GET is the verb every intermediary assumes it may
 // retry or prefetch freely. A repeat creates no duplicate rows — CreateWalletAsync keeps the
 // existing wallet — but the verb should still say what the call does. Issue #206.
+// Nothing on this path throws BusinessRuleException, so the catch that used to sit here could
+// never run and its "400" documented a status the endpoint cannot return. A failure now reaches
+// GlobalExceptionHandler, which logs it with its own type and a trace id. Issue #210.
 // userId: User id.
 // walletService: Wallet service.
-// Returns: The wallets that were created.
+// Returns: The user's default wallets, created or already existing.
 // 200: Default wallets created.
-// 400: Wallet creation failed.
 app.MapPost("/api/wallet/create-default/{userId}", async (Guid userId, IWalletService walletService) =>
 {
-    try
-    {
-        var wallets = await walletService.CreateDefaultWalletsAsync(userId);
-        return Results.Ok(ApiResponse<IEnumerable<WalletDTO>>.Ok(wallets, "کیف پول‌های پیش‌فرض با موفقیت ایجاد شدند"));
-    }
-    catch (BusinessRuleException ex)
-    {
-        return Results.BadRequest(ApiResponse<IEnumerable<WalletDTO>>.Fail(ex.Message));
-    }
+    var wallets = await walletService.CreateDefaultWalletsAsync(userId);
+    return Results.Ok(ApiResponse<IEnumerable<WalletDTO>>.Ok(wallets, "کیف پول‌های پیش‌فرض با موفقیت ایجاد شدند"));
 })
 .WithTags("Wallets");
 
--- a/src/Wallet/Wallet.Application/WalletService.cs
+++ b/src/Wallet/Wallet.Application/WalletService.cs
@@ -238,45 +238,44 @@ public class WalletService : IWalletService
     /// Every other asset's wallet is created lazily on first deposit — see IncreaseBalanceAsync.
     /// </summary>
     /// <param name="userId">User id.</param>
-    /// <returns>The wallets that were created.</returns>
+    /// <returns>
+    /// The user's default wallets: the rows just created, or the rows that already existed.
+    /// </returns>
     public async Task<IEnumerable<WalletDTO>> CreateDefaultWalletsAsync(Guid userId)
     {
         var wallets = new List<WalletDTO>();
 
-        try
-        {
-            var irrWallet = WalletEntity.Create
-            (
-                 userId,
-                 CurrenciesConstant.Toman
-            );
-            var irrResult = await _walletRepository.CreateWalletAsync(irrWallet);
-            wallets.Add(_walletMapper.MapRequired(irrWallet));
+        // Map the row the repository returns, not the entity handed to it. CreateWalletAsync
+        // returns the *existing* row when the user already has that wallet and discards the new
+        // entity, whose Balance is 0 by construction, so mapping the argument reports an empty
+        // wallet for one that holds funds. Issue #210.
+        var irrWallet = WalletEntity.Create
+        (
+             userId,
+             CurrenciesConstant.Toman
+        );
+        var irrResult = await _walletRepository.CreateWalletAsync(irrWallet);
+        wallets.Add(_walletMapper.MapRequired(irrResult));
 
 
-            var mauaWallet = WalletEntity.Create
-            (
-                 userId,
-                 CurrenciesConstant.Maua
-            );
-            var mauaResult = await _walletRepository.CreateWalletAsync(mauaWallet);
-            wallets.Add(_walletMapper.MapRequired(mauaWallet));
+        var mauaWallet = WalletEntity.Create
+        (
+             userId,
+             CurrenciesConstant.Maua
+        );
+        var mauaResult = await _walletRepository.CreateWalletAsync(mauaWallet);
+        wallets.Add(_walletMapper.MapRequired(mauaResult));
 
 
-            var creditMauaWallet = WalletEntity.Create
-            (
-                 userId,
-                 CurrenciesConstant.Credit_MAUA
-            );
-            var creditMauaResult = await _walletRepository.CreateWalletAsync(creditMauaWallet);
-            wallets.Add(_walletMapper.MapRequired(creditMauaWallet));
+        var creditMauaWallet = WalletEntity.Create
+        (
+             userId,
+             CurrenciesConstant.Credit_MAUA
+        );
+        var creditMauaResult = await _walletRepository.CreateWalletAsync(creditMauaWallet);
+        wallets.Add(_walletMapper.MapRequired(creditMauaResult));
 
-            return wallets;
-        }
-        catch (Exception ex)
-        {
-            throw new InvalidOperationException("خطا در ایجاد کیف پول‌های پیش‌فرض", ex);
-        }
+        return wallets;
     }
 
     public Task<(bool Success, string Message)> SettleTradeAsync(
--- a/tests/TallaEgg.AllServices.Tests/DefaultWalletCreationResponseTests.cs
+++ b/tests/TallaEgg.AllServices.Tests/DefaultWalletCreationResponseTests.cs
@@ -1,0 +1,204 @@
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging.Abstractions;
+using TallaEgg.Core;
+using Wallet.Application;
+using Wallet.Application.Mappers;
+using Wallet.Core;
+using Wallet.Infrastructure;
+
+namespace TallaEgg.AllServices.Tests;
+
+/// <summary>
+/// <c>WalletService.CreateDefaultWalletsAsync</c> seeds a new user's Toman, MAUA and CREDIT_MAUA
+/// wallets, and is called once per registration. Two defects in what it reported are fixed here
+/// (issue #210), both found while reviewing #208.
+///
+/// <para>
+/// <b>The response was built from entities the repository had discarded.</b>
+/// <c>WalletRepository.CreateWalletAsync</c> returns the <i>existing</i> row when the user
+/// already has that wallet, but the mapper was handed the throwaway new entity instead of the
+/// return value. A repeat call for a funded user therefore answered 200 describing an empty
+/// wallet: idempotent in the database, but not in what it said. <c>WalletDTO</c> carries no id,
+/// so <c>UpdatedAt</c> is the field that says which row was described, and it is asserted
+/// alongside the balances.
+/// </para>
+///
+/// <para>
+/// <b>Every failure was wrapped in <c>InvalidOperationException</c>.</b> That flattened a SQL
+/// timeout, a dropped connection and a bug into one type and one fixed message, made the
+/// endpoint's <c>catch (BusinessRuleException)</c> unreachable, and left the real cause reachable
+/// only through <c>InnerException</c>. The wrapper is gone, on the precedent set by #134: what
+/// <c>GlobalExceptionHandler</c> already handles is not re-handled on the way up.
+/// </para>
+/// </summary>
+public class DefaultWalletCreationResponseTests : IDisposable
+{
+    private readonly SqliteConnection _connection;
+
+    public DefaultWalletCreationResponseTests()
+    {
+        _connection = new SqliteConnection("DataSource=:memory:");
+        _connection.Open();
+
+        using var setup = NewContext();
+        setup.Database.EnsureCreated();
+    }
+
+    public void Dispose() => _connection.Dispose();
+
+    private WalletDbContext NewContext() =>
+        new(new DbContextOptionsBuilder<WalletDbContext>().UseSqlite(_connection).Options);
+
+    private static WalletRepository NewRepository(WalletDbContext context) =>
+        new(NullLogger<WalletRepository>.Instance, context);
+
+    private static WalletService NewService(IWalletRepository repository) =>
+        new(repository, new WalletMapper(), NullLogger<WalletService>.Instance);
+
+    /// <summary>
+    /// The defect itself. A second call for a user whose MAUA wallet already holds funds must
+    /// describe the stored row, not the discarded one it tried to insert.
+    /// </summary>
+    [Fact]
+    public async Task CreateDefaultWalletsAsync_UserAlreadyHasAFundedWallet_ReportsTheStoredRow()
+    {
+        using var context = NewContext();
+        var repository = NewRepository(context);
+        var service = NewService(repository);
+        var userId = Guid.NewGuid();
+
+        await service.CreateDefaultWalletsAsync(userId);
+        await service.DepositAsync(userId, CurrenciesConstant.Maua, 100m);
+        await repository.LockBalanceAsync(userId, CurrenciesConstant.Maua, 40m);
+
+        var second = await service.CreateDefaultWalletsAsync(userId);
+
+        using var verify = NewContext();
+        var stored = await verify.Wallets.SingleAsync(
+            w => w.UserId == userId && w.Asset == CurrenciesConstant.Maua);
+
+        var reported = Assert.Single(second, w => w.Asset == CurrenciesConstant.Maua);
+        Assert.Equal(60m, reported.Balance);
+        Assert.Equal(40m, reported.LockedBalance);
+
+        // What identifies the row, in the absence of an id on the DTO: a freshly built
+        // WalletEntity carries DateTime.UtcNow, never the stored row's timestamp.
+        Assert.Equal(stored.UpdatedAt, reported.UpdatedAt);
+    }
+
+    /// <summary>
+    /// The other two default wallets are untouched by the deposit, so their stored balance is
+    /// zero either way. Asserted so the test above cannot pass on a service that reports the
+    /// deposit against every wallet it returns.
+    /// </summary>
+    [Fact]
+    public async Task CreateDefaultWalletsAsync_CalledTwice_ReportsAllThreeDefaultWallets()
+    {
+        using var context = NewContext();
+        var service = NewService(NewRepository(context));
+        var userId = Guid.NewGuid();
+
+        await service.CreateDefaultWalletsAsync(userId);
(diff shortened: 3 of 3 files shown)


MohKardan:
## Self-review — `/code-review 222`

Ten findings. Two applied in `021151d`, both against this PR's own tests. Eight not applied,
with reasons; three of those are real pre-existing defects worth their own issues, and one is a
factual correction to the finding.

### Applied

**1. The regression guard covered one of the three mapping fixes.** Correct, and the sharpest
finding here. Only MAUA was funded, so Toman and CREDIT_MAUA were asserted at `Balance` 0 —
exactly the value the bug produced. Reverting those two `MapRequired` calls left all five tests
green, so two thirds of the fix had no guard at all. All three wallets now carry a distinct
non-zero balance (5000 IRT, 100 MAUA with 40 locked, 250 CREDIT_MAUA). Verified by reverting
only those two mappings and re-running:

```
Failed  CreateDefaultWalletsAsync_UserAlreadyHasFundedWallets_ReportsTheStoredRows
Failed! - Failed: 1, Passed: 4, Skipped: 0, Total: 5
```

`..._CalledTwice_ReportsAllThreeDefaultWallets` is folded into it — it existed to stop the
funded-wallet test passing by accident, and a test that asserts all three distinct balances
does that itself.

**2. Nothing exercised a failure on the *second* wallet.** Correct, and it is the state part 3's
decision is entirely about: `ThrowingRepository` throws on every call, so the exception escaped
before any row was committed and the partial set was never constructed. Added
`..._RepositoryFailsOnTheSecondWallet_LeavesTheFirstCommitted`, which wraps the real SQLite-backed
repository, forwards the first create and refuses the second, then asserts exactly one row — the
IRT one — survives. That turns "no transaction, by decision" into something executable: adding a
transaction here now fails a test that names the decision, instead of passing silently.

### Not applied — real, pre-existing, out of scope

**3. The failed `Added` entity poisons the next `SaveChangesAsync`.** Confirmed, and the most
serious thing this review turned up. The unique index is real
(`WalletDbContext.cs:34`, `(UserId, Asset)`), so `CreateWalletAsync`'s duplicate-race catch at
`WalletRepository.cs:62-72` is reachable; EF only calls `AcceptAllChanges` on success, so the
entity whose insert failed stays `Added` on the shared `DbContext` and is re-flushed by the next
`SaveChangesAsync`, taking the second wallet's insert down with it. Two concurrent calls for one
user end with a 500 and no MAUA or CREDIT_MAUA row.

Pre-existing and untouched by this PR — the same three calls on the same context before and
after. It also cuts against part 3's premise in a way worth recording: the argument for leaving
the three creates untransacted assumes a partial set is the worst case, and this path can
produce one from a race rather than from an outage. Neither the fix nor the analysis belongs in
a PR about what the method reports. **Opening as its own issue.**

**4. `GetWalletAsync` is not sargable.** Confirmed: `w.Asset.ToUpper() == asset.ToUpper()`
(`WalletRepository.cs:24`) translates to `UPPER(Asset)` and cannot seek the unique index, so each
of the three pre-checks scans. Pre-existing, on every wallet read in the service, not just this
path. **Rolling into the same issue** — it is one repository method and the two findings meet
there.

**5. Six sequential round trips per registration.** Same method, same issue. Worth saying that
the fix the finding proposes (one `GetUserWalletsAsync`, then `AddRange` + a single
`SaveChanges`) is also the transaction from part 3 arriving by the back door, so it is a design
question for the owner and not a tidy-up.

### Not applied — considered and decided otherwise

**6. Removing `catch (BusinessRuleException)` leaves this the only endpoint of eight without a
400.** This is the tradeoff the PR body already argues, and the finding states the other side of
it fairly. Restating the decision: the issue's complaint is a branch that cannot run and a
comment promising a status that cannot be returned. Keeping the filter leaves both, changing only
*why* the branch is unreachable. The finding is right that no test pins the "nothing here throws
`BusinessRuleException`" assumption, and no test can — an assumption about which exceptions a
future implementation will not throw has nothing to assert against. If a currency guard is added
here, as the sibling paths have, the catch comes back in the same commit as the guard, which is
where it belongs. Left as is; the reviewer can overrule.

**7. `Asset` is now echoed from the stored row rather than the canonical constant.** The
mechanism is right — `IsValidCurrency` reads an `OrdinalIgnoreCase` dictionary, `GetWalletAsync`
compares case-insensitively, so a lazily-created `"maua"` row can exist and would now be echoed
with that casing — but the conclusion overstates it. Every other read already behaves this way:
`GetBalanceAsync` maps the stored entity, and `DepositAsync` returns `result.walletEntity.Asset`
verbatim (`WalletService.cs:206`). This PR makes `create-default` consistent with them rather
than introducing a new hazard, and reporting the row as it is stored is the point of the fix. If
non-canonical casing can reach a wallet row, the bug is that lazy creation does not normalize the
asset code, which is a different fix on a different path.

**8. The 200 message still says the wallets were "successfully created".** Fair, and it does now
sit slightly against the `<returns>` this commit corrected. Not changed: it is a pre-existing
Persian string, no caller reads the body, and rewording an API's Persian message is its own
change with its own review. Noted for whoever touches this endpoint next.

**9. Three copy-pasted blocks; replace with a loop.** and **10. Use
`WalletMapper.Map(IEnumerable<WalletEntity>)` instead of three `MapRequired` calls.** Both are
fair readings of the code and both are refactors nobody asked for.
`STANDARDS.md` §3: *"Never rewrite or refactor code that was not part of the request."* The
finding's own point — that the triplication is what let the same mistake be made three times —
is a good argument for a follow-up, and a bad reason to widen a bug fix that has to be reviewed
against a live service.

**11. `irrWallet`/`irrResult` name the Toman wallet after the rial.** True, and a leftover from
the IRR→IRT migration (`scripts/migrate-irr-to-irt.sql`). Renaming is cheap here because the
lines were rewritten anyway, which is exactly the reasoning that turns a bug fix into a diff
nobody can review by eye. Same answer as 9 and 10 — flagged, not fixed.

### Re-verified after the change

```

dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 832, Skipped: 0, Total: 832
```

MohKardan:
Follow-up from this review opened as #223 — the `Added` entity that survives a failed insert, plus the two `WalletRepository` findings that sit in the same two methods. Nothing there is caused or fixed by this PR.

Issue #218, opened 2026-09-04
Expose the running version at runtime (e.g. a /version endpoint), now that it means something


MohKardan:
## Why

Follow-up to #217. "Which build is on the server?" only has an answer if something running can be asked. Right now nothing can.

## What's already free

No extra wiring needed: with `<VersionPrefix>` now set in `Directory.Build.props` (#217) and no `SourceLink` package installed anywhere in the repo, a plain `dotnet build` inside this git checkout already embeds the commit hash — confirmed by reading a built DLL's metadata after #217:

```
InformationalVersion: 1.1.0+86563935ed8a05e173290fdaf07f2a8ecc7dd317
```

That's the .NET SDK's own built-in behavior when a `.git` directory is discoverable, not something #217 added on purpose. The commit hash is already sitting in every built assembly today.

## What's missing

A way to read it from a *running* service without RDP/SSH and a file-properties dialog — e.g.:

- Log it once at startup in each of the four services (Users.Api, Wallet.Api, Orders.Api, the bot), the way the rest of the boot sequence already logs to Serilog.
- And/or a cheap `GET /version` (or `/health` extended to include it) on each API returning `{ version, commitHash }`, read via `Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>()`.

## Scope note

Deliberately not built as part of #217 — that PR is versioning-and-release only. This is a small, separate, low-risk addition once someone decides it's worth doing.

Issue #52, opened 2026-07-27
LockedBalance is never fully released: lock uses full-precision price, fills use the rounded stored price


MohKardan:
## Summary

`LockedBalance` is never fully released. An order locks collateral computed from a **full-precision** price, but every fill consumes collateral computed from the **rounded** price that the `Orders.Price` column can actually store. The difference stays locked forever, on every order, growing with volume.

Observed on real data on 2026-07-27: **2 IRT stranded** on a single market-maker order.

## Why this is being filed

Three reasons, in order of importance:

1. **It is a slow leak in a financial ledger, and it never self-corrects.** Each stranded amount is small, but nothing ever releases it. Collateral that a user has posted becomes permanently unusable to them, and the wallet's `LockedBalance` drifts away from the sum of its actual obligations. Over time no one can tell whether a non-zero `LockedBalance` means "real open exposure" or "accumulated residue".
2. **It breaks the invariant that future reconciliation will depend on.** The natural check — *for every wallet, `LockedBalance` equals the collateral required by that user's open orders* — is already false today. Any alerting built on it would fire constantly and be switched off, which is how real problems get missed later.
3. **The codebase already identified this exact failure mode and the fix was applied one layer too shallow.** The XML doc on `CurrenciesConstant.RoundToCurrencyPrecision` explicitly warns that a mismatch between the locked amount and the settled amount leaves "یک باقی‌ماندهٔ کوچک در LockedBalance". That helper was correctly applied to the `quantity × price` multiplication — but the **price itself** still differs between the two sides of the comparison, so the mismatch survived.

This is not urgent (fee rates are zero, magnitudes are tiny, no user is currently harmed) but it should be fixed before launch, because retrofitting a ledger invariant after real balances exist is much harder than starting with it true.

## Evidence from production data

Market-maker buy order `547FC451`, admin entered **79,000,000 IRT per mesghal** for 1000 g:

| | Value |
|---|---|
| Locked at order placement | `18,237,222,402` IRT |
| `1000 × Orders.Price` as stored | `18,237,222,400` IRT |
| **Difference** | **2 IRT** |

After one 10 g fill (trade `B3451E86`, `182,372,224` IRT consumed):

```
LockedBalance = 18,054,850,178
990 g × 18,237,222.40 = 18,054,850,176      ← the same 2 IRT, still stranded
```

## Root cause

Five steps, each individually reasonable:

[...]

PR #226, opened 2026-09-04
[Docs] Describe every Wallet and Users endpoint, and remove three that could not be — issue #191

### LINKED ISSUE #191: Wallet.Api and Users.Api endpoints carry no Swagger descriptions
## Context

#189 deleted five hand-written API documents on the grounds that the live Swagger UIs had superseded them. That is true for one service and only partly true for the other two. Measured:

| Service | `WithSummary` | `WithDescription` |
|---|---|---|
| `Orders.Api` | 11 | 3 |
| `Wallet.Api` | **0** | **0** |
| `Users.Api` | **0** | **0** |

`IncludeXmlComments` does not close the gap: it cannot pick up XML comments written above minimal-API lambdas.

## Impact

The generated page for Wallet and Users is a list of paths with request/response schemas and no prose — nothing that says what an endpoint does, when it fails, or what a given error means. The documents removed in #189 did carry that layer, including a table of the Persian error messages the wallet returns (`موجودی کافی نیست`, `کیف پول پیدا نشد`, `مقدار باید بزرگتر از صفر باشد`).

Separately, all three services mount Swagger inside `if (app.Environment.IsDevelopment())` (`Wallet.Api/Program.cs:160`, `Users.Api/Program.cs:209`, `Orders.Api/Program.cs:264`). That is the right call — Swagger has no consumer in Production — but it means a deployed server has no API documentation at all. Noting it here for completeness; it is not proposed for change.

## Proposed work

Add `.WithSummary()`, and `.WithDescription()` where an endpoint has a rule worth stating, to the endpoints of `Wallet.Api` (10) and `Users.Api` (15). `Orders.Api` already establishes the house style, so this is filling in a pattern the repository has rather than inventing one.

Worth covering explicitly, because they are the things a schema cannot express:

- `lockBalance` / `unlockBalance` — that `LockBalance` enforces no balance rule by design, the ceiling living in the separate `CREDIT_<ASSET>` wallet row (see `CLAUDE.md`).
- `deposit` / `increaseBalance` / `changeBalance` — the `ReferenceId` idempotency contract added in #157.
- The credit ledger: `CREDIT_<ASSET>` exists per tradable base asset, and `CREDIT_IRT` deliberately does not exist.

## What this issue does not propose

Restoring the deleted documents. They were wrong — `WALLET_API_SWAGGER.md` documented two endpoints that do not exist and omitted four that do. Descriptions belong next to the code that they describe, so that they cannot drift out of sync again the way those files did.

MohKardan:
## Description

`Wallet.Api` and `Users.Api` had **0** `WithSummary` and **0** `WithDescription` between them, so
their Swagger pages were paths and schemas with no prose. `Orders.Api` already establishes the
house pattern, so this fills one in rather than inventing one. `IncludeXmlComments` cannot close
the gap — it does not pick up XML comments above minimal-API lambdas, which is exactly why #132
converted those to plain `//` comments.

Three endpoints were **removed instead of described**, on the product owner's instruction, because
describing them truthfully would have documented behaviour nobody should rely on.

## Issue/Task Reference

Closes #191

## Type of Change

- [x] Documentation (docs/comments only)
- [x] Fix (removes three endpoints — see below)

### Descriptions added

Every remaining endpoint in both services carries a summary, and a description wherever there is a
rule a schema cannot express. Counted against the code, not the issue: `Wallet.Api` has 10 mapped
endpoints today and `Users.Api` 13 (the issue's figures predate `/version` from #218 and the
endpoint #204 removed).

Covered explicitly, as the issue asked:

- **`lockBalance` / `unlockBalance`** — that `LockBalance` deliberately enforces no *sufficiency*
  rule, the ceiling living in the separate `CREDIT_<ASSET>` row a single-asset write cannot see;
  and that `unlockBalance`, by contrast, *is* guarded against the stored `LockedBalance` (#52).
- **`deposit` / `withdrawal`** — the `ReferenceId` idempotency contract from #157, including that a
  request without a reference is never deduplicated.
- **The credit ledger** — `CREDIT_<ASSET>` per tradable base asset, and `create-default` seeding
  only IRT/MAUA/CREDIT_MAUA with every other wallet created lazily on first write.
- **`changeBalance`** — exactly-once resting on the `TradeSettlements` primary key rather than on
  code order, so outbox redelivery is safe; plus the self-trade, non-zero-fee (#35) and
  locked-collateral refusals.

### Endpoints removed

- **`POST /api/wallet/increaseBalance`** — accepted a `ReferenceId` and dropped it. It called the
  three-argument `IncreaseBalanceAsync`, so `refId` defaulted to `null` and the #157 deduplication
  never ran: a non-idempotent twin of `deposit` advertising the very field that would have made it
  idempotent. No production caller; the live admin top-up path goes through `deposit`/`withdrawal`.
- **`POST /api/user/validate-invitation`** — could not succeed. It reached
  `UserRepository.GetInvitationByCodeAsync`, whose body was `throw new NotImplementedException()`,
  and `UsersDbContext` maps no `Invitations` set, so any non-empty code answered 500. No caller.
- **`POST /api/user/register-with-invitation`** — validated no invitation despite its name, and took
  the raw `User` entity as its body, so a caller picked their own `Id`, `Status` and `Role`. No caller.

Support code that existed only for those routes goes with them — a client method aimed at a deleted
route is worse than none: two client methods, two request records, the unreachable service and
repository members, and the matching test-double members.
`UserService.RegisterUserAsync(User)` is **kept**: it is the seam eight tests in
`DefaultWalletCreationFailureTests` use.

### Comment blocks

The per-endpoint `//` blocks restating parameters and status codes are replaced by the metadata
rather than left beside it, so there is one description per endpoint instead of two that can drift.
Comments explaining *why* code is as it is are untouched. Several of the replaced blocks were
already wrong — three claimed `404` for lookups that answer `400`.

### Unit tests

```
dotnet build TallaEgg.sln

dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 836, Skipped: 0, Total: 836
```

### End-to-end

`driver.ps1 start`, then `driver.ps1 smoke --users 20 --quotes 20 --trades 50 --seed 7`:

```
=== Done in 00:00:19.8. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:  BTC/IRT: 17   MAUA/IRT: 17   SEKE_BAHAR/IRT: 16
```

Date/time run: 2026-09-04.

### Swagger actually renders the text

This is the point #132 turned on — comments were written that Swagger never saw. Read back off the
**generated OpenAPI documents** of the running services:

```
WALLET : 10 of 10 operations carry a summary
USERS  : 13 of 13 operations carry a summary
ORDERS : 11 of 28 operations carry a summary   (unchanged, out of scope)

deleted endpoints still present?
  wallet increaseBalance:         False
  users validate-invitation:      False
  users register-with-invitation: False
```

One sample per service, verbatim from `/swagger/v1/swagger.json`:

**Wallet — `POST /api/wallet/lockBalance`**
> **Reserve funds against an open order**
> Moves the amount out of Balance and into LockedBalance. It deliberately enforces no sufficiency
> rule, and must not: customers trade on credit and are expected to go negative down to a ceiling
> held in a separate CREDIT_&lt;ASSET&gt; wallet row, which a single-asset write cannot see, so that
> check belongs in the caller that can read both rows. Only the amount itself is validated — it has
> to be positive, or a negative one would mint money by running the arithmetic backwards. Takes no
> reference and is therefore not idempotent; a repeat locks again. Retries internally when it loses
> an optimistic-concurrency race.

**Users — `POST /api/user/register`**
> **Register a Telegram user**
> Requires an invitation code that some existing user already holds; an unknown code is refused with
> «کد دعوت معتبر نیست.». The new user starts Pending with the RegularUser role and is given an
> invitation code of their own. Default wallets are requested from Wallet.Api as part of this call,
> and a failure there is logged without failing the registration — so a "registered" user may
> briefly have no wallet rows. Note that both outcomes answer 200: read the `success` field rather
> than the status code.

**Orders — `POST /api/orders`** (unchanged, for contrast)
> **Create an order**
> Creates a limit or market order, determining the maker/taker role automatically.

## Security Checklist

- [x] No secrets/tokens in diff
- [x] Removing `register-with-invitation` closes a path that let a caller set their own `Role`

## Code Quality

- [x] Comments and descriptions in English; Persian quoted verbatim only where it is data — the
      error strings the services actually return
- [x] No claim written from memory: every sentence was read off the code path it describes
- [x] No behaviour changed on any endpoint that remains

## Breaking Changes?

- [x] Breaking changes

Three endpoints are gone. None had a caller anywhere in the repository, and two could never
succeed, so nothing in this system is affected. An external client calling them directly would now
get 404 — but Swagger is Development-only and these APIs are reached solely by the bot's
hand-written typed clients.

## Deployment Notes

No migration, no configuration, no feature flag. Deploy as any other build.
**Rollback**: revert the merge commit; the removed endpoints come back with it.

## Notes for the reviewer

Flagging, not fixing, two things found on the way and left alone as out of scope:

1. `Users.Core` no longer uses any `Affiliate.Core` type — `Invitation` was the last one. The
   `ProjectReference` is left in place, since `STANDARDS.md` records it as load-bearing.
2. `POST /api/user/register` answers `200` for both success and failure (the lambda returns a bare
   `ApiResponse<UserDto>` rather than an `IResult`). Documented as-is rather than corrected.

### DIFF
--- a/src/TallaEgg/TallaEgg.Infrastructure/Clients/IWalletApiClient.cs
+++ b/src/TallaEgg/TallaEgg.Infrastructure/Clients/IWalletApiClient.cs
@@ -39,11 +39,6 @@ public interface IWalletApiClient
     /// Releases locked balance when an order is cancelled.
     /// </summary>
     Task<(bool Success, string Message)> UnlockBalanceAsync(Guid userId, string asset, decimal amount);
-    /// <summary>
-    /// Increase balance for order placement
-    /// Increases a balance.
-    /// </summary>
-    Task<(bool Success, string Message)> IncreaseBalanceAsync(Guid userId, string asset, decimal amount);
 
     /// <summary>
     /// Validate if user has sufficient balance for order
--- a/src/TallaEgg/TallaEgg.Infrastructure/Clients/UsersApiClient.cs
+++ b/src/TallaEgg/TallaEgg.Infrastructure/Clients/UsersApiClient.cs
@@ -110,55 +110,6 @@ public class UsersApiClient : IUsersApiClient
             return ApiResponse<PagedResult<UserDto>>.Fail("خطای غیرمنتظره");
         }
     }
-    public async Task<(bool isValid, string message)> ValidateInvitationCodeAsync(string invitationCode)
-    {
-        try
-        {
-            var request = new { InvitationCode = invitationCode };
-            var json = System.Text.Json.JsonSerializer.Serialize(request);
-            var content = new StringContent(json, Encoding.UTF8, "application/json");
-
-            using var response = await _httpClient.PostAsync($"{_baseUrl}/user/validate-invitation", content);
-            var payload = await response.Content.ReadAsStringAsync();
-
-            if (!response.IsSuccessStatusCode)
-            {
-                _logger.LogWarning("Users API returned {StatusCode} while validating invitation code {InvitationCode}. Payload: {Payload}",
-                    (int)response.StatusCode, invitationCode, payload);
-
-                return (false, $"خطا در اعتبارسنجی کد دعوت: {payload}");
-            }
-
-            var result = System.Text.Json.JsonSerializer.Deserialize<ValidateInvitationResponse>(payload, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
-            if (result is null)
-            {
-                _logger.LogError("Users API returned invalid payload while validating invitation code {InvitationCode}. Payload: {Payload}", invitationCode, payload);
-                return (false, "پاسخ نامعتبر از سرویس کاربران دریافت شد.");
-            }
-
-            return (result.IsValid, result.Message ?? "بررسی کد دعوت انجام شد.");
-        }
-        catch (TaskCanceledException ex)
-        {
-            _logger.LogError(ex, "Users API request timed out while validating invitation code {InvitationCode}", invitationCode);
-            return (false, "پاسخ‌گویی سرویس کاربران زمان‌بر شد");
-        }
-        catch (HttpRequestException ex)
-        {
-            _logger.LogError(ex, "Users API communication error while validating invitation code {InvitationCode}", invitationCode);
-            return (false, "خطای ارتباط با سرویس کاربران");
-        }
-        catch (System.Text.Json.JsonException ex)
-        {
-            _logger.LogError(ex, "Users API returned invalid JSON while validating invitation code {InvitationCode}", invitationCode);
-            return (false, "ساختار پاسخ سرویس کاربران نامعتبر است");
-        }
-        catch (Exception ex)
-        {
-            _logger.LogError(ex, "Unexpected error while validating invitation code {InvitationCode}", invitationCode);
-            return (false, "خطای غیرمنتظره");
-        }
-    }
     public async Task<(bool success, string message, Guid? userId)> RegisterUserAsync(long telegramId, string invitationCode, string? username, string? firstName, string? lastName)
     {
         var request = new RegisterUserRequest
@@ -650,12 +601,6 @@ public class UsersApiClient : IUsersApiClient
         }
     }
 
-
-    private class ValidateInvitationResponse
-    {
-        public bool IsValid { get; set; }
-        public string Message { get; set; } = "";
-    }
 
     private class RegisterUserResponse
     {
--- a/src/TallaEgg/TallaEgg.Infrastructure/Clients/WalletApiClient.cs
+++ b/src/TallaEgg/TallaEgg.Infrastructure/Clients/WalletApiClient.cs
@@ -176,48 +176,6 @@ public class WalletApiClient : IWalletApiClient
         }
     }
 
-    public async Task<(bool Success, string Message)> IncreaseBalanceAsync(
-        Guid userId,
-        string asset,
-        decimal amount)
-    {
-        try
-        {
-            
-            var request = new WalletRequest
-            {
-                UserId = userId,
-                Asset = asset,
-                Amount = amount
-            };
-
-            var json = JsonSerializer.Serialize(request);
-            var stringContent = new StringContent(json, Encoding.UTF8, "application/json");
-
-            // Assuming there's an unlock endpoint - if not, we might need to implement it
-            var response = await _httpClient.PostAsync("api/wallet/increaseBalance", stringContent);
-
-            if (response.IsSuccessStatusCode)
-            {
-                _logger.LogInformation("Successfully increaseBalance {Amount} {Asset} for user {UserId}",
-                    amount, asset, userId);
-                return (true, "موجودی با موفقیت آزاد شد");
-            }
-            else
-            {
-                _logger.LogWarning("Failed to increase balance for user {UserId}, asset {Asset}, amount {Amount}. Status: {Status}",
-                    userId, asset, amount, response.StatusCode);
-                return (false, "خطا در آزاد کردن موجودی");
-            }
-        }
-        catch (Exception ex)
-        {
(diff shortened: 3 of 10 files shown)


MohKardan:
## Review response — 13 findings, 11 accepted, 2 declined

Every finding that asserted a description was false was **reproduced against the running stack**
before I touched anything, rather than settled by reading. Nine held. Fixed in c679040.

### Accepted and fixed

| # | Finding | How it was confirmed |
|---|---|---|
| 1 | `transactions/{userId}` reads the legacy `WalletTransactions` table that nothing writes | Deposited 1,000,000 IRT, then read the endpoint: `count = 0` |
| 2 | `register` — "both outcomes answer 200" is not true of a duplicate Telegram id | Registered the same id twice: first `200`, second `500` |
| 3 | `users/list` — `pageNumber` is unvalidated | `?pageNumber=0` → `500` |
| 4 | `by-role` returns every user's `InvitationCode` | Response fields include `invitationCode` |
| 5 | `unlockBalance` — a negative amount is `500`, not `400` | `Amount = -5` → `500` |
| 6 | `by-role` — a numeric role is not refused | `/by-role/42` → `200 []`; `/by-role/NotARole` → `400` |
| 7 | The id lookups answer an **empty** body, not `null` | `ContentLength = 0` on an unknown phone number |
| 8 | Three retained comments name the wrong exception type | `BusinessRuleException` is `sealed class BusinessRuleException : Exception` — neither an `InvalidOperationException` nor "a plain Exception" |
| 9 | `lockBalance` — "only the amount is validated" is wrong | It validates the asset and creates a missing wallet, which the `balances` description in the same file already said it did |
| 10 | `create-default` states its idempotency twice | Trimmed the comment to the `#206`/`#210` rationale the description cannot carry |
| 12 | `balances` — an empty list is not normal for a registered user | Contradicted `create-default` in the same file; registration seeds three wallets |

Findings 9 and 12 are the ones I'm most glad of: both were **self-contradictions inside a single
file**, exactly the failure mode this issue exists to prevent.

Finding 1 is the most serious. `GET /api/wallet/transactions/{userId}` has no caller and cannot
work — the same shape as the three endpoints this PR deletes. **I have not deleted it**: which
endpoints go is a product decision, and three were named. It is now documented as it behaves, with
`— currently always empty` in the summary so it cannot be missed on the Swagger page. Worth a
follow-up decision.

### Declined

**#11 — `UserService.RegisterUserAsync(User)` is now a test-only seam.** True, and disclosed in the
PR body. The suggested fix retargets eight tests in `DefaultWalletCreationFailureTests` and teaches
`OneUserRepository` to answer `GetUserIdByInvitationCodeAsync`. That is a test refactor this issue
did not ask for, and those tests guard #206/#209/#210 behaviour I would rather not disturb inside a
documentation change. Better as its own issue.

**#13 — `/version` summary and description are duplicated across two services.** Correct, but the
proposed shared constant is an uninvited refactor touching three services, which `STANDARDS.md`
scope discipline rules out here. Per-endpoint metadata living next to its own endpoint is also the
pattern the repository already has. Flagging, not fixing.

### What did not change

No endpoint behaviour. Every defect these descriptions now report is **reported, not repaired** —
the scope of #191 is prose. Six live bugs are named in the descriptions themselves so the next
reader meets them on the Swagger page rather than in production:

- the transactions endpoint returning nothing
- duplicate registration answering 500
- `pageNumber=0` answering 500
- a negative unlock amount answering 500
- `by-role` accepting numeric input
- the empty-body id lookups breaking a non-nullable `Guid` deserialize

```
dotnet build TallaEgg.sln    Build succeeded. 0 Warning(s) 0 Error(s)
dotnet test  TallaEgg.sln    Passed! Failed: 0, Passed: 836, Skipped: 0, Total: 836

WALLET : 10 of 10 operations carry a summary
USERS  : 13 of 13 operations carry a summary
```

Issue #233, opened 2026-09-07
Two JSON serialisers run side by side, and three files use both — the casing tolerance that hides #229 is accidental


MohKardan:
Found while preparing work on #229. Not a defect today; a latent one, and the reason #229 looks safer than it is.

## What is there

Both `Newtonsoft.Json` and `System.Text.Json` are in active use across the solution. Measured on `main`:

| | calls |
|---|---|
| `JsonConvert.*` (Newtonsoft) | 34 |
| `JsonSerializer.*` (System.Text.Json) | 50 |

Newtonsoft arrives through a single `PackageReference` on `TallaEgg.Core` (`TallaEgg.Core.csproj:20`, version 13.0.4). **Eleven projects reference `TallaEgg.Core`**, so it is transitively available almost everywhere, and nothing marks it as legacy.

Files using Newtonsoft:

```
src/Order/Orders.Api/Program.cs
src/Order/Orders.Application/OrderService.cs
src/TallaEgg/TallaEgg.Core/Services/TelegramLoggerService.cs
src/TallaEgg/TallaEgg.Infrastructure/Clients/UsersApiClient.cs
TelegramBot/.../Clients/AffiliateApiClient.cs
TelegramBot/.../Clients/OrderApiClient.cs
TelegramBot/.../TestBotToken.cs
```

## Three files use both

```
src/TallaEgg/TallaEgg.Core/Services/TelegramLoggerService.cs
src/TallaEgg/TallaEgg.Infrastructure/Clients/UsersApiClient.cs
TelegramBot/.../Clients/OrderApiClient.cs
```

`OrderApiClient` is the clearest case — it **serialises with one library and deserialises with the other**:

```csharp
// writes
OrderApiClient.cs:420   System.Text.Json.JsonSerializer.Serialize(order)
OrderApiClient.cs:448   System.Text.Json.JsonSerializer.Serialize(payload)

// reads
OrderApiClient.cs:91    JsonConvert.DeserializeObject<ApiResponse<PagedResult<OrderHistoryDto>>>(payload)
OrderApiClient.cs:141   JsonConvert.DeserializeObject<ApiResponse<PagedResult<TradeHistoryDto>>>(payload)
OrderApiClient.cs:188   JsonConvert.DeserializeObject<ApiResponse<List<OrderHistoryDto>>>(payload)
OrderApiClient.cs:233   JsonConvert.DeserializeObject<ApiResponse<List<OrderHistoryDto>>>(payload)
[...]

PR #236, opened 2026-09-07
[Fix] Give OrderDto one wire name per value — issue #235

### LINKED ISSUE #235: OrderDto's four JSON names are crossed, so POST /api/orders takes two values as four fields
## What happens

`OrderDto` — the request body of `POST /api/orders`, the platform's order-entry endpoint — carries four `[JsonPropertyName]` attributes, and all four are **crossed**: each property serializes under the *other* property's name.

| C# property | wire name | backing |
|---|---|---|
| `Asset` | `"symbol"` | real field |
| `Symbol` | `"asset"` | alias → `Asset` |
| `Quantity` | `"Amount"` | alias → `Amount` |
| `Amount` | `"quantity"` | real field |

[`src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs:40,48,61,67`](../blob/main/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs#L40-L68)

These are the **only four `[JsonPropertyName]` attributes in the entire repository**, and all four are in this one class.

Two consequences, both measured on net9.0 against the real `TallaEgg.Core` type rather than reasoned about.

### 1. Two values are presented as four fields

One order with `Asset = "MAUA/IRT"` and `Amount = 2.5` serializes to:

```json
{"id":"...","symbol":"MAUA/IRT","asset":"MAUA/IRT","Amount":2.5,"quantity":2.5,"price":100,...}
```

`symbol` and `asset` are the same value. `Amount` and `quantity` are the same value. A client reading the Swagger schema sees four independent fields where the server has two.

Note also that `"Amount"` is now **the only PascalCase name left on the wire** anywhere in the platform. `[JsonPropertyName]` overrides the naming policy, so #229 could not and did not touch it — verified: after #229, `JsonSerializer.Serialize(new OrderDto { Amount = 7m }, Web)` still contains `"Amount"`.

### 2. On the way in, the last key wins — silently, and by JSON key order

```
{"symbol":"AAA/IRT","asset":"BBB/IRT"}   ->  Asset = BBB/IRT     ("AAA/IRT" discarded)
{"asset":"BBB/IRT","symbol":"AAA/IRT"}   ->  Asset = AAA/IRT     ("BBB/IRT" discarded)

{"quantity":1,"Amount":999}              ->  Amount = 999        (1 discarded)
{"Amount":999,"quantity":1}              ->  Amount = 1          (999 discarded)
```

**The accepted order quantity depends on the order of keys in the request body.** JSON object key order carries no meaning — clients, proxies and serializers may reorder freely — so this is a silent, order-dependent override on the quantity field of a trading endpoint. No error, no warning, no log line.

## Why nobody has hit it

The only caller is the bot. `BotHandler.cs:1459` constructs `OrderDto` setting **only** the backing properties (`Asset`, `Amount`), and `OrderApiClient.SubmitOrderAsync` serializes the whole object, so both members of each pair always carry identical values and the ambiguity never resolves to anything wrong. Key order is stable because it comes from the property declaration order.

So this is latent, not live. It becomes live the moment a second client writes the JSON itself.

## Why now

#97 (Telegram login + wallet balance view) is a browser client, and a browser client builds its request from the Swagger schema. That schema advertises `symbol`, `asset`, `Amount` and `quantity` as four fields. A developer who sends `symbol` and `asset` — reasonably, since both are documented and `asset` is `[Required]` — gets whichever the serializer emitted last, and the same for the quantity pair.

This is the same "cheap now, breaking later" shape as #229: today it is a DTO nobody else consumes; after a client ships against that schema, changing it is a breaking change.

## Suggested fix

Collapse each pair to one property with one wire name, and delete the aliases:

- `Asset` (or `Symbol` — pick one; `Symbol` matches `BestPricesDto.Symbol` and the rest of the codebase) → `"symbol"`
- `Amount` (or `Quantity`) → `"quantity"`

Then remove all four `[JsonPropertyName]` attributes so the DTO takes the framework camelCase default like everything else. Callers to update: `BotHandler.cs:1459` (construction), `Orders.Api/Program.cs:633,637` (manual validation reads `request.Symbol` / `request.Quantity`), and `OrderService.CreateOrderAsync`.

Worth checking during that change: the `[Required]` / `[Range]` / `[StringLength]` attributes on this DTO are not executed — minimal APIs do not run DataAnnotations, which is why `Program.cs:632-641` validates by hand. The attributes are decorative today, which is adjacent evidence that this class was written against assumptions that do not hold. Deciding whether to delete them or wire up real validation belongs to whoever does this.

---

# Preventing the class, not just this instance

#229 and this issue are the same failure twice: **the JSON wire format is decided in scattered places, by whoever is editing, with no written rule and no reason recorded.** #229 was one service setting a naming policy; this is one DTO overriding four names. Both stayed invisible for the same reason — every internal consumer is tolerant — and both get expensive at the same moment, when a browser client ships.

Four measures, cheapest first.

### 1. Write the rule down — `docs/process/STANDARDS.md`

There is currently **no statement anywhere** of what the platform's JSON wire format is. `STANDARDS.md` covers C# naming, file naming, branch naming and test naming, and says nothing about the shape of an HTTP payload. That gap is the root cause: neither #229 nor this DTO contradicted a rule, because there was no rule.

Proposed addition to §2 (Naming Conventions):

> **JSON wire format.** Every API answers in the ASP.NET Core default camelCase. Do not set
> `PropertyNamingPolicy` in any service. Do not use `[JsonPropertyName]` on a shared DTO unless an
> external contract requires that exact spelling, and record the reason in a comment when you do.
> One value gets one wire name — no alias properties, and never two names for the same field.

### 2. Extend the #229 guard test to cover attributes

`tests/TallaEgg.AllServices.Tests/JsonNamingPolicyConsistencyTests.cs` (added in #234) already sweeps the deployed API project trees plus `TallaEgg.Core` for `PropertyNamingPolicy`, `AddJsonOptions` and `AddNewtonsoftJson`. Adding `JsonPropertyName` to that token list would have caught this DTO, and costs one line.

Since there is a legitimate use for the attribute, pair it with an explicit allowlist so that adding one is a deliberate, reviewed act rather than an unnoticed one:

```csharp
// Wire names pinned against an external contract. Adding to this list is a contract decision:
// it opts a field out of the platform's camelCase default permanently.
private static readonly string[] AllowedPropertyNameOverrides = [ /* empty today */ ];
```

An empty allowlist is the honest starting state — after this issue there should be no `[JsonPropertyName]` in the repo at all.

### 3. A round-trip test for request DTOs

Serialize a populated DTO, deserialize it back, assert equality. That single assertion catches every failure mode above: an alias pair fails it (the discarded value shows up), and so does any future field where the read name and the write name disagree. It is a much stronger guard than a text scan because it tests behaviour, and it belongs next to the existing DTO tests.

### 4. Treat the Swagger schema as the contract, before #97 starts

The schema is what a browser client is built from, and neither #229 nor this issue was visible from the C# — both were only visible on the wire. Before #97 writes its first request, someone should read the generated schema for the endpoints it will call and confirm each field means one thing. That is a one-off review, not a process, and it is cheapest now: #97 has not shipped, so nothing depends on the current shape.

---

**Scope note.** Found while doing #229 (PR #234) and deliberately left out of it: `[JsonPropertyName]` overrides the naming policy in both directions, so these four names are byte-for-byte identical before and after that PR. Fixing them changes the request contract of `POST /api/orders`, which is a different and larger change than #229's one-line deletion — `STANDARDS.md` scope discipline says flag it for the owner rather than fix it inline. Full discussion in [PR #234's review response](../pull/234#issuecomment-5568323572).

MohKardan:
**Branch Name**: `fix/order-dto-crossed-json-names`
**Linked Issue**: closes #235

---

## Description

`OrderDto` — the request body of `POST /api/orders` — carried the repository's only four
`[JsonPropertyName]` attributes, and all four were crossed. Two values reached the wire as four
fields, and on the way in the last key won, so the accepted order quantity depended on the order
of the keys in the request body. This removes the attributes and the two alias properties, and
adds the four preventive measures the issue proposes.

Everything below was measured against the real types on net9.0 and against the three running
services, not reasoned about.

---

## Type of Change

---

## The defect, measured before the change

Confirmed first that these are the only four `[JsonPropertyName]` attributes anywhere:

```
$ grep -rn "JsonPropertyName" --include="*.cs" .        # whole repo, not just src/ TelegramBot/ tests/
src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs:40:        [JsonPropertyName("symbol")]
src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs:48:        [JsonPropertyName("asset")]
src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs:61:        [JsonPropertyName("Amount")]
src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs:67:        [JsonPropertyName("quantity")]
```

### 1. Two values, four fields

One order with `Asset = "MAUA/IRT"` and `Amount = 2.5`:

```
before  {"id":"…","symbol":"MAUA/IRT","asset":"MAUA/IRT","Amount":2.5,"quantity":2.5,"price":100000,…}
after   {"id":"…","asset":"MAUA/IRT","amount":2.5,"price":100000,…}

  asset value  before: 2 fields ["asset", "symbol"]     after: 1 field ["asset"]
  amount value before: 2 fields ["Amount", "quantity"]  after: 1 field ["amount"]
```

### 2. The last key wins, and which one that is has no meaning

```
                                       BEFORE                  AFTER
{"symbol":"AAA/IRT","asset":"BBB/IRT"}  ->  Asset=BBB/IRT       Asset=BBB/IRT
{"asset":"BBB/IRT","symbol":"AAA/IRT"}  ->  Asset=AAA/IRT       Asset=BBB/IRT
{"quantity":1,"Amount":999}             ->  Amount=999          Amount=999
{"Amount":999,"quantity":1}             ->  Amount=1            Amount=999
```

Before, reversing the key order reversed the accepted quantity. After, the removed names are
simply unknown members and the result cannot depend on their position. **The state is now
unreachable**, not merely unused.

### 3. `"Amount"` really was the last PascalCase name on the wire

Not from the C# — from the three published schemas, walked property by property:

```
PascalCase property names, all three services, BEFORE:  OrderDto.Amount
PascalCase property names, all three services, AFTER:   none
```

`[JsonPropertyName]` beats the naming policy in both directions, which is exactly why #229 left
these four bytes untouched and could not have caught them.

---

## Part A — the DTO

Removed all four attributes, so the DTO takes the framework camelCase default like everything
else.

### The aliases go, they do not stay

`Symbol` and `Quantity` were pure aliases (`get => Asset; set => Asset = value;`). **Keeping them
without the attributes would have reproduced the defect exactly**, under camelCase names: `asset`
and `symbol` would both still be emitted, both still bind, and the last key would still win. That
is the defect, not a smaller version of it, so the aliases had to go with the attributes.

`Asset` and `Amount` are the survivors rather than `Symbol` and `Quantity`. The issue offers the
choice and notes `Symbol` matches `BestPricesDto`. Measured across all three schemas, the platform
is genuinely split:

| spelled `asset` / `amount` | spelled `symbol` / `quantity` |
|---|---|
| `OrderDto`, `OrderHistoryDto`, `WalletRequest` | `AcceptQuoteRequest`, `BestPricesDto`, `PublishQuoteRequest`, `TradeDto`, `TradeHistoryDto`, `PositionDto` |

`symbol`/`quantity` is the larger group, but the deciding fact is narrower: **`OrderHistoryDto` is
the response of this very endpoint.** `POST /api/orders` returns
`ApiResponse<CreateOrderResponse>` whose `Order` is an `OrderHistoryDto` with `asset` and `amount`.
A client works one endpoint at a time, so a request and response that disagree costs it more than
a platform-wide majority it can look up once. Keeping `Asset`/`Amount` also matches the `Order`
entity and `CreateOrderCommand`, and leaves `BotHandler.cs:1459` unchanged.

The wider `asset`-versus-`symbol` split is real and predates this issue. It is flagged below, not
fixed here.

### The DataAnnotations go too

`[Required]`, `[Range]` and `[StringLength]` are not executed: minimal APIs do not run
DataAnnotations, and nothing in this repository calls `Validator.TryValidateObject`, registers a
validation filter, or references FluentValidation — checked, not assumed. `Program.cs:632-641`
validates by hand for that reason.

They were not, however, harmless decoration. They reached the **published schema**:

```
OrderDto, BEFORE:  "required": ["Amount","asset","id","side"]
                   "asset":  {"maxLength":20,"minLength":0,…}
                   "Amount": {"minimum":0.00000001,…}
```

So the schema told a client that the two *alias* names were the required ones, and advertised a
length limit and a minimum that the server never applies. That is the same failure as the crossed
names: the schema promising what the server does not do.

I deleted them rather than wiring up real validation:

- Two of the six sat on the alias properties that are gone anyway, and two more (`[Required]` on
  `Guid Id` and on `OrderSide Side`) sit on non-nullable value types where `[Required]` can never
  fail even if it ran.
- Nothing enforced today is lost. The hand-written checks cover the same fields in the same order
  with the same Persian messages, and they still run.
- Wiring up a validation filter would change the 400 body of a live trading endpoint and duplicate
  or replace those checks. It is also a platform-wide decision — all endpoints or none — not a
  property of this DTO.
- **`OrderDto` was the only schema in any of the three services declaring `required`, `maxLength`
  or `minimum` at all.** After this change that count is zero everywhere. Deleting them makes this
  DTO consistent with every other request body; keeping them would have kept it the exception.

Per the issue, the Persian messages on the attributes were not translated — they were deleted with
the attributes they belonged to.

---

## Every consumer of `OrderDto`, found by search rather than taken from the issue's list

```
$ grep -rn "OrderDto" --include="*.cs" .
```

| Site | What it does | Change |
|---|---|---|
| `TelegramBot/…/BotHandler.cs:1459` | builds the DTO, setting only `Asset`, `Amount`, `Price`, `UserId`, `Side`, `Type`, `TradingType` | **none** — it never touched the aliases |
| `TelegramBot/…/Clients/OrderApiClient.cs:418` | `SubmitOrderAsync`, `JsonSerializer.Serialize(order)` with no options | **none** |
| `TelegramBot/…/Clients/IOrderApiClient.cs:15` | the interface | none |
| `src/Order/Orders.Api/Program.cs:633,636` | manual validation, read `request.Symbol` / `request.Quantity` | now `request.Asset` / `request.Amount` |
| `src/Order/Orders.Application/OrderService.cs:66,82,99,101,108,142,143` | seven reads of `request.Symbol` / `request.Quantity` | same rename |
| `tests/…/Fakes/BotTestDoubles.cs:28,54` | records submitted DTOs | none |

**The simulator does not construct an `OrderDto`** — checked, the issue's list was right to be
unsure. It drives the real `BotHandler`, which takes the quote-fill branch
(`AcceptQuoteAsync`) whenever an active quote exists, and the simulator always publishes one. So
`POST /api/orders` is exercised by the bot's live no-quote path and by the direct call below, not
by `driver.ps1 smoke`.

`Program.cs:611` also reads `request.Symbol` / `request.Quantity`, but on `AcceptQuoteRequest`, a
different type. Untouched.

---

## Deployment: the bot and Orders.Api do **not** have to ship together

The issue asks for this to be checked, and asks for a backward-compatible route if one exists.
There is one, and it already holds — measured, not argued.

**An un-updated bot against the new API — over real HTTP, against the running service:**

```
POST /api/orders
{"Id":"00000000-…","symbol":"MAUA/IRT","asset":"MAUA/IRT","Amount":0.25,"quantity":0.25,"Price":22900000,…}

-> 200 {"success":true,"data":{"order":{"asset":"MAUA/IRT","amount":0.25,"price":22900000,…}}}
```

The old bot sends *both* spellings of each value, so the new DTO finds `asset` for `Asset` and
`Amount` for `Amount`; `symbol` and `quantity` are ignored as unknown members. **An old bot keeps
working against a new API.**

The reverse direction cannot be posted to a service that no longer exists, so it was measured
in-process against both DTO shapes:

```
old bot payload -> NEW OrderDto : Asset=MAUA/IRT  Amount=2.5   [OK]
new bot payload -> OLD OrderDto : Asset=MAUA/IRT  Amount=2.5   [OK]
```

`{"Asset":…}` matches the old DTO's `"asset"` name case-insensitively, and `{"Amount":…}` matches
its `"Amount"` exactly. **Both directions work.**

**What that compatibility rests on, and how it fails.** Entirely on
`PropertyNameCaseInsensitive = true` in `Orders.Api` — the line #234 deliberately kept. With it
off, both directions lose the asset:

```
strict reading:  old -> NEW : Asset=""  [BROKEN]     new -> OLD : Asset=""  [BROKEN]
```

The failure is loud rather than silent. A body carrying only the removed names is refused
outright:

```
POST {"symbol":"MAUA/IRT","quantity":0.40,…}
-> 400 {"success":false,"message":"نماد معاملاتی الزامی است"}
```

An empty asset cannot reach the matching engine — the hand-written check rejects it first. So the
worst case of a mixed deployment is a rejected order, never a trade on the wrong quantity.

**Recommendation:** deploy them together anyway. The compatibility is real but it depends on one
setting in one file staying as it is, and there is no reason to spend it. If they do go
separately, either order is safe.

---

## Part B — preventing the class

All four measures from the issue are done.

### 1. The rule is written down — `docs/process/STANDARDS.md` §2

A new **JSON Wire Format** subsection carrying the issue's proposed text verbatim, plus why it
exists: neither #229 nor #235 contradicted a rule, because there was none, and both get expensive
at the same moment.

### 2. The #229 guard now covers `[JsonPropertyName]`, behind an allowlist

`JsonNamingPolicyConsistencyTests` gains
`WireNameOverrides_OnAnySerializedType_AreAbsentOrAllowlisted`, paired with

```csharp
// Wire names pinned against an external contract. Adding to this list is a contract decision:
// it opts a field out of the platform's camelCase default permanently.
private static readonly string[] AllowedPropertyNameOverrides = [];
```

Two deliberate departures from the issue's "one line":

- **A separate test, not another token in `NamingPolicyOverrides`.** The existing failure message
  is about a *host* overriding its response naming policy; an attribute is a different act with a
  different fix, and the allowlist only applies to one of them.
- **A wider root list.** The naming policy is wired where a host is configured, so the three APIs
  plus `TallaEgg.Core` is right for it. An attribute travels with the *type*, so it matters
  wherever a serialized type is declared — and the bot declares wire types of its own
  (`NotifyMatchingEngineRequest`, `InvitationDto`, `CancelActiveOrdersResponseDto`). Both bot
  project trees are swept as well.

The shared file walk was factored into one `ScanForTokens` helper so the second guard is not a
copy of the first.

### 3. Behavioural round-trip guards — `tests/…/RequestDtoWireContractTests.cs`

Two theories over **all twelve request DTOs declared in `TallaEgg.Core`** — every type bound as a
request body by a `MapPost`/`MapPut` in Orders, Users or Wallet.
(`AbandonOutboxMessageRequest` and `UpdateUserRoleRequest` are declared inside the API projects'
top-level statements, which the test project does not reference; noted in the file.)

- `RequestDto_SerializedAfterBeingRead_ReturnsEveryFieldUnchanged` — fill every field with a value
  of its own, read the body, write it back, require every field to still hold what was sent.
- `RequestDtoWireName_EachOne_GovernsExactlyOneValue` — send one field at a time and watch which
  C# properties move. Two names that disturb the same properties are one field with a spare
  spelling; a name that disturbs nothing is written but cannot be read back.

Both are driven off `JsonTypeInfo`, the serializer's own contract, so they do not assume the naming
convention they are guarding.

**One correction to the issue.** It says a round-trip test catches an alias pair because "the
discarded value shows up". The naive form does not: both aliases always carry the same value after
population, so a populated `OrderDto` round-trips cleanly even before this fix. What makes it
catch the defect is giving **each wire name its own distinct value** — then `symbol` and `asset`
collide and one comes back changed. That is why the tests are written this way, and why the
second theory exists as well: it names the cause rather than the symptom.

**Verified they are not vacuous** by restoring the pre-fix DTO and re-running:

```
WireNameOverrides_OnAnySerializedType_AreAbsentOrAllowlisted("src/TallaEgg/TallaEgg.Core") [FAIL]
  OrderDtos.cs:56: [JsonPropertyName("symbol")]      OrderDtos.cs:58: [JsonPropertyName("asset")]
  OrderDtos.cs:64: [JsonPropertyName("Amount")]      OrderDtos.cs:70: [JsonPropertyName("quantity")]

RequestDto_SerializedAfterBeingRead_ReturnsEveryFieldUnchanged(OrderDto) [FAIL]
  Sent: {…,"symbol":"value-1","asset":"value-2","Amount":4,"quantity":5,…}
  Back: {…,"symbol":"value-2","asset":"value-2","Amount":5,"quantity":5,…}

RequestDtoWireName_EachOne_GovernsExactlyOneValue(OrderDto) [FAIL]
  "symbol" and "asset" both set Asset+Symbol
  "Amount" and "quantity" both set Amount+Quantity
```

### 4. Swagger schema review, before #97 starts

Read the generated schema of all three services, walked property by property, before and after.

**Fixed by this PR**

- `OrderDto` advertised four fields for two values; now two. No schema in any service now carries
  two names for one value.
- `OrderDto.Amount` was the only PascalCase name on the wire platform-wide; now there are none.
- `OrderDto` was the only schema declaring `required`, `maxLength` or `minimum`. All three counts
  are now zero everywhere, so no schema promises validation the server does not perform.

**Found, not fixed — worth an issue before #97 writes its first request**

1. **The same concept has two names across endpoints.** A trading pair is `asset` in `OrderDto`,
   `OrderHistoryDto` and `WalletRequest`, and `symbol` in `AcceptQuoteRequest`, `BestPricesDto`,
   `PublishQuoteRequest`, `TradeDto`, `TradeHistoryDto` and `PositionDto`; quantity is `amount` or
   `quantity` along the same fault line. Both order-entry paths a browser client would use are on
   opposite sides of it: `POST /api/orders` takes `asset`/`amount`, `POST /api/quotes/accept`
   takes `symbol`/`quantity`. Each endpoint is now internally consistent with its own response,
   which is what this PR could fix without breaking something else; unifying the vocabulary is a
   larger contract change and belongs to whoever owns #97.
2. **Every schema carries `additionalProperties: false`, which the server does not enforce.** A
   strict client-side validator would reject a body the server accepts happily — the old bot's
   payload above is exactly such a body. Harmless today, confusing to whoever generates a typed
   client from this schema.
3. **`OrderDto.Id` is a request field nothing reads.** The order id is generated server-side, and
   its doc comment says "User id." — next to a separate `UserId` property. A client following the
   schema could reasonably send the user id as `id`. Left alone here per scope discipline;
   deleting the field is a further contract change.

Happy to open issues for these three; say the word and I will.

---

### Build

```
```

### Unit tests

```
dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 871, Skipped: 0, Total: 871
```

841 on `main` plus 30 new: 24 from `RequestDtoWireContractTests` (12 DTOs x 2 theories) and 6 from
the new `[JsonPropertyName]` sweep.

`scripts/check-doc-paths.sh` — `397 tracked text file(s) checked, every reference resolves.`

### End-to-end

```
=== Done in 00:00:29.84. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:
```

Settlement counts are the delta the driver asserts, not absolutes.

### A real order through the bot's own client, checked in the database

`driver.ps1 smoke` covers the quote-fill path, which does not go through `OrderDto` — so the
changed path was driven separately, through the bot's real `OrderApiClient.SubmitOrderAsync` (the
instance `BotHandler` holds) with the `OrderDto` built from exactly the seven properties
`BotHandler.cs:1459` sets, against the running Orders.Api and the real database:

```
Request body the bot puts on the wire:
  {"Id":"00000000-…","Asset":"MAUA/IRT","Amount":0.50,"Price":23000000,"UserId":"…","Side":0,"Type":1,…}

```

Read straight out of `TallaEggOrders.dbo.Orders`:

```
Asset='MAUA/IRT'  Amount=0.50000000  Price=23000000.00000000  Status=Confirmed
Asset='MAUA/IRT'  Amount=0.25000000  Price=22900000.00000000  Status=Confirmed   <- old-bot payload
Asset='MAUA/IRT'  Amount=0.30000000  Price=22800000.00000000  Status=Confirmed   <- browser payload
```

`Asset` and `Amount` land correctly from all three payload shapes. The three probe orders were
cancelled afterwards so nothing rests in the book. Prices are the live MAUA/IRT quote
(bid 23,215,848.38 / ask 23,262,326.52 at the time), not invented figures.

What this does **not** cover: the Telegram conversation above `SubmitOrderAsync`. That code is
untouched by this PR, and the smoke run drives it on the quote path.

### Environment

---

## Security Checklist

- [x] `.gitignore` blocks sensitive files — `config/appsettings.global.json` is untouched and untracked
- [x] Input validation present — the hand-written checks in `Program.cs` are unchanged in effect,
      and now read the surviving property names

---

## Code Quality

- [x] Follows `STANDARDS.md` naming, including the section this PR adds

---

## Documentation

- [x] `docs/process/STANDARDS.md` §2 gains the JSON wire format rule
- [x] Swagger/OpenAPI: the schema changes with the DTO; reviewed above
- [x] `OrderDto` carries a `<remarks>` block recording what was wrong and why each decision went
      the way it did — recording no reason is what made #229 unanswerable from the code

---

## Breaking Changes?

- [ ] No breaking changes
- [x] Breaking changes (see details below)

**Formally breaking**: the request contract of `POST /api/orders` loses the `symbol` and
`quantity` fields and renames `Amount` to `amount`.

**In practice**: no consumer breaks. The only caller is the bot, which sets the backing properties
and whose payload the new API still accepts unchanged (measured above). A body using *only* the
removed names is refused with a 400 and a Persian message, not silently mis-read.

Per `STANDARDS.md` §11 this is a MINOR-shaped change, not MAJOR: no external integration exists to
break. No version bump is included — it is not this PR's call to make.

---

## Deployment Notes

**Order**: the bot and `Orders.Api` may be deployed in either order or together; together is
recommended. See the Deployment section above for the measurements.

**Post-deployment verification**
- [ ] `GET /swagger/v1/swagger.json` on Orders shows `OrderDto` with `asset` and `amount` and no
      `symbol`/`quantity`
- [ ] One order placed from the bot appears in `Orders` with the right `Asset` and `Amount`

**Rollback**: revert the commit and redeploy. An old `Orders.Api` reads the new bot's payload
correctly, so a rollback of the API alone is also safe.

---

## Related Issues

- Closes #235
- Follows #229 / PR #234, where this was found and deliberately left out
- Relevant to #97 (browser client) — the Swagger review above was done for it
- Untouched, as the issue requires: #233 (two serialization libraries in one file) and #230
  (migration timing)

---

## Final Checklist

- [x] All tests pass locally
- [x] Commit message is clear

### DIFF
--- a/docs/process/STANDARDS.md
+++ b/docs/process/STANDARDS.md
@@ -80,9 +80,10 @@ placement at repo root is existing history, not a pattern to copy.
 #### JSON Wire Format
 
 **Every API answers in the ASP.NET Core default camelCase.** Do not set `PropertyNamingPolicy` in
-any service. Do not use `[JsonPropertyName]` on a shared DTO unless an external contract requires
-that exact spelling, and record the reason in a comment when you do. **One value gets one wire
-name** — no alias properties, and never two names for the same field.
+any service. Do not use `[JsonPropertyName]` — or Newtonsoft's `[JsonProperty]`, which this
+repository is equally able to reach for — on a shared DTO unless an external contract requires that
+exact spelling, and record the reason in a comment when you do. **One value gets one wire name** —
+no alias properties, and never two names for the same field.
 
 This paragraph exists because its absence was the root cause of two defects. Neither #229 (one
 service serving PascalCase because it set the naming policy to `null`) nor #235 (`OrderDto`
--- a/docs/process/STANDARDS.md
+++ b/docs/process/STANDARDS.md
@@ -77,6 +77,25 @@ placement at repo root is existing history, not a pattern to copy.
 - **Test files**: `{ClassName}Tests.cs` (e.g., `WalletServiceTests.cs`)
 - **Project folders**: PascalCase matching project name
 
+#### JSON Wire Format
+
+**Every API answers in the ASP.NET Core default camelCase.** Do not set `PropertyNamingPolicy` in
+any service. Do not use `[JsonPropertyName]` on a shared DTO unless an external contract requires
+that exact spelling, and record the reason in a comment when you do. **One value gets one wire
+name** — no alias properties, and never two names for the same field.
+
+This paragraph exists because its absence was the root cause of two defects. Neither #229 (one
+service serving PascalCase because it set the naming policy to `null`) nor #235 (`OrderDto`
+presenting two values as four crossed fields, so the accepted quantity depended on the order of
+the keys in the request body) contradicted a rule — there was no rule to contradict, and both
+stayed invisible because every internal consumer reads case-insensitively. They get expensive at
+the same moment: when a browser client ships against the published schema.
+
+`tests/TallaEgg.AllServices.Tests/JsonNamingPolicyConsistencyTests.cs` enforces the first two
+sentences and holds the allowlist for the third; the two behavioral guards in
+`RequestDtoWireContractTests.cs` enforce the last. Adding a name to that allowlist is a contract
+decision, not a formatting one.
+
 #### Branch Names (Git)
 - **Feature**: `feat/{description}` (e.g., `feat/add-wallet-transaction`)
 - **Bugfix**: `fix/{description}` (e.g., `fix/null-reference-wallet`)
--- a/src/Order/Orders.Api/Program.cs
+++ b/src/Order/Orders.Api/Program.cs
@@ -629,11 +629,12 @@ app.MapPost("/api/orders", async (TallaEgg.Core.DTOs.Order.OrderDto request, Ord
 {
     try
     {
-        // Validate request
-        if (string.IsNullOrWhiteSpace(request.Symbol))
+        // Validate request. Minimal APIs do not run DataAnnotations, so these hand-written checks
+        // are the only validation this endpoint has — see the remarks on OrderDto (issue #235).
+        if (string.IsNullOrWhiteSpace(request.Asset))
             return Results.BadRequest(new { success = false, message = "نماد معاملاتی الزامی است" });
         
-        if (request.Quantity <= 0)
+        if (request.Amount <= 0)
             return Results.BadRequest(new { success = false, message = "مقدار سفارش باید بیشتر از صفر باشد" });
         
         if (request.Price <= 0)
--- a/src/Order/Orders.Application/OrderService.cs
+++ b/src/Order/Orders.Application/OrderService.cs
@@ -63,7 +63,7 @@ public class OrderService
         try
         {
             _logger.LogInformation("Creating unified order for user {UserId} with symbol {Symbol}, side {Side}, type {Type}",
-                request.UserId, request.Symbol, request.Side, request.Type);
+                request.UserId, request.Asset, request.Side, request.Type);
 
             // 1. Validate authorization
             var canCreateOrder = true;
@@ -79,7 +79,7 @@ public class OrderService
             // 3. Validate user balance before creating order
             var userId = request.UserId;
             var assetToCheck = request.Side == TallaEgg.Core.Enums.Order.OrderSide.Buy
-                ? request.Symbol.Split('/')[1] : request.Symbol.Split('/')[0];
+                ? request.Asset.Split('/')[1] : request.Asset.Split('/')[0];
 
             // The price is rounded to the column's precision before anything else uses it.
             //
@@ -96,16 +96,16 @@ public class OrderService
 
             request.Price = CurrenciesConstant.RoundOrderPrice(request.Price);
 
-            ValidateTradingLimits(request.Symbol, request.Quantity, request.Price);
+            ValidateTradingLimits(request.Asset, request.Amount, request.Price);
 
-            var (_, amountToCheck) = ComputeCollateral(request.Symbol, orderSide, request.Quantity, request.Price);
+            var (_, amountToCheck) = ComputeCollateral(request.Asset, orderSide, request.Amount, request.Price);
 
             _logger.LogInformation("Validating balance for user {UserId}: {Amount} {Asset}",
                 userId, amountToCheck, assetToCheck);
 
             
             var validateCreditAndBalance =
-                await _walletApiClient.ValidateCreditAndBalanceAsync(request.UserId, request.Symbol, request.Quantity, request.Price);
+                await _walletApiClient.ValidateCreditAndBalanceAsync(request.UserId, request.Asset, request.Amount, request.Price);
 
             var hasSufficientBalance = request.Side == OrderSide.Buy
                 ? validateCreditAndBalance.HasSufficientCreditAndBalanceQuote : validateCreditAndBalance.HasSufficientCreditAndBalanceBase;
@@ -139,8 +139,8 @@ public class OrderService
 
             // Limit orders start as Makers
             var limitCommand = new CreateOrderCommand(
-                request.Symbol,
-                request.Quantity,
+                request.Asset,
+                request.Amount,
                 request.Price,
                 userId,
                 orderSide,
--- a/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs
+++ b/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs
@@ -1,9 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Text;
-using System.Text.Json.Serialization;
 using System.Threading.Tasks;
 using TallaEgg.Core.Enums.Order;
 
@@ -29,50 +27,47 @@ namespace TallaEgg.Core.DTOs.Order
     /// Unified order creation request for all order types
     /// A single order-creation request covering every order type.
     /// </summary>
+    /// <remarks>
+    /// One value, one wire name (issue #235). This class used to carry the repository's only four
+    /// <c>[JsonPropertyName]</c> attributes and all four were crossed: <c>Asset</c> serialized as
+    /// <c>"symbol"</c> while an alias called <c>Symbol</c> serialized as <c>"asset"</c>, and the
+    /// same for <c>Amount</c>/<c>Quantity</c>. Two values reached the wire as four fields, and on
+    /// the way in the last key won, so the accepted quantity depended on the order of the keys in
+    /// the request body.
+    ///
+    /// The aliases went with the attributes rather than surviving them: kept as plain properties
+    /// they would have produced the same four fields again under camelCase names, which is the
+    /// defect, not a smaller version of it. <c>Asset</c> and <c>Amount</c> are the survivors
+    /// because this endpoint's own response uses those names (<see cref="OrderHistoryDto"/>), so
+    /// request and response now agree.
+    ///
+    /// The DataAnnotations went too. Minimal APIs do not execute them — nothing in this repository
+    /// calls <c>Validator.TryValidateObject</c> or registers a validation filter — so they enforced
+    /// nothing while making the generated schema advertise constraints the server does not apply.
+    /// Orders.Api validates this request by hand instead.
+    /// </remarks>
     public class OrderDto
     {
         /// <summary>
         /// User id.
         /// </summary>
-        [Required(ErrorMessage = "شناسه کاربر الزامی است")]
         public Guid Id { get; set; }
 
-        [JsonPropertyName("symbol")]
+        /// <summary>
+        /// Asset symbol, as a trading pair — for example <c>MAUA/IRT</c>.
+        /// </summary>
         public string Asset { get; set; } = "";
         /// <summary>
-        /// Asset symbol.
-        /// alias
+        /// Order quantity, in the base asset.
         /// </summary>
-        [Required(ErrorMessage = "نماد دارایی الزامی است")]
-        [StringLength(20, ErrorMessage = "نماد دارایی نمی‌تواند بیش از 20 کاراکتر باشد")]
-        [JsonPropertyName("asset")]
-        public string Symbol
-        {
-            get => Asset;
-            set => Asset = value;
-        }
-
-        /// <summary>
-        /// Order quantity.
-        /// alias
-        /// </summary>
-        [Required(ErrorMessage = "مقدار سفارش الزامی است")]
-        [Range(0.00000001, double.MaxValue, ErrorMessage = "مقدار سفارش باید بزرگتر از صفر باشد")]
-        [JsonPropertyName("Amount")]
-        public decimal Quantity
-        {
-            get => Amount;
-            set => Amount = value;
-        }
-        [JsonPropertyName("quantity")]
         public decimal Amount { get; set; }
+
         /// <summary>
         /// Price. Required for limit orders, optional for market orders.
         /// </summary>
         public decimal Price { get; set; }
         public Guid UserId { get; set; }
 
-        [Required(ErrorMessage = "سمت سفارش یا جهت سفارش (خرید یا فروش(")]
         public OrderSide Side { get; set; }
         public OrderType Type { get; set; }
         public OrderStatus Status { get; set; }
--- a/tests/TallaEgg.AllServices.Tests/JsonNamingPolicyConsistencyTests.cs
+++ b/tests/TallaEgg.AllServices.Tests/JsonNamingPolicyConsistencyTests.cs
@@ -33,21 +33,23 @@ public class JsonNamingPolicyConsistencyTests
     ];
 
     /// <summary>
-    /// Where a <c>[JsonPropertyName]</c> can reach the wire. That is a wider question than where
-    /// the naming policy is wired: an attribute travels with the type, so it matters wherever a
-    /// serialized type is <em>declared</em>, not only where a host is configured. The bot declares
-    /// request and response types of its own (<c>NotifyMatchingEngineRequest</c>,
-    /// <c>InvitationDto</c>), so its two project trees are swept alongside the APIs and the shared
-    /// kernel that holds most DTOs.
+    /// Where a pinned wire name can reach the wire, which is every line of C# the product ships.
     /// </summary>
+    /// <remarks>
+    /// Deliberately two whole trees rather than a list of projects. Where the naming policy is
+    /// wired is a question about hosts, so <see cref="JsonWiringRoots"/> names the three that serve
+    /// HTTP; an attribute instead travels with the type, so it matters wherever a serialized type
+    /// is <em>declared</em> — and they are declared all over. <c>TallaEgg.Infrastructure/Clients</c>
+    /// declares <c>RegisterUserResponse</c> and <c>UpdateRoleResponse</c>, the bot declares
+    /// <c>NotifyMatchingEngineRequest</c> and <c>InvitationDto</c>, and the Application layers
+    /// declare more. Enumerating the projects that hold one today would leave this guard passing
+    /// vacuously the first time somebody declares a DTO somewhere new, which is the failure mode a
+    /// guard like this is most prone to.
+    /// </remarks>
     public static TheoryData<string> SerializedTypeRoots =>
     [
-        "src/User/Users.Api",
-        "src/Wallet/Wallet.Api",
-        "src/Order/Orders.Api",
-        "src/TallaEgg/TallaEgg.Core",
-        "TelegramBot/TallaEgg.TelegramBot.Core",
-        "TelegramBot/TallaEgg.TelegramBot.Infrastructure",
+        "src",
+        "TelegramBot",
     ];
 
     /// <summary>
@@ -61,6 +63,22 @@ public class JsonNamingPolicyConsistencyTests
     /// until one exists this list has nothing to hold.
     /// </remarks>
     private static readonly string[] AllowedPropertyNameOverrides = [];
+
+    /// <summary>
+    /// Both spellings of "pin this field's wire name", one token: <c>JsonProperty</c> is a prefix
+    /// of <c>JsonPropertyName</c>, so it matches System.Text.Json's attribute and Newtonsoft's
+    /// alike.
+    /// </summary>
+    /// <remarks>
+    /// Newtonsoft is not hypothetical here. <c>TallaEgg.Core</c> references it, and
+    /// <c>UsersApiClient</c> and <c>AffiliateApiClient</c> serialize real request bodies with
+    /// <c>JsonConvert.SerializeObject</c> — so a <c>[JsonProperty("…")]</c> would reproduce #235
+    /// exactly, in a library the System.Text.Json guards cannot see.
+    /// </remarks>
+    private static readonly string[] WireNamePins =
+    [
+        "JsonProperty",
+    ];
 
     /// <summary>
     /// Every way an ASP.NET Core host can move its responses off the default camelCase.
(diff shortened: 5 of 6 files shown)


MohKardan:
## Review response — four findings, all four accepted

`/code-review high 236`. Every finding landed on the preventive guards rather than on the
production change, which is the right place for them to land: the guards are the part of this PR
whose only job is to stop #235 happening again, so a gap there is a gap in the thing being bought.
All four are fixed in 26c7e95, and two of them were verified by planting the exact evasion each
one describes.

The review also independently reproduced the PR's claims — build clean, 871 tests green, doc-path
check green, and it confirmed by its own search that `OrderDto` is request-only, that the three
consumers are the three named, and that the outbox stores `TradeDto` rather than `OrderDto` so no
persisted payload replays through the renamed properties. That last one I had not checked, and it
is the kind of thing a rename like this dies on.

---

### 1. Both guards were System.Text.Json only, and Newtonsoft is in reach — **accepted**

The sharpest of the four. `TallaEgg.Core` references Newtonsoft, and `UsersApiClient` and
`AffiliateApiClient` serialize **real request bodies** with `JsonConvert.SerializeObject`
(`UsersApiClient.cs:124,271,324,386`, `AffiliateApiClient.cs:33,86`). A `[JsonProperty("…")]`
would reproduce #235 exactly, in a library that neither a `JsonPropertyName` text scan nor a
`System.Text.Json` round trip can see. #234's review response had already noted that
`OrderApiClient` alone uses two serialization libraries; I did not carry the consequence over to
this guard.

The fix is one token, because `JsonProperty` is a prefix of `JsonPropertyName` and so matches both
attributes at once. `STANDARDS.md` names both as well — the written rule had the same omission,
and a rule that only forbids one of two spellings is worse than no rule, because it reads as a
considered choice.

Confirmed there are no false positives: zero `JsonProperty` occurrences on a code line anywhere in
`src/` or `TelegramBot/` today. `///` doc comments that name the attribute are skipped by `IsCode`.

### 2. `SerializedTypeRoots` enumerated projects and missed ones — **accepted**

Also correct, and correct for a reason worth naming: I had enumerated the projects that hold a DTO
*today*. `src/TallaEgg/TallaEgg.Infrastructure/Clients` holds `RegisterUserResponse` and
`UpdateRoleResponse` and was not on the list. This is exactly the vacuous-pass failure that #234's
review finding 3 raised about the first version of the naming-policy guard, arriving a second time
in a different disguise.

It now sweeps `src` and `TelegramBot` whole. Two theory cases instead of six, and nowhere left to
hide.

Verified in the direction that matters, by planting the review's own scenario — Newtonsoft's
attribute, in the tree the old list omitted:

```
WireNameOverrides_OnAnySerializedType_AreAbsentOrAllowlisted(relativeRoot: "src") [FAIL]
  src\TallaEgg\TallaEgg.Infrastructure\Clients\TempEvasionProbe.cs:7: [Newtonsoft.Json.JsonProperty("Symbol")]
```

The previous version passed on that file. Both findings 1 and 2 are covered by that one probe.

### 3. `Equals(object, object)` is reference equality — **accepted**

Right, and the consequence is worse than a wrong answer: a DTO with an inline-initialised
collection would read as "moved" on **every** probe, so `moved` would never be empty and the
`inert` list would never fill. The "writes fields it cannot read back" assertion would retire
itself and keep reporting green. A guard that goes quiet without saying so is the failure mode
this whole PR is about.

Property values are now compared as serialized JSON. No DTO in the list has such a property today,
which is precisely why it would have gone unnoticed until one did.

The related note — that members with no sentinel are never probed — is by design, and the remarks
now say so where a reader will meet it rather than only at `SentinelJson`.

### 4. The allowlist matched a bare quoted name anywhere on the line — **accepted**

Fair. It now matches the attribute's own argument, so allowlisting `"symbol"` cannot exempt some
other line that merely mentions it.

The deeper half of the finding — that allowlisting a name still cannot tell you whether it
collides with a value another property already carries — is not solvable by a text scan at all,
and I have stopped pretending the failure message covers it by asking the reviewer to check by
hand. That collision *is* #235, and `RequestDtoWireName_EachOne_GovernsExactlyOneValue` decides it
by behaviour. The remarks now say which guard answers which question, so the pair reads as two
halves of one rule rather than one guard with a caveat.

---

### Re-verified after all four changes

```
dotnet build TallaEgg.sln --no-incremental   0 Warning(s)  0 Error(s)
dotnet test TallaEgg.sln                     Passed! - Failed: 0, Passed: 867, Total: 867
scripts/check-doc-paths.sh                   397 tracked text file(s) checked, every reference resolves.
```

867 rather than 871 only because the attribute theory dropped from six roots to two — the coverage
went up, the case count went down.

And all three guards still fail on the pre-fix DTO, which is the property that matters:

```
WireNameOverrides_OnAnySerializedType_AreAbsentOrAllowlisted("src")   [FAIL]  all four attributes named
RequestDto_SerializedAfterBeingRead_ReturnsEveryFieldUnchanged(OrderDto)   [FAIL]
RequestDtoWireName_EachOne_GovernsExactlyOneValue(OrderDto)                [FAIL]
  "symbol" and "asset" both set Asset+Symbol
  "Amount" and "quantity" both set Amount+Quantity
```

No production code changed in 26c7e95, so the end-to-end evidence in the PR body still stands on
the build it was taken from; the stack was re-run against this commit regardless.

MohKardan:
The Swagger review this issue asked for as its fourth preventive measure turned up three things
that could not be fixed inside PR #236 without breaking something else. They are now tracked
together in #237:

1. The platform's split `asset`/`symbol` vocabulary — the two order-entry endpoints sit on opposite
   sides of it (`POST /api/orders` takes `asset`/`amount`, `POST /api/quotes/accept` takes
   `symbol`/`quantity`).
2. `additionalProperties: false` on all 31 schemas this project declares, while no service enforces
   it — demonstrated by a 200 on a body carrying two fields the schema does not declare.
3. `OrderDto.Id` — a request field nothing reads, documented as "User id." four lines above a
   separate `UserId` property.

Each needs a decision rather than a correction, and all three are cheapest to settle before #97
writes its first request.

PR #239, opened 2026-09-07
[Fix] Make the published schema describe what the servers do — issue #237

### LINKED ISSUE #237: The Swagger schema misleads a client three ways: split vocabulary, unenforced additionalProperties, and a dead id field
Three things the Swagger schema review in #235 turned up. None is a defect today — every one of
them is a way the published schema misleads whoever builds a client from it, which is #97. They
are filed together because they were found together, in one pass over the generated schemas of all
three services, and because they share a cause: **nobody has ever read this schema as a contract**,
because until #97 nothing consumed it.

Measured on 2026-09-07 against `main` at `1563b7d`, by walking
`GET /swagger/v1/swagger.json` on Users (5136), Wallet (60933) and Orders (5140) property by
property — not read off the C#.

#235 fixed everything in that review that could be fixed without breaking something else: the
crossed names, the last PascalCase name on the wire, and the only schema advertising
DataAnnotations the server does not run. These three are what was left, and each needs a decision
rather than a correction.

---

## 1. The same concept has two names, depending on the endpoint

A trading pair is `asset` in three schemas and `symbol` in six. Its quantity is `amount` or
`quantity` along exactly the same fault line.

| | `asset` / `amount` | `symbol` / `quantity` |
|---|---|---|
| Orders | `OrderDto`, `OrderHistoryDto` | `AcceptQuoteRequest`, `BestPricesDto`, `PublishQuoteRequest`, `TradeDto`, `TradeHistoryDto`, `PositionDto` |
| Wallet | `WalletRequest` | `TradeDto` |

**The two order-entry paths are on opposite sides of it**, which is what makes this worth an issue
rather than a note:

```
POST /api/orders          {"asset": "MAUA/IRT", "amount": 2.5,   "price": …}
POST /api/quotes/accept   {"symbol": "MAUA/IRT", "quantity": 2.5}
```

A browser client that has a "place an order" function has to know both vocabularies and pick per
endpoint. Same for reading results back: an order's history is `asset`/`amount`, a trade's is
`symbol`/`quantity`.

### Why #235 did not fix this

It could only have made things worse from where it stood. `OrderDto` had to pick one of its two
crossed pairs, and the deciding fact was narrow: **`OrderHistoryDto` is the response of that very
endpoint.** `POST /api/orders` returns `ApiResponse<CreateOrderResponse>` whose `Order` is an
`OrderHistoryDto` with `asset` and `amount`. Renaming the request to `symbol`/`quantity` would have
joined the larger group at the cost of making one endpoint's own request and response disagree,
which is worse for a client that works an endpoint at a time.

So every endpoint is now internally consistent, and the split between endpoints is untouched and
pre-existing.

### What the decision is

Not "rename things until they match" — that is a breaking change to six schemas including
`TradeDto`, which is also **the outbox payload**, serialized to the database with a bare
`JsonSerializer` and read back by `OutboxProcessorService`. A rename there is a data migration, not
a rename. (Migration timing is #230's territory.)

Realistic options, cheapest first:

1. **Write the split down and leave it.** Document in `docs/process/STANDARDS.md` §2, next to the
   JSON wire format rule #235 added, that the platform has two vocabularies, which endpoints use
   which, and that new endpoints match their neighbours. Costs nothing, breaks nothing, and stops
   the next person deciding at random — which is exactly the gap that produced #229 and #235.
2. **Unify one direction on new endpoints only.** Pick `symbol`/`quantity` (the majority) as the
   spelling for anything added from now on, and let the existing split age out.
3. **Unify everything.** Breaking, touches the outbox payload, and needs #97 to be either unstarted
   or coordinated. Only worth it if it happens before a browser client ships.

My recommendation is (1) now and (2) as the standing rule, with (3) reconsidered only if #97 finds
the split genuinely painful. **(1) should happen before #97 writes its first request**, which is
the whole reason this is filed now.

---

## 2. Every schema says `additionalProperties: false`, and no service enforces it

31 of the 32 object schemas across the three services carry it. The one exception is
`HttpValidationProblemDetails`, which is the framework's own and deliberately extensible — so
**every schema this project declares** says it.

The servers do not behave that way. `System.Text.Json` ignores unknown members by default, and
nothing sets `JsonUnmappedMemberHandling.Disallow`. Demonstrated over real HTTP against the running
Orders.Api — a body carrying two fields the schema does not declare at all:

```
POST /api/orders
{"Id":"00000000-…","symbol":"MAUA/IRT","asset":"MAUA/IRT","Amount":0.25,"quantity":0.25,"Price":22900000,…}

-> 200  order created, asset MAUA/IRT, amount 0.25
```

`symbol` and `quantity` no longer exist in `OrderDto`. The schema says that body is invalid; the
server accepts it happily.

That leniency is load-bearing, incidentally — it is what lets an un-updated bot keep talking to a
post-#235 Orders.Api (see PR #236's deployment section). So the schema is wrong, not the server.

**Why it matters for #97:** a generated or validating client — most OpenAPI codegen, and anything
running a request through a JSON-schema validator — will reject a body the server would have taken.
The failure lands on the client side, before the request is sent, and looks nothing like a server
problem.

Worth checking whether this is Swashbuckle's default here or something the project configured. If
it is configuration, dropping it is a one-line change and the schema stops lying. If it is a
default, the decision is whether to override it or to enforce it server-side.

---

## 3. `OrderDto.Id` is a request field nothing reads, documented as something it is not

```csharp
/// <summary>
/// User id.
/// </summary>
public Guid Id { get; set; }
```

`src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs`

Three separate problems in four lines:

- **Nothing reads it.** `OrderService.CreateOrderAsync` builds a `CreateOrderCommand` from `Asset`,
  `Amount`, `Price`, `UserId`, `Side`, `TradingType` and `Notes`. The order's id is generated
  server-side. A client that sends `id` has it silently discarded — confirmed by grep: there is no
  `request.Id` anywhere.
- **The doc comment is wrong**, and wrong in a way that invites a real mistake. It says "User id."
  on a property named `Id`, four lines above a separate `UserId` property. A developer reading the
  schema and this comment could reasonably send the user's id as `id` and the order's user as
  nothing.
- **The schema still advertises it**, as a `uuid`, with nothing marking it optional or
  server-assigned.

Until #235 it also carried `[Required(ErrorMessage = "شناسه کاربر الزامی است")]` — "user id is
required" — which is where the confusion is documented most plainly. That attribute is gone; the
field and its comment are not.

The fix is either delete the property or correct the comment, and deleting it is another change to
the request contract, which is why #235 flagged it instead of doing it
(`STANDARDS.md` §3, scope discipline). Correcting only the comment is safe and can happen at any
time.

---

## Suggested order of work

1. **(3)** — smallest, and safe in its comment-only form.
2. **(2)** — one investigation, possibly a one-line change; needs an answer before a typed client
   is generated.
3. **(1)** — a decision plus a paragraph, and the one with a deadline: it is cheapest while #97 has
   written nothing.

## Not in scope here

- #233 — two serialization libraries in one file.
- #230 — migration timing, which the `TradeDto` rename in option (3) above would run into.

## Context

- Found during #235 / PR #236, in the Swagger review that issue asked for as its fourth preventive
  measure.
- The camelCase rule and the two behavioural wire-contract guards from #235 already prevent a
  *new* field arriving with two names or a pinned spelling; none of them can see any of the three
  above, because all three are consistent-but-wrong rather than inconsistent.
- Relevant to #97, which is the first client that will read this schema instead of the C#.

MohKardan:
**Branch Name**: `fix/swagger-schema-tells-the-truth`
**Linked Issue**: closes #237

---

## Description

Three ways the published Swagger schema misleads whoever builds a client from it, found together
in the #235 schema review. None of them is a defect in the running system — each is the schema
saying something the servers do not do. All three are addressed here.

Everything below was measured against the three running services, before and after, by walking
`GET /swagger/v1/swagger.json` — not read off the C#.

---

## Type of Change

Done in the order the issue suggests: (3), then (2), then (1).

---

## Part 3 — `OrderDto.Id`

Confirmed nothing reads it, exactly as the issue asks:

```
$ grep -rn "request\.Id" --include="*.cs" .        # whole repo
(no matches)
```

`OrderService.CreateOrderAsync` builds its `CreateOrderCommand` from `Asset`, `Amount`, `Price`,
`UserId`, `Side`, `TradingType` and `Notes`; the order's id is generated server-side.
`BotHandler.cs:1459` never set the property — the value the old bot puts on the wire is
`Guid.Empty`, emitted only because the property existed.

The comment was corrected first (the safe half), then the property was deleted on the owner's
decision. The reason is recorded in the class's existing `<remarks>` block, next to #235's.

**One correction to the issue.** It says "the schema still advertises it … with nothing marking it
optional or server-assigned", and lists the wrong doc comment among the reasons. The comment never
reached the schema at all: `IncludeXmlComments` loads only the executing assembly's XML file, and
`OrderDto` lives in `TallaEgg.Core`, which sets no `GenerateDocumentationFile`. Measured — `id`
carried no `description` in the published schema either before or after:

```
OrderDto.id, BEFORE:  {"type": "string", "format": "uuid"}
OrderDto.id, AFTER:   absent
```

So the wrong comment misled a reader of the C#, and the schema misled a reader of the schema; they
were two problems, not one, and both are gone.

### Nothing breaks

`Id` becomes an unknown member, which `System.Text.Json` ignores — the same leniency part 2 is
about. An un-updated bot still posts it, and still gets an order. Measured over real HTTP against
the running service, below.

---

### The question the issue asks: default or configuration?

**Swashbuckle's default.** Not configuration — checked, not assumed:

```
$ grep -rn "SchemaFilter\|DocumentFilter\|OperationFilter\|AdditionalProperties\|UnmappedMember" --include="*.cs" .
(no matches)
```

All three `AddSwaggerGen` blocks set a title and `IncludeXmlComments`, and nothing else.
`SchemaGeneratorOptions` also offers no switch for it — in the shipped assembly's metadata,
`AdditionalPropertiesAllowed` exists only on `OpenApiSchema`, never as a generator option. So
overriding the default needs a filter; there is no line to delete.

### The decision: correct the schema, not the server

The issue already establishes that the leniency is load-bearing — it is what let an un-updated bot
keep talking to a post-#235 Orders.Api (PR #236's deployment section). Reproduced here on the
unmodified build, with the issue's own body:

```
POST /api/orders
{"Id":"00000000-…","symbol":"MAUA/IRT","asset":"MAUA/IRT","Amount":0.25,"quantity":0.25,"Price":23429694.35,…}

-> 200  success=True  asset=MAUA/IRT  amount=0.25
```

`symbol` and `quantity` do not exist in `OrderDto`, and `Id`/`Amount`/`Price` are PascalCase
spellings the schema does not declare either. The schema said that body was invalid. So the schema
was what was wrong.

### How

A `PreSerializeFilter` on each service's `UseSwagger` clears the flag. Clearing it omits the
keyword entirely, and an absent `additionalProperties` means "allowed" in OpenAPI 3 — which is what
these services actually do.

Only the blanket flag is cleared. A schema that names a type for its extra members is describing a
dictionary, which is a real statement about the payload; `HttpValidationProblemDetails` — the
framework's own, deliberately extensible — is untouched at `additionalProperties: {}`.

**Where it lives.** Once, in `TallaEgg.Core.Swagger.SwaggerSchemaConventions`; each service calls it
in one line. That follows `TallaEgg.Core.Cors.CorsExtensions`, which is the same thing for CORS —
cross-service startup wiring in the shared kernel, called from all five hosts, with
`CorsExtensionsTests` testing the helper directly.

It is a function over `OpenApiDocument` rather than the `ISchemaFilter` Swashbuckle offers for
mutating schemas, and that is forced: `ISchemaFilter` is a Swashbuckle type, and `Users.Api` pins
`Swashbuckle.AspNetCore` **7.0.0** while Orders and Wallet are on **9.0.3**, so a 9.0.3 reference in
`TallaEgg.Core` reaches Users.Api as a package downgrade (NU1605), which `TreatWarningsAsErrors`
makes a build failure. `Microsoft.OpenApi` carries no such problem — Users.Api's own Swashbuckle
already resolves 1.6.22, and Orders/Wallet resolve 1.6.23, so the 1.6.22 reference added to
`TallaEgg.Core` is satisfied everywhere and upgrades nothing:

```
```

The cost of that choice is one gap, recorded on the helper: a schema Swashbuckle inlines rather than
referencing would keep the flag. All 31 closed schemas were components, so it is hypothetical today.

The Swashbuckle version skew is real and pre-existing; flagged below, not fixed here.

---

## Part 1 — the two vocabularies

The owner chose option (1) plus (2): write the split down, and make `symbol`/`quantity` the
spelling for new endpoints. `docs/process/STANDARDS.md` §2 gains a subsection next to the JSON wire
format rule #235 added, carrying the issue's table, the rule, and why unifying is not a rename.

**One refinement to the issue's option (2).** "Pick `symbol`/`quantity`, the majority" can collide
with #235's own deciding principle — that a request agrees with its own response. A new endpoint
returning an `OrderHistoryDto` (`asset`/`amount`) but told to accept `symbol`/`quantity` would
reproduce exactly the mismatch #235 refused to create. The written rule is therefore ordered: match
the endpoint's own response first; where nothing constrains it, `symbol`/`quantity`.

The table was verified against the three live documents rather than copied:

```
wallet  TradeDto            symbol, quantity     orders  OrderDto             asset, amount
wallet  WalletRequest       asset,  amount       orders  OrderHistoryDto      asset, amount
orders  AcceptQuoteRequest  symbol, quantity     orders  PositionDto          symbol, quantity
orders  BestPricesDto       symbol               orders  PublishQuoteRequest  symbol
orders  TradeDto            symbol, quantity     orders  TradeHistoryDto      symbol, quantity
```

### How big option (3) would be, if it is ever wanted

Asked during review; measured, since the answer decides whether (3) stays on the table.

`TradeDto` is one class in `TallaEgg.Core` serving four roles: the outbox payload, the Orders→Wallet
request body for `/api/wallet/changeBalance`, part of the `POST /api/orders` response, and a
published schema in two services. Both the write (`OrderMatchingRepository.cs:207`) and the reads
(`OutboxProcessorService.cs:262,283`) use **default** `JsonSerializer` options — PascalCase in the
database, and **case-sensitive** on the way back.

```
OutboxMessages, local Orders database: 9,361 rows, every one carrying "Symbol" and "Quantity"
  Completed 8,651 · Failed 101 · Abandoned 609
```

The rows that must stay readable are Pending plus Failed — `/api/outbox/redrive-all-failed` and
`/api/outbox/{id}/redrive` re-drive Failed ones, while Abandoned "can never be re-driven again" by
design. The failure mode of getting this wrong is silent: a renamed property reading an old payload
yields `Symbol = ""` and `Quantity = 0`, a settlement for nothing, and no code on that path rejects
it.

So the code side is small and the risk is entirely in deployment. It needs (a) a data migration
rewriting the JSON keys in `OutboxMessages.Payload` — `scripts/migrate-irr-to-irt.sql` is the
precedent for a data-migrating `.sql` — or a drained and frozen queue across the deploy; (b) Orders
and Wallet shipped in lockstep, because `asset` versus `symbol` is a different name, not a different
case, so the case-insensitive binding that carried #235 through a mixed deployment does not bridge
it; and (c) six schemas across two services plus the `POST /api/orders` response. Worth doing only
before #97 ships against the current shape, and only as its own piece of work.

---

## Found, not fixed

Flagged per `STANDARDS.md` §3 rather than fixed inline.

1. **Two hand-written clients post PascalCase to a camelCase schema.** `WalletApiClient`'s
   `TradeTransactionAndBalanceChangeAsync` and `OrderApiClient`'s `SubmitOrderAsync` both serialize
   with default `JsonSerializer` options, so `/api/wallet/changeBalance` receives
   `{"Symbol":…,"Quantity":…}` and `/api/orders` receives `{"Asset":…,"Amount":…}` — the body printed
   in this PR's own testing section — while both published schemas say camelCase. It works only
   because minimal-API binding is case-insensitive, and tightening that (the natural follow-up to
   #229/#235) would break every bot order. Same family as #229 and #235, and invisible to their
   guards for the same reason: consistent-but-wrong rather than inconsistent.
2. **Five more server-assigned properties on `OrderDto`.** `Status`, `Role`, `CreatedAt`,
   `UpdatedAt` and `ParentOrderId` are never read by `CreateOrderAsync` — the same defect as the
   `Id` this PR removes, and `Role` and `Status` are worse, because a client can plausibly believe
   it is choosing the maker/taker role. Five more request-contract changes, so the owner's call and
   not folded in here. (`Type` looks like a sixth and is not: it is read, at `OrderService.cs:66`,
   in a log line.)
3. **`OrderDto.Price`'s doc comment is untrue.** It says price is "optional for market orders";
   `Orders.Api/Program.cs:679` rejects `Price <= 0` unconditionally and `CreateOrderAsync` never
   branches on `Type`. Exactly the defect this PR fixed on `Id`, one property away — left alone
   because the owner approved that one specific correction, not a licence to keep going.
4. **`OrderDto` declares no `required` while the endpoint 400s on three fields.** Deliberate, and
   one PR old: #236 deleted this DTO's DataAnnotations because they were unexecuted decoration
   reaching the schema, and took the `required` count to zero across all three services. Making the
   schema state real constraints means running validation for real, which is a platform-wide
   decision #236 already weighed and declined.
5. **Swashbuckle version skew.** `Users.Api` 7.0.0 versus 9.0.3 in the other two — what rules out a
   shared `ISchemaFilter`.
6. Untouched as the issue requires: #233 (two serialization libraries in one file) and #230
   (migration timing).

Items 1–4 are worth issues; say the word and I will open them.

---

### Build

```
```

### Unit tests

```
dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 884, Skipped: 0, Total: 884
```

880 on `main` plus 4 new, all in `SwaggerSchemaConventionsTests`: a closed object schema opens, a
dictionary schema keeps its value type, the already-open framework schema is not rewritten, and a
document with no components does not throw.

No existing test needed changing. `RequestDtoWireContractTests` drives itself off `JsonTypeInfo`, so
`OrderDto` losing a property is simply one fewer field to check.

```
scripts/check-doc-paths.sh
404 tracked text file(s) checked, every reference resolves.
```

### The published schema, before and after

The whole point of this issue is something only visible on the wire, so all three documents were
captured from the running services before any change and again at the end, and compared
programmatically rather than by eye:

| service | `additionalProperties: false` before | after | rest of the document |
|---|---|---|---|
| Users (5136) | 6 | 0 | identical |
| Wallet (60933) | 2 | 0 | identical |
| Orders (5140) | 23 | 0 | identical but for `OrderDto.id`, removed |

31 of 32 object schemas before, matching the issue's count exactly; `HttpValidationProblemDetails`
was the one exception and still is, still at `additionalProperties: {}`.

Asserted, not eyeballed: stripping every `additionalProperties: false` from the before-documents
makes Users and Wallet compare equal to the after-documents, and leaves Orders differing in exactly
one property.

### End-to-end

```
=== Done in 00:00:15.79. Registered 20 (18 approved), trades attempted 50, errors 0 ===
Filled trades by symbol:
```

Settlement counts are the delta the driver asserts, not absolutes. A baseline run on the unmodified
build beforehand was green on the same seed.

### The server is still lenient — proved over real HTTP, not argued

The issue's own extra-field body, sent to the running Orders.Api before the change and again after,
with the same user and the live MAUA/IRT quote:

```
{"Id":"00000000-…","symbol":"MAUA/IRT","asset":"MAUA/IRT","Amount":0.25,"quantity":0.25,"Price":…,…}

BEFORE  -> 200  success=True  asset=MAUA/IRT  amount=0.25
AFTER   -> 200  success=True  asset=MAUA/IRT  amount=0.25
```

Identical. After the change that body carries **three** kinds of undeclared field — `symbol` and
`quantity`, which never existed; `Id`, which this PR removes; and the PascalCase spellings — and it
is still accepted. That is the un-updated bot's payload, and it still places an order. The schema
simply no longer claims otherwise.

### The bot's own client, through the changed path

`driver.ps1 smoke` takes the quote-fill path (`AcceptQuoteAsync`), which never touches
`POST /api/orders`, so the changed DTO was driven separately through the bot's real
`OrderApiClient.SubmitOrderAsync`, with `OrderDto` built from exactly the seven properties
`BotHandler.cs:1459` sets, against the running Orders.Api and the real database:

```
Body the bot now puts on the wire:
  {"Asset":"MAUA/IRT","Amount":0.50,"Price":23483462.99,"UserId":"…","Side":0,"Type":1,…}
    - no "Id" at all any more

```

Read straight out of `TallaEggOrders.dbo.Orders`:

```
Asset='MAUA/IRT'  Amount=0.50000000  Price=23483462.99000000  Status=Confirmed
```

Prices are the live MAUA/IRT quote at the time (buy 23,483,462.99 / sell 23,530,476.94), not
invented figures. Every probe order was cancelled afterwards, so nothing rests in the book.

### Environment

---

## Security Checklist

- [x] Input validation unchanged — the hand-written checks in `Program.cs` are untouched, and the
      server's tolerance of unknown members is deliberately preserved, not widened

Swagger remains Development-only in all three services; this changes what the document says, not
who can reach it.

---

## Code Quality

- [x] Follows `STANDARDS.md` naming, including the section this PR adds
- [x] One dependency added, and only the one needed: `Microsoft.OpenApi` 1.6.22 in `TallaEgg.Core`,
      the version all three services already resolve. No Swashbuckle reference in the shared kernel,
      and no version moves anywhere.

---

## Documentation

- [x] `docs/process/STANDARDS.md` §2 gains the trading-pair vocabulary subsection
- [x] Swagger/OpenAPI: the schema is the subject of this PR; diffed above
- [x] `OrderDto`'s `<remarks>` records why `Id` went, alongside #235's reasoning

---

## Breaking Changes?

- [ ] No breaking changes
- [x] Breaking changes (see details below)

**Formally breaking**: the request contract of `POST /api/orders` loses the `id` field.

**In practice**: nothing breaks. The only caller is the bot, which never set it, and a body that
still sends `id` is accepted with the member ignored — measured above. The schema change in part 2
is strictly a relaxation: every body valid against the old schema is valid against the new one.

Per `STANDARDS.md` §11 this is MINOR-shaped, not MAJOR: no external integration exists to break. No
version bump is included — not this PR's call.

---

## Deployment Notes

Swagger is Development-only in all three services, so the two schema checks below belong here,
against a Development-configured run of the build being shipped — not after deployment, where the
endpoint is not mapped and would answer 404 whether the fix is present or missing.

- [x] `GET /swagger/v1/swagger.json` on each service contains no `"additionalProperties": false`
- [x] `OrderDto` in the Orders schema has no `id`

**Order**: the bot and `Orders.Api` may deploy in either order or together. The compatibility rests
on the same thing #236 measured — `PropertyNameCaseInsensitive` in `Orders.Api`, plus
`System.Text.Json` ignoring unknown members — and this PR deliberately preserves both.

**Post-deployment verification** — what production can actually answer:
- [ ] One order placed from the bot appears in `Orders` with the right `Asset` and `Amount`

**Rollback**: revert the commit and redeploy. An older `Orders.Api` reads the new bot's payload
correctly (it simply finds no `id`, which nothing read), so rolling back the API alone is safe too.

---

## Related Issues

- Closes #237
- Follows #235 / PR #236, whose Swagger review found all three of these
- Relevant to #97 (browser client) — the reason this was worth doing before it starts
- Untouched as the issue requires: #233, #230

---

## Final Checklist

- [x] All tests pass locally
- [x] Commit message is clear

### DIFF
--- a/docs/process/STANDARDS.md
+++ b/docs/process/STANDARDS.md
@@ -103,9 +103,13 @@ A trading pair is spelled `asset` in some schemas and `symbol` in others, and it
 `amount` or `quantity` along exactly the same line. This is real and pre-existing, not a mistake
 waiting to be tidied:
 
-| `asset` / `amount` | `symbol` / `quantity` |
-|---|---|
-| `OrderDto`, `OrderHistoryDto`, `WalletRequest` | `AcceptQuoteRequest`, `BestPricesDto`, `PublishQuoteRequest`, `TradeDto`, `TradeHistoryDto`, `PositionDto` |
+| | `asset` / `amount` | `symbol` / `quantity` |
+|---|---|---|
+| **request bodies** | `OrderDto`, `WalletRequest` | `AcceptQuoteRequest`, `PublishQuoteRequest`, `TradeDto` |
+| **responses** | `OrderHistoryDto` | `BestPricesDto`, `TradeDto`, `TradeHistoryDto`, `PositionDto` |
+
+`TradeDto` is on both rows: it is the body Orders posts to `/api/wallet/changeBalance` and part of
+what `POST /api/orders` returns.
 
 The two order-entry paths sit on opposite sides of it: `POST /api/orders` takes `asset`/`amount`,
 `POST /api/quotes/accept` takes `symbol`/`quantity`. A client that places orders has to know both
--- a/docs/process/STANDARDS.md
+++ b/docs/process/STANDARDS.md
@@ -97,6 +97,38 @@ sentences and holds the allowlist for the third; the two behavioral guards in
 `RequestDtoWireContractTests.cs` enforce the last. Adding a name to that allowlist is a contract
 decision, not a formatting one.
 
+#### Trading-pair vocabulary: two spellings, and which one a new endpoint takes
+
+A trading pair is spelled `asset` in some schemas and `symbol` in others, and its quantity is
+`amount` or `quantity` along exactly the same line. This is real and pre-existing, not a mistake
+waiting to be tidied:
+
+| `asset` / `amount` | `symbol` / `quantity` |
+|---|---|
+| `OrderDto`, `OrderHistoryDto`, `WalletRequest` | `AcceptQuoteRequest`, `BestPricesDto`, `PublishQuoteRequest`, `TradeDto`, `TradeHistoryDto`, `PositionDto` |
+
+The two order-entry paths sit on opposite sides of it: `POST /api/orders` takes `asset`/`amount`,
+`POST /api/quotes/accept` takes `symbol`/`quantity`. A client that places orders has to know both
+and pick per endpoint.
+
+**A new endpoint takes the spelling its own response already uses.** Where nothing constrains it —
+a genuinely new pair of request and response — use `symbol`/`quantity`, the larger group. That
+order matters and is not the same as "follow the majority": #235 kept `OrderDto` on `asset`/`amount`
+precisely because `OrderHistoryDto` is the response of that very endpoint, and a client works one
+endpoint at a time, so a request and response that disagree cost it more than a platform-wide
+majority it can look up once. Applying the majority rule first would reintroduce the mismatch that
+issue refused to create.
+
+**Unifying the two is not a rename**, which is why the split is written down rather than removed
+(issue #237). `TradeDto` is also the outbox payload: it is serialized into `OutboxMessages.Payload`
+with a bare `JsonSerializer` and read back case-sensitively, so renaming its properties silently
+turns every payload not yet settled into a settlement for an empty symbol and zero quantity.
+Changing it means a data migration over the stored payloads, plus a lockstep deployment of Orders
+and Wallet — `asset` versus `symbol` is a different name, not a different case, so the
+case-insensitive binding that makes ordinary DTO renames survive a mixed deployment does not bridge
+it. Worth doing only before a browser client (#97) ships against the current shape, and only as its
+own piece of work.
+
 #### Branch Names (Git)
 - **Feature**: `feat/{description}` (e.g., `feat/add-wallet-transaction`)
 - **Bugfix**: `fix/{description}` (e.g., `fix/null-reference-wallet`)
--- a/src/Order/Orders.Api/Program.cs
+++ b/src/Order/Orders.Api/Program.cs
@@ -323,7 +323,25 @@ app.UseTallaEggCors();
 // bought for nothing.
 if (app.Environment.IsDevelopment())
 {
-    app.UseSwagger();
+    // Swashbuckle stamps `additionalProperties: false` on every object schema it generates and
+    // offers no option to turn that off, but no service enforces it: System.Text.Json ignores
+    // unknown members and nothing sets JsonUnmappedMemberHandling.Disallow. That leniency is
+    // deliberate and load-bearing — it is what lets an un-updated bot keep talking to a post-#235
+    // API — so the schema is what is wrong here, not the server. Clearing the flag omits the
+    // keyword, and an absent `additionalProperties` means "allowed" in OpenAPI 3, which is what
+    // these services actually do. Issue #237.
+    app.UseSwagger(options => options.PreSerializeFilters.Add((document, _) =>
+    {
+        foreach (var schema in document.Components.Schemas.Values)
+        {
+            // A schema that names a type for its extra members is describing a dictionary, which
+            // is a real statement about the payload. Only the blanket flag is cleared.
+            if (schema.AdditionalProperties is null)
+            {
+                schema.AdditionalPropertiesAllowed = true;
+            }
+        }
+    }));
     app.UseSwaggerUI(c =>
     {
         c.SwaggerEndpoint("/swagger/v1/swagger.json", "TallaEgg Orders API V1");
--- a/src/Order/Orders.Api/Program.cs
+++ b/src/Order/Orders.Api/Program.cs
@@ -23,6 +23,7 @@ using TallaEgg.Core.Startup;
 using TallaEgg.Infrastructure.Clients;
 using TallaEgg.TelegramBot.Infrastructure.Clients;
 using CancelActiveOrdersResponseDto = TallaEgg.Core.DTOs.Order.CancelActiveOrdersResponseDto;
+using TallaEgg.Core.Swagger;
 
 var builder = WebApplication.CreateBuilder(args);
 
@@ -323,25 +324,11 @@ app.UseTallaEggCors();
 // bought for nothing.
 if (app.Environment.IsDevelopment())
 {
-    // Swashbuckle stamps `additionalProperties: false` on every object schema it generates and
-    // offers no option to turn that off, but no service enforces it: System.Text.Json ignores
-    // unknown members and nothing sets JsonUnmappedMemberHandling.Disallow. That leniency is
-    // deliberate and load-bearing — it is what lets an un-updated bot keep talking to a post-#235
-    // API — so the schema is what is wrong here, not the server. Clearing the flag omits the
-    // keyword, and an absent `additionalProperties` means "allowed" in OpenAPI 3, which is what
-    // these services actually do. Issue #237.
-    app.UseSwagger(options => options.PreSerializeFilters.Add((document, _) =>
-    {
-        foreach (var schema in document.Components.Schemas.Values)
-        {
-            // A schema that names a type for its extra members is describing a dictionary, which
-            // is a real statement about the payload. Only the blanket flag is cleared.
-            if (schema.AdditionalProperties is null)
-            {
-                schema.AdditionalPropertiesAllowed = true;
-            }
-        }
-    }));
+    // Swashbuckle stamps `additionalProperties: false` on every object schema and no service
+    // enforces it; an absent `additionalProperties` means "allowed" in OpenAPI 3, which is what
+    // these services actually do. Why, at length, on the helper (issue #237).
+    app.UseSwagger(options =>
+        options.PreSerializeFilters.Add((document, _) => document.AllowUndeclaredMembers()));
     app.UseSwaggerUI(c =>
     {
         c.SwaggerEndpoint("/swagger/v1/swagger.json", "TallaEgg Orders API V1");
--- a/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs
+++ b/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs
@@ -47,10 +47,11 @@ namespace TallaEgg.Core.DTOs.Order
     /// Orders.Api validates this request by hand instead.
     ///
     /// An <c>Id</c> property went the same way (issue #237). It was a request field nothing read —
-    /// an order's id is assigned by the server — documented as "User id." four lines above the
-    /// separate <see cref="UserId"/>, so a client following it could reasonably have sent the user's
-    /// id as <c>id</c> and left the order's user empty. Removing it costs no caller anything: the
-    /// bot never set it, and a body that still carries <c>id</c> is accepted with the member ignored.
+    /// an order's id is assigned by the server — and it was documented as "User id." while a
+    /// separate <see cref="UserId"/> sat a few properties further down the same class, so a client
+    /// following that comment could reasonably have sent the user's id as <c>id</c> and left the
+    /// order's user empty. Removing it costs no caller anything: the bot never set it, and a body
+    /// that still carries <c>id</c> is accepted with the member ignored.
     /// </remarks>
     public class OrderDto
     {
--- a/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs
+++ b/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs
@@ -45,14 +45,15 @@ namespace TallaEgg.Core.DTOs.Order
     /// calls <c>Validator.TryValidateObject</c> or registers a validation filter — so they enforced
     /// nothing while making the generated schema advertise constraints the server does not apply.
     /// Orders.Api validates this request by hand instead.
+    ///
+    /// An <c>Id</c> property went the same way (issue #237). It was a request field nothing read —
+    /// an order's id is assigned by the server — documented as "User id." four lines above the
+    /// separate <see cref="UserId"/>, so a client following it could reasonably have sent the user's
+    /// id as <c>id</c> and left the order's user empty. Removing it costs no caller anything: the
+    /// bot never set it, and a body that still carries <c>id</c> is accepted with the member ignored.
     /// </remarks>
     public class OrderDto
     {
-        /// <summary>
-        /// User id.
-        /// </summary>
-        public Guid Id { get; set; }
-
         /// <summary>
         /// Asset symbol, as a trading pair — for example <c>MAUA/IRT</c>.
         /// </summary>
--- a/src/TallaEgg/TallaEgg.Core/Swagger/SwaggerSchemaConventions.cs
+++ b/src/TallaEgg/TallaEgg.Core/Swagger/SwaggerSchemaConventions.cs
@@ -1,0 +1,52 @@
+using Microsoft.OpenApi.Models;
+
+namespace TallaEgg.Core.Swagger;
+
+/// <summary>
+/// Makes a generated OpenAPI document describe what these services actually do.
+/// </summary>
+public static class SwaggerSchemaConventions
+{
+    /// <summary>
+    /// Removes the blanket <c>additionalProperties: false</c> Swashbuckle stamps on every object
+    /// schema it generates.
+    /// </summary>
+    /// <remarks>
+    /// No service enforces that keyword and none is meant to (issue #237). <c>System.Text.Json</c>
+    /// ignores unknown members and nothing sets <c>JsonUnmappedMemberHandling.Disallow</c>, so a
+    /// request body carrying undeclared fields is accepted — deliberately: that tolerance is what
+    /// lets an un-updated bot keep talking to a post-#235 API. A client generated from the closed
+    /// schema, or one running the body through a validator, would reject a request the server
+    /// would have taken, and the failure lands before the request is ever sent.
+    ///
+    /// Responses are opened too, and for a reason of their own rather than as collateral: adding a
+    /// field to a response is a MINOR change under <c>STANDARDS.md</c> §11, so a closed response
+    /// schema promises something the platform does not — it would turn every such addition into a
+    /// break for any client that validates what it receives.
+    ///
+    /// It is cleared here rather than at generation time because an <c>ISchemaFilter</c> — the
+    /// hook Swashbuckle offers for this, since <c>SchemaGeneratorOptions</c> carries no switch for
+    /// it — is a Swashbuckle type, and this assembly cannot reference Swashbuckle: Users.Api pins
+    /// 7.0.0 while Orders and Wallet are on 9.0.3, and a 9.0.3 reference here reaches Users.Api as
+    /// a package downgrade (NU1605), which <c>TreatWarningsAsErrors</c> makes a build failure.
+    /// <c>Microsoft.OpenApi</c>, which this needs instead, is already resolved by all three.
+    ///
+    /// A schema that names a type for its extra members is describing a dictionary, which is a
+    /// real statement about the payload; only the blanket flag is cleared.
+    /// </remarks>
+    public static void AllowUndeclaredMembers(this OpenApiDocument document)
+    {
+        ArgumentNullException.ThrowIfNull(document);
+
+        var schemas = document.Components?.Schemas;
+        if (schemas is null) return;
+
+        foreach (var schema in schemas.Values)
+        {
+            if (schema.AdditionalProperties is null)
+            {
+                schema.AdditionalPropertiesAllowed = true;
+            }
+        }
+    }
+}
--- a/src/TallaEgg/TallaEgg.Core/TallaEgg.Core.csproj
+++ b/src/TallaEgg/TallaEgg.Core/TallaEgg.Core.csproj
@@ -17,6 +17,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!-- The OpenAPI object model only, not Swashbuckle: Users.Api pins Swashbuckle 7.0.0
+         while Orders and Wallet are on 9.0.3, so a Swashbuckle reference here would reach
+         Users.Api as a downgrade (NU1605). 1.6.22 is the version its 7.0.0 already
+         resolves. See Swagger/SwaggerSchemaConventions.cs (issue #237). -->
+    <PackageReference Include="Microsoft.OpenApi" Version="1.6.22" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageReference Include="Serilog.AspNetCore" Version="9.0.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="7.0.0" />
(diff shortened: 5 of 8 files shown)


MohKardan:
## Review response

Twelve findings, self-review with `/code-review 239`. Every factual claim was checked against the
code before I accepted or rejected it. Six are fixed in 018d3ee; four are real and out of scope, and
I say below why rather than just citing §3; two are wrong in part, and I say which part.

---

### Fixed

**The NU1605 justification for triplicating the filter was wrong, and the triplication with it.**

This is the finding that mattered. My description argued a shared helper would need a Swashbuckle
reference in `TallaEgg.Core`, which would reach `Users.Api` (Swashbuckle 7.0.0) as a downgrade from
Orders/Wallet's 9.0.3. That is true of an `ISchemaFilter`, which is a Swashbuckle type — but not of
a function over `OpenApiDocument`, which needs only `Microsoft.OpenApi`. And `Users.Api` already
resolves `Microsoft.OpenApi` 1.6.22 through its own Swashbuckle:

```
src/User/Users.Api    -> Microsoft.OpenApi/1.6.22
src/Order/Orders.Api  -> Microsoft.OpenApi/1.6.23
src/Wallet/Wallet.Api -> Microsoft.OpenApi/1.6.23
```

A `1.6.22` reference in `TallaEgg.Core` is satisfied, not downgraded, in all three. Verified rather
than reasoned:

```
```

So the filter now lives once, in `TallaEgg.Core.Swagger.SwaggerSchemaConventions`, and each service
calls it in one line. The review is also right that this is the repository's established shape:
`TallaEgg.Core.Cors.CorsExtensions` is exactly this — cross-service startup wiring in the shared
kernel, called from all five hosts, with `CorsExtensionsTests` testing the helper directly rather
than scanning the call sites.

The published document is **byte-identical before and after the hoist**, on all three services —
compared programmatically, not by eye.

**A guard test, in the form the repository actually uses.** The review asked for one and proposed a
source-text sweep like `JsonNamingPolicyConsistencyTests`. Hoisting makes that unnecessary: the
helper is a plain function over `OpenApiDocument`, so the test project can reference it and test the
behaviour, which is strictly better than scanning three files for a token.
`SwaggerSchemaConventionsTests` covers four cases — a closed object schema opens, a dictionary
schema keeps its value type, the already-open framework schema is not rewritten, and a document with
no components does not throw. 880 → 884.

**The null `Components` dereference.** Taken. It costs one character and the filter runs inside the
middleware, so a throw there is a 500 on the schema endpoint rather than a startup failure — a
cheap guard in the wrong place is still cheap, and this is the right place.

**Response schemas were opened on a rationale that only covered requests.** A fair hit. My comment
justified the change with "`System.Text.Json` ignores unknown members", which is about request
binding, while the filter also opens `OrderHistoryDto`, `BestPricesDto`, the `ApiResponse<>`
wrappers and the rest.

I kept the behaviour and fixed the reasoning, because opening them is right for a reason of its own:
a closed response schema promises something this platform does not offer. Adding a field to a
response is a MINOR change under `STANDARDS.md` §11 — no external integration to break — and a
closed schema turns every such addition into a break for any client that validates what it receives.
`additionalProperties: false` on a response was true of one build, not of the contract. That is now
what the helper's `<remarks>` says.

**"Four lines above the separate `UserId`."** Wrong, and I had copied it from the issue without
checking. `Id` sat at line 54 and `UserId` at line 69, with `Asset`, `Amount` and `Price` and their
comments in between. Reworded to "a few properties further down the same class". A paragraph whose
whole job is to record why a wrong comment was dangerous cannot itself contain a wrong fact.

**The vocabulary table did not say which entries are requests and which are responses.** The rule
under it is "a new endpoint takes the spelling its own response already uses", which cannot be
applied from a flat list of nine type names. Split into request and response rows, with `TradeDto`
appearing on both and a line saying why.

---

### Right, and deliberately not in this PR

**Five more server-assigned properties on `OrderDto`.** Correct, and I checked it:

```
$ grep -o "request\.[A-Za-z]*" src/Order/Orders.Application/OrderService.cs | sort | uniq -c
      7 request.Price      7 request.Asset     4 request.UserId
      4 request.Side       4 request.Amount    1 request.TradingType
      1 request.Type       1 request.Notes
```

`Status`, `Role`, `CreatedAt`, `UpdatedAt` and `ParentOrderId` are never read. The review is right
that `Role` and `Status` are worse than the `Id` this PR removes, because a client can plausibly
believe it is choosing the maker/taker role.

One correction: it lists `Type` among them, and `Type` **is** read — at `OrderService.cs:66`. Only
in a log line, so it changes no behaviour, but it is not in the same category as the other five.

Not fixed here because the owner was asked specifically about `Id` and answered about `Id`.
Deleting five more properties is five more request-contract changes, which is the owner's call, not
mine (`STANDARDS.md` §3). It is worth its own issue and I will open one on request.

**`OrderDto` declares no `required` while the endpoint 400s on three fields.** True — the server
rejects an empty `Asset`, `Amount <= 0` and `Price <= 0` at `Orders.Api/Program.cs:673-680`, and the
schema marks nothing required.

But this is not an oversight, it is one PR old and deliberate: #236 deleted the DataAnnotations from
this DTO precisely because they were unexecuted decoration reaching the schema, and recorded that
`OrderDto` had been the only schema in any service declaring `required`, `maxLength` or `minimum`,
and that the count was now zero everywhere. Re-adding `required` to one DTO makes it the exception
again. Getting the schema to state real constraints means running validation for real, which #236
also weighed and declined as a platform-wide decision — all endpoints or none. Same answer here.

**`OrderApiClient.SubmitOrderAsync` serializes PascalCase against a camelCase schema.** Correct, and
a good catch that it is the same defect I filed as "found, not fixed" for `WalletApiClient` while
printing the evidence for this one in my own testing section. Added to that list — it belongs in the
same issue, and that issue is not this one.

**The `Price` doc comment is untrue.** Verified: `Orders.Api/Program.cs:679` rejects `Price <= 0`
unconditionally and `CreateOrderAsync` never branches on `request.Type`, so "optional for market
orders" is false. Same class as the `Id` comment, same file.

I am flagging rather than fixing it: the owner approved a comment correction for one specific
property after the issue argued that specific case, and helping myself to the next one because it is
nearby is how a scoped change stops being one. It is a one-line fix and I will make it the moment
it is asked for.

---

### Partly wrong

**"An `ISchemaFilter` sees inline schemas a `PreSerializeFilter` misses."** The mechanism is right
and I have made the comment stop implying Swashbuckle offers no hook at all — `SchemaGeneratorOptions`
has no switch for this, but `ISchemaFilter` exists and is the generation-time hook.

It is still not available here, for the reason the same review disproved for the *document* filter:
`ISchemaFilter` is a Swashbuckle type, so a shared one needs Swashbuckle in `TallaEgg.Core`, and
*that* reference genuinely is an NU1605 downgrade for Users.Api's pinned 7.0.0. The choice is a
shared document filter or three copies of a schema filter, and after this review it is the shared
one. The residual gap — an inlined object schema keeping the flag — is real, hypothetical today (all
31 closed schemas were components, measured), and now recorded in the helper's `<remarks>` instead
of being invisible.

**The post-deployment checklist could not be performed.** Right about the fact and I have corrected
the checklist: Swagger is Development-only in all three services, so `GET /swagger/v1/swagger.json`
404s on the production VM and those two items could never have been ticked there.

They were not pointless, they were in the wrong section — they are how you verify the build before
it ships, not the deployment after. Moved to pre-deployment, against a Development-configured run of
the same build. What remains post-deployment is what production can actually answer: the services
start, and an order placed from the bot lands with the right `Asset` and `Amount`.

---

### Re-verified after the changes

```
dotnet build TallaEgg.sln --no-incremental   Build succeeded.  0 Warning(s)  0 Error(s)
dotnet test TallaEgg.sln                     Passed! - Failed: 0, Passed: 884, Total: 884
scripts/check-doc-paths.sh                   404 tracked text file(s) checked, every reference resolves.

  BTC/IRT: 17 filled   MAUA/IRT: 17 filled   SEKE_BAHAR/IRT: 16 filled

POST /api/orders with the issue's extra-field body -> 200, order created, then cancelled
Published document on all three services: byte-identical to the pre-hoist capture
```

Issue #243, opened 2026-09-07
Swashbuckle is two majors apart across the three APIs (7.0.0 vs 9.0.3), which forecloses a shared ISchemaFilter


MohKardan:
`Users.Api` is on `Swashbuckle.AspNetCore` **7.0.0**; `Orders.Api` and `Wallet.Api` are on **9.0.3**.
Two majors apart, in three projects that do the same job.

Nothing is broken by it. It surfaced in #237 because it is not free: it decides what shape a shared
Swagger helper can take, and it will decide the same thing again next time.

Measured on `main` at `4958452`.

---

## What is there

```
src/User/Users.Api/Users.Api.csproj:21      Swashbuckle.AspNetCore  7.0.0
src/Order/Orders.Api/Orders.Api.csproj:23   Swashbuckle.AspNetCore  9.0.3
src/Wallet/Wallet.Api/Wallet.Api.csproj:21  Swashbuckle.AspNetCore  9.0.3
```

Resolved transitively, they are also two versions of the OpenAPI object model, though a compatible
pair:

```
src/User/Users.Api    -> Microsoft.OpenApi/1.6.22
src/Order/Orders.Api  -> Microsoft.OpenApi/1.6.23
src/Wallet/Wallet.Api -> Microsoft.OpenApi/1.6.23
```

## What it cost, concretely

#237 needed the same schema convention applied in all three services. The natural shape for that in
Swashbuckle is an `ISchemaFilter`, registered once next to `IncludeXmlComments`, which also sees
schemas Swashbuckle inlines rather than referencing.

That shape is unavailable. `ISchemaFilter` is a Swashbuckle type, so a shared one has to live in
`TallaEgg.Core` with a Swashbuckle reference — and a 9.0.3 reference there reaches `Users.Api`,
which pins 7.0.0 directly, as a package downgrade (NU1605). `TreatWarningsAsErrors` makes that a
build failure.

So #237 shipped a document-level filter over `Microsoft.OpenApi` instead
(`TallaEgg.Core.Swagger.SwaggerSchemaConventions`), which works, is tested, and produces a
byte-identical document — but carries a recorded gap: an inlined object schema would keep the flag.
Every object schema is a component today, so the gap is hypothetical. It stops being hypothetical
the day someone adds an anonymous request body.

That is the real cost of the skew: not a bug, a foreclosed option, recorded in a `<remarks>` block
that the next person has to find.

## Why it is probably a small fix

[...]

Issue #240, opened 2026-09-07
OrderDto still carries five request fields nothing reads, and a doc comment that contradicts the server


MohKardan:
`OrderDto` still carries five request fields nothing reads, and one doc comment that is false. Both
are what #237 fixed for `Id`, one property over; both were found in that PR's review and
deliberately left out of it, because #237's owner approved a decision about `Id` and not a licence
to keep going.

Measured on `main` at `4958452`, after #237 merged.

---

## 1. Five properties nothing reads

`POST /api/orders` binds `OrderDto`. Every `request.*` read on that path:

```
$ grep -o "request\.[A-Za-z]*" src/Order/Orders.Application/OrderService.cs | sort | uniq -c
      7 request.Price      7 request.Asset     4 request.UserId
      4 request.Side       4 request.Amount    1 request.TradingType
      1 request.Type       1 request.Notes

$ grep -n "request\." src/Order/Orders.Api/Program.cs   # the /api/orders handler
      request.Asset, request.Amount, request.Price      # the three hand-written checks
```

So these five are bound, published in the schema, and discarded:

| property | what actually decides it |
|---|---|
| `Status` | `Order.CreateMakerOrder` / `CreateTakerOrder` |
| `Role` | computed from the matching result |
| `CreatedAt` | the entity |
| `UpdatedAt` | the entity |
| `ParentOrderId` | set by the matching engine when a taker is linked to a maker |

[`src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs:73-81`](../blob/main/src/TallaEgg/TallaEgg.Core/DTOs/Order/OrderDtos.cs#L73-L81)

**`Role` and `Status` are worse than the `Id` #237 removed.** A client reading the schema sees an
enum called `role` on an order-entry request and can reasonably believe it is choosing whether to
post as maker or taker — a real trading concept in this system, with real consequences for fees and
collateral. It sends `{"role": 1, "status": 3, "parentOrderId": "<maker-id>"}`, gets `200`, and
receives a plain maker order at whatever status the entity assigned. No error, no warning.

`Id` at least looked like an identifier a server would assign. These look like choices.

**Not on this list:** `Type`. It *is* read, at
[`OrderService.cs:66`](../blob/main/src/Order/Orders.Application/OrderService.cs#L66) — but only in
a log line, and `CreateOrderAsync` never branches on it. It behaves like the other five and is
documented differently by accident, which is its own small question; see part 2.

[...]

PR #245, opened 2026-09-08
[Docs] Record the request-validation contract the schemas do not state — issue #242

### LINKED ISSUE #242: The schemas declare no required fields while the endpoints reject three: the validation decision #236 deferred twice
#237 made the schema stop claiming a constraint the servers do not enforce
(`additionalProperties: false`). The mirror image is still there: **the schemas declare no
constraints at all, while the endpoints enforce several.** A client built from the document sends a
body the document says is valid and gets a 400.

This is a decision that was deferred twice, by #235/PR #236 and again by #237, each time
deliberately and each time without anywhere to record it. This issue is that record.

Measured on `main` at `4958452`, against the three published documents.

---

## What the documents say

```
required        declared anywhere in Users, Wallet or Orders:  0
maxLength       declared anywhere:                             0
minimum         declared anywhere:                             0
```

Every property of every request body is optional and unbounded, in all three services.

## What the servers do

`POST /api/orders` rejects three things by hand, because minimal APIs do not execute
DataAnnotations and nothing in this repository calls `Validator.TryValidateObject`, registers a
validation filter, or references FluentValidation:

```csharp
// src/Order/Orders.Api/Program.cs:660-667
if (string.IsNullOrWhiteSpace(request.Asset))  -> 400 "نماد معاملاتی الزامی است"
if (request.Amount <= 0)                       -> 400 "مقدار سفارش باید بیشتر از صفر باشد"
if (request.Price <= 0)                        -> 400 "قیمت برای سفارش محدود الزامی است"
```

`OrderService.CreateOrderAsync` then adds more (`ValidateTradingLimits`, the balance and credit
check), and other endpoints have their own hand-written checks.

So a generated client sends `{"asset": "MAUA/IRT", "amount": 0.25}` — valid against the schema,
since nothing is required — and gets a 400 in Persian about limit orders. The failure is at least
loud, unlike the `additionalProperties` one, which failed silently on the client side.

## How it got here, and why nothing is a mistake

`OrderDto` used to carry `[Required]`, `[Range]` and `[StringLength]`. They were never executed, but
they *did* reach the schema — so the document advertised a length limit and a minimum the server
never applied, and marked the wrong (alias) properties as required. PR #236 deleted them, and
recorded that `OrderDto` had been the only schema in any service declaring `required`, `maxLength`
or `minimum`, so the count went to zero everywhere and the DTO stopped being the exception.

That was the right call for #235: attributes that lie are worse than no attributes. It leaves the
opposite gap, which #236 named and declined to close, because closing it properly means running
validation for real — and that is a platform-wide decision, all endpoints or none, that changes the
400 body of live trading endpoints.

#237's review raised it again and got the same answer for the same reason.

## The actual decision

Three options, cheapest first.

1. **Leave it, and write it down.** Add to `STANDARDS.md` §2 that request schemas state no
   constraints, that validation is hand-written per endpoint, and that a client must expect a 400 it
   cannot predict from the document. Costs nothing, and stops the next reviewer re-raising it — which
   has now happened twice.
2. **Declare, without enforcing differently.** Add a Swashbuckle schema filter that marks the fields
   the hand-written checks already require. The document becomes accurate; no runtime behaviour
   changes; the risk is that the filter and the checks drift, which is how the DataAnnotations got
   into this state in the first place.
3. **Enforce for real.** A validation filter or endpoint filter that runs DataAnnotations, with the
   attributes restored on the DTOs. The schema then describes enforcement because it *is*
   enforcement. This changes the 400 response body of every endpoint — today they return
   `{"success": false, "message": "…"}` with a Persian message, and a filter returns
   `HttpValidationProblemDetails` — so it needs the bot's error handling checked end to end, and it
   is a bigger change than it sounds.

(2) is tempting and is probably the trap: it recreates exactly the "schema promises what the server
does not run" shape that #236 deleted, just in a different file.

My recommendation is **(1) now**, and (3) only if #97 finds the unpredictable 400s genuinely
painful. A browser client can handle a 400 it did not predict; what it cannot handle is a document
that is wrong, and after #237 the document is no longer wrong — it is silent.

## Related

- #235 / PR #236 — deleted the unexecuted attributes and deferred this
- #237 — fixed the over-statement; this is the under-statement, deferred again
- #97 — the client that would meet these 400s

MohKardan:
**Branch Name**: `docs/record-request-validation-contract`
**Linked Issue**: closes #242

---

## Description

The published schemas declare **no constraints at all** — `required`, `maxLength` and `minimum` are
zero in all three services — while the endpoints enforce roughly fifty hand-written checks. A client
built from the document sends a body the document calls valid and gets refused.

#242 is not a defect report, it is a decision that had been deferred twice with nowhere to record
it. The owner chose **option (1): leave the behavior, write the contract down.** This is that
record.

Everything below was measured against the three running services on 2026-09-08 by walking
`GET /swagger/v1/swagger.json` and probing each writable endpoint over real HTTP — not read off the
C#.

---

## Type of Change
- [ ] Hotfix (urgent bug fix)
- [x] Documentation (docs/comments only)

---

## The measurement

The issue claims three zeros. Confirmed, and then some:

```
required (schema-level)   Users 0   Wallet 0   Orders 0
maxLength                 Users 0   Wallet 0   Orders 0
minimum                   Users 0   Wallet 0   Orders 0
```

Zero as well, everywhere: `minLength`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `pattern`,
`minItems`, `maxItems`, `uniqueItems`, `multipleOf`. The only validation keyword in any of the three
documents is `enum` — 7 occurrences, every one generated from a C# enum type rather than declared.

`required: true` does appear 56 times and is **not** a counter-example: every occurrence is on a
**parameter** object, never a schema. Those are path parameters, which OpenAPI requires to be marked
required. No request-body property is required in any service.

Scale: 51 paths, 39 object schemas, 195 properties. 24 writable endpoints (4 Users, 6 Wallet, 14
Orders), 14 distinct request schemas, 65 properties between them.

The issue's headline example, reproduced exactly:

```
POST /api/orders   {"asset": "MAUA/IRT", "amount": 0.25}
-> 400 {"success":false,"message":"قیمت برای سفارش محدود الزامی است"}
```

## Changes Made

- **`docs/design/API_REQUEST_VALIDATION.md`** (new) — every hand-written check on all 24 writable
  endpoints, with its file, line, kind (shape / state / cross-field) and refusal message; what a
  schema-valid body actually gets; why the state is deliberate; what a client should do; and how to
  re-measure the whole thing.
- **`docs/process/STANDARDS.md` §2** — the rule, next to the wire-format rules #235 and #237 left
  there, pointing at the document. Includes why a schema filter that declares constraints without
  enforcing them is not the fix.
- **`docs/process/INDEX.md`** — `docs/design/` added to "Existing today" plus a navigation line.
  `API_CONTRACT.md` and `DATABASE_SCHEMA.md` stay on the "Planned" list; this document is neither of
  them and is not named as if it were.

`docs/design/` did not exist. `STANDARDS.md` §4 has described it as the home for API contracts all
along, and `INDEX.md` listed it under "Planned", so this creates it rather than inventing a location.

## Three things the issue did not have

Found while measuring, and the reason the inventory was worth building rather than trusting the
three-line summary.

**1. A schema-valid body lands in five outcomes, not one.** `{}` produces `400`, `404`, `200` with
`success: false`, `200` having *acted*, or `500`. A client cannot key off the status code; `success`
is the reliable signal.

**2. Two endpoints answer 500, not 400.** Worse than an unpredictable 400 — the envelope carries a
generic message with nothing to show a customer:

```
POST /api/quotes/accept   {"quantity": 1}   -> 500 {"success":false,"message":"خطای داخلی سرور"}
```

`NullReferenceException` on a null `symbol`, confirmed in `run-logs/orders.out.log`.
`QuoteFillService.cs:70` does `symbol.Split('/')[0]` with no null check, and the quantity guard at
`:59` runs first, so a body that satisfies the quantity check reaches the split.

**3. Four endpoints accept an empty body and act on the C# defaults.** An omitted property is not an
error, it is a value:

| endpoint | omitted | what gets written |
|---|---|---|
| `POST /api/user/update-phone` | `phoneNumber`, `telegramId` | `""` onto the user whose TelegramId is `0` |
| `PUT /api/user/status` | `newStatus`, `telegramId` | `Pending` onto the user whose TelegramId is `0` |
| `POST /api/autoquote-settings/{Base}/{Quote}/enabled` | `isEnabled` | `false` — auto-quoting **off** |
| `POST /api/symbols/{Base}/{Quote}/active` | `isActive` | `false` — the symbol **disabled** |

`TelegramId` is a `long`, its default is `0`, and the seeded root administrator's TelegramId is `0`
(migration `20250918224949_init`). **An empty body addressed to the platform's SuperAdmin row is a
successful write, answered 200.** Discovered by hitting it — see "One accidental write" below. The
last two rows were derived from the code and deliberately not fired, because firing them would have
disabled a live trading symbol.

## Why option (1), and not (2) or (3)

The issue calls (2) "probably the trap" and (3) "bigger than it sounds". Both hold, and the
measurement makes the case sharper than either phrasing.

**(3) would not do what it claims.** Counting the inventory: about **16** checks are expressible as
a DataAnnotation (`[Required]` on a string, `[Range]`, one `[RegularExpression]`), over 7 of the 14
request schemas. About **31** are cross-field (`buyPrice <= sellPrice`, `buyer != seller`,
`quantity * price >= MinNotional`), stateful (does the wallet exist, is the quote active, is the
collateral locked, has the pending quote expired) or catalogue lookups (`IsValidCurrency`,
`GetTradingPairInfo`, Dealer mode). No attribute carries those. So restoring the attributes and
running them makes the schema describe roughly a **third** of enforcement — while changing the 400
body of all 24 writable endpoints, and leaving each one answering *two* error shapes:
`HttpValidationProblemDetails` from the filter and `ApiResponse<T>` from everything it cannot
replace. Full cost, partial benefit, document still not a contract.

Several of the checks are not attribute-expressible even in principle: `Guid.Empty` on a
non-nullable `Guid` — `publishedByUserId`, `adminUserId` — cannot fail `[Required]`. That is the
exact trap #236 named when it deleted `[Required]` from `OrderDto.Id` and `OrderDto.Side`.

**And the bot-side cost is silent.** The bot reads `message` out of the `ApiResponse<T>` envelope at
17 call sites — 9 in `WalletApiClient.cs`, 8 in `OrderApiClient.cs` — and shows it to the customer.
Deserializing an `HttpValidationProblemDetails` body into `ApiResponse<T>` **does not throw**: it
finds no `message`, leaves it null, and the `?? "خطا در …"` fallback fires. The bot would not break;
it would quietly stop telling customers *why* an order was refused, and no test would catch it.
(`UsersApiClient` differs — it logs and returns null on failure, so it degrades to a generic message
rather than a wrong one.)

So the honest name for that work is not "turn on validation" but **"unify the platform's error
contract"**. It belongs to its own issue, reconsidered only if #97 finds these 400s painful.

**(2) is worse than the issue suggests**, for the same arithmetic: a filter could mark only that
same third, so the document moves from *silent* to *incomplete but apparently complete*. #237 had
just finished removing an over-statement; this would add a new one in a different file.

## Testing Completed

Documentation only — no behavior change, so the bar is that nothing regressed.

```
scripts/check-doc-paths.sh
409 tracked text file(s) checked, every reference resolves.
```

```
```

```
dotnet test TallaEgg.sln
Passed! - Failed: 0, Passed: 894, Skipped: 0, Total: 894
```

### End-to-end
- [x] Not applicable — no behavior changed. The stack was run (`driver.ps1 start`) to *measure* the
      schemas and probe the endpoints, which is what produced the document; no code was changed
      after that.
- [x] Date/time run: 2026-09-08

### Environment

## One accidental write, disclosed

Probing `POST /api/user/update-phone` and `PUT /api/user/status` with `{}` was expected to answer
400 "user not found". It answered **200**: `TelegramId` defaults to `0` and the seeded root
administrator's TelegramId is `0`. That is finding 3 above, and it was found by causing it.

What actually changed on the local dev database, against the seed values in migration
`20250918224949_init`:

| field | seed | after the probe | now |
|---|---|---|---|
| `PhoneNumber` | `null` | `""` | `null` — restored |
| `Status` | `0` (Pending) | `0` | `0` — never differed |
| `LastActiveAt` | `null` | bumped | bumped; both endpoints do this by design |

Restored with `{"telegramId":0,"phoneNumber":null}` through the service's own endpoint and verified
on the wire (`"phoneNumber":null`). Nothing reads that field for this row — `GetUserIdByPhoneNumber`
returns null early on a blank string and `/api/userByPhone/` will not route an empty segment — so
the window was inert as well as brief. No production system was touched.

## Security Checklist

- [x] No names, emails, identifiers, personal paths or server addresses in the diff
- [x] Persian error messages are quoted verbatim, never translated

**One thing a reviewer should weigh.** The document states, publicly, that four endpoints accept an
empty body and act on it — including two that write to the administrator's row. This repository is
public. Judged worth writing: none of it is reachable without already being able to call the
service, all three APIs are loopback-only with no HTTPS binding, and Swagger is Development-only, so
this is not a remotely exploitable path. Leaving it undocumented would mean the next client meets it
by causing it, exactly as this PR did. Say the word if you would rather it were trimmed.

## Code Quality

- [x] Follows `STANDARDS.md`, including the section this PR adds
- [x] Written in English; Persian appears only as verbatim quotations of user-facing messages
- [x] No commented-out code
- [x] No dependencies added

## Documentation

- [x] `docs/design/API_REQUEST_VALIDATION.md` added
- [x] `docs/process/STANDARDS.md` §2 gains the rule and the pointer; its folder tree gains `design/`
- [x] `docs/process/INDEX.md` gains the folder and a navigation entry
- [x] Swagger/OpenAPI: unchanged — this PR describes the schemas, it does not alter them

## Breaking Changes?

- [x] No breaking changes
- [ ] Breaking changes

No code, no schema, no configuration. Three Markdown files.

## Deployment Notes

**Pre-deployment**: none. **Post-deployment**: none. **Rollback**: revert the commit.

No version bump — `STANDARDS.md` §11 reserves those for changes a caller could notice.

## Found, not fixed

Flagged per `STANDARDS.md` §3 rather than folded in. Both are described in the new document as
current behavior; neither is fixed here, and both are worth their own issue — say the word.

1. **Two paths answer 500 for a null `symbol`** — `POST /api/quotes/accept` and `POST /api/quotes`.
   `QuoteFillService.cs:70` splits a symbol it never null-checks. A one-line guard each, but it
   changes what those endpoints return, which is behavior and not this PR's scope.
2. **Four endpoints act on an empty body**, two of them writing to the administrator's row. The fix
   is a required-field check per endpoint, which is the very thing #242 decided not to do
   platform-wide — but these four are a narrower, arguably separate question, since the default
   value is actively destructive rather than merely unvalidated.

Out of scope as the issue requires: #240, #241, #243.

## Related Issues

- Closes #242
- Follows #235 / PR #236, which opened this gap deliberately, and #237 / PR #239, which deferred it again
- Relevant to #97 — the client that will meet these refusals

## Final Checklist

- [x] All tests pass locally
- [x] No secrets or sensitive data
- [x] Commit message is clear
- [x] Documentation updated

### DIFF
--- a/docs/design/API_REQUEST_VALIDATION.md
+++ b/docs/design/API_REQUEST_VALIDATION.md
@@ -1,0 +1,280 @@
+# Request validation: what the published schemas say, and what the servers do
+
+**Measured**: 2026-09-08, against `main` at `b8a2d14`, on the three running services.
+**Method**: walking `GET /swagger/v1/swagger.json` on Users (5136), Wallet (60933) and Orders
+(5140), and probing each writable endpoint over real HTTP — not read off the C#.
+
+This document exists because the published schemas declare **no constraints at all**, while the
+endpoints enforce roughly fifty. A client built from the document sends a body the document says
+is valid and gets refused. That is a deliberate state, recorded in issue #242 and summarized in
+[`../process/STANDARDS.md`](../process/STANDARDS.md) §2; this file is the detail that rule points
+at, because the list of what is actually enforced is the part a client cannot discover any other
+way.
+
+Read it as a snapshot. It will drift, and "How to re-measure" at the end says how to check.
+
+---
+
+## 1. What the documents declare
+
+```
+required   (schema-level, an array of property names)   Users 0   Wallet 0   Orders 0
+maxLength                                               Users 0   Wallet 0   Orders 0
+minimum                                                 Users 0   Wallet 0   Orders 0
+```
+
+Zero as well, everywhere: `minLength`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`,
+`pattern`, `minItems`, `maxItems`, `uniqueItems`, `multipleOf`.
+
+The only validation keyword present in any of the three documents is `enum` — 7 occurrences, every
+one generated from a C# enum type (`UserRole`, `UserStatus`, `OrderRole`, `OrderSide`,
+`OrderStatus`, `OrderType`, `TradingType`) rather than declared by anyone.
+
+`required: true` does appear 56 times, and it is not a counter-example: every occurrence is on a
+**parameter** object, never on a schema. Those are path parameters, which OpenAPI requires to be
+marked required. **No request-body property is required in any service.**
+
+Scale: 51 paths, 39 object schemas, 195 properties. 24 writable endpoints — 4 in Users, 6 in
+Wallet, 14 in Orders — taking 14 distinct request-body schemas between them, with 65 properties.
+
+## 2. What a schema-valid body actually gets
+
+There is no single answer. `{}` — valid against every request schema in all three services —
+produces **five** different outcomes:
+
+| outcome | endpoints |
+|---|---|
+| `400` with a Persian message | most of them |
+| `404` | `POST /api/user/update-role`, `POST /api/outbox/{id}/abandon` |
+| `200` with `success: false` | `POST /api/user/register` |
+| `200` having **acted on C# defaults** | 4 endpoints — see §4 |
+| `500` | 2 endpoints — see §3 |
+
+Measured, `{}` to every writable endpoint whose failure path provably writes nothing:
+
+| endpoint | status | body |
+|---|---|---|
+| `POST /api/user/register` | 200 | `{"success":false,"message":"کد دعوت معتبر نیست."}` |
+| `POST /api/user/update-phone` | 200 | succeeded — §4 |
+| `PUT  /api/user/status` | 200 | succeeded — §4 |
+| `POST /api/user/update-role` | 404 | `{"success":false,"message":"کاربر یافت نشد."}` |
+| `POST /api/wallet/deposit` | 400 | `{"success":false,"message":"کیف پول وجود ندارد","data":null}` |
+| `POST /api/wallet/withdrawal` | 400 | `{"success":false,"message":"کیف پول وجود ندارد","data":null}` |
+| `POST /api/wallet/lockBalance` | 400 | `{"success":false,"message":"کیف پول پیدا نشد","data":null}` |
+| `POST /api/wallet/unlockBalance` | 400 | `{"success":false,"message":"کیف پول پیدا نشد","data":null}` |
+| `POST /api/wallet/changeBalance` | 400 | `{"success":false,"message":"Invalid symbol ''. Expected BASE/QUOTE.","data":null}` |
+| `POST /api/orders` | 400 | `{"success":false,"message":"نماد معاملاتی الزامی است"}` |
+| `POST /api/quotes` | 400 | `{"success":false,"message":"خطا در انتشار مظنه.","data":null}` |
+| `POST /api/quotes/accept` | 400 | `{"success":false,"message":"مقدار باید بزرگ‌تر از صفر باشد.","data":null}` |
+| `POST /api/quotes/pending/{id}/approve` | 400 | `{"success":false,"message":"این مظنه پیدا نشد.","data":null}` |
+| `POST /api/quotes/pending/{id}/reject` | 400 | `{"success":false,"message":"این مظنه پیدا نشد.","data":null}` |
+| `POST /api/outbox/{id}/abandon` | 404 | `{"success":false,"message":"پیام یافت نشد.","data":null}` |
+
+The refusal body is the platform's `ApiResponse<T>` envelope — `{"success", "message", "data"}` —
+with `message` in Persian and meant to be shown to the customer. A client should read `message`,
+and should not assume a non-200 status: `register` refuses inside a 200.
+
+A body that is *almost* complete behaves the same way:
+
+```
+POST /api/orders   {"asset": "MAUA/IRT", "amount": 0.25}
+-> 400 {"success":false,"message":"قیمت برای سفارش محدود الزامی است"}
+```
+
+## 3. Schema-valid bodies that answer 500
+
+Two endpoints answer `500`, not `400`, for a body the schema permits. For a client this is worse
+than an unpredictable 400: the envelope carries a generic message with nothing to show a customer.
+
+```
+POST /api/quotes/accept   {"quantity": 1}   -> 500 {"success":false,"message":"خطای داخلی سرور"}
+POST /api/quotes          {}                -> 400, but through the generic catch, not the domain check
+```
+
+Both are a `NullReferenceException` on a null `symbol`.
+[`QuoteFillService.cs:70`](../../src/Order/Orders.Application/Services/QuoteFillService.cs#L70)
+does `symbol.Split('/')[0]` with no null check, and the quantity guard at
+[`:59`](../../src/Order/Orders.Application/Services/QuoteFillService.cs#L59) runs first — so a body
+that satisfies the quantity check reaches the split.
+
+`POST /api/wallet/unlockBalance` has the same shape for a different reason: a negative `amount`
+throws `ArgumentOutOfRangeException` at
+[`WalletRepository.cs:343`](../../src/Wallet/Wallet.Infrastructure/WalletRepository.cs#L343) and
+surfaces as 500. That one is already stated in the endpoint's own Swagger description.
+
+## 4. Schema-valid bodies that are accepted, having acted on C# defaults
+
+The sharpest consequence of nothing being required: **an omitted property is not an error, it is a
+value** — and on four endpoints that value is meaningful and destructive.
+
+| endpoint | property omitted | what gets written |
+|---|---|---|
+| `POST /api/user/update-phone` | `phoneNumber`, `telegramId` | `""` onto the user whose TelegramId is `0` |
+| `PUT /api/user/status` | `newStatus`, `telegramId` | `UserStatus.Pending` onto the user whose TelegramId is `0` |
+| `POST /api/autoquote-settings/{Base}/{Quote}/enabled` | `isEnabled` | `false` — auto-quoting turned **off** |
+| `POST /api/symbols/{Base}/{Quote}/active` | `isActive` | `false` — the symbol **disabled** |
+
+The first two are not hypothetical. `TelegramId` is a `long`, its default is `0`, and the seeded
+root administrator's TelegramId is `0` (migration `20250918224949_init`). An empty body addressed
+to the platform's SuperAdmin row is a successful write answered `200` — confirmed by accident while
+measuring this document, and reverted.
+
+The last two were derived from the code rather than fired, because firing them would have disabled
+a live trading symbol.
+
+## 5. Every hand-written check, by endpoint
+
+**shape** — expressible as a DataAnnotation on the DTO.
+**state** — needs a database read or a runtime catalogue lookup.
+**cross-field** — depends on another property of the same body.
+
+### Users.Api — 4 writable endpoints, and **not one field-shape check**
+
+Every refusal here is "this row does not exist". Nothing about the body itself is ever checked.
+
+| endpoint | check | kind | outcome |
+|---|---|---|---|
+| `POST /api/user/register` | invitation code resolves to an existing user ([`UserService.cs:33`](../../src/User/Users.Application/UserService.cs#L33)) | state | **200** `success:false` «کد دعوت معتبر نیست.» |
+| | a duplicate TelegramId is not checked; the unique index refuses it | — | 500 |
+| `POST /api/user/update-phone` | user exists by TelegramId ([`UserService.cs:79`](../../src/User/Users.Application/UserService.cs#L79)) | state | 400 «کاربر یافت نشد.» |
+| | phone format, non-emptiness, uniqueness — none | — | accepted |
+| `PUT /api/user/status` | user exists by TelegramId ([`UserService.cs:98`](../../src/User/Users.Application/UserService.cs#L98)) | state | 400 «کاربر یافت نشد.» |
+| | transition legality — none; any status may follow any other | — | accepted |
+| `POST /api/user/update-role` | user exists by UserId ([`Program.cs:462`](../../src/User/Users.Api/Program.cs#L462)) | state | **404** «کاربر یافت نشد.» |
+| | transition legality, and the caller's own role — none | — | accepted |
+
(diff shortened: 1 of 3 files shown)


MohKardan:
## Review response — all 13 findings verified, all 13 fixed (`c0ac61c`)

I checked every finding against the code before acting on it. **All thirteen were real**, including
three that changed what the document says rather than how it says it. One of them exposed a claim I
had derived instead of measured, and measuring it showed I was wrong.

### The three that changed the substance

**1. `reject` does not share `approve`'s checks (finding 1).** Confirmed —
`PendingQuote.Reject` (`PendingQuote.cs:196-205`) has one guard, `Status != Pending`. `Approve` has
three (`:177`, `:180`, `:183`). My "as above" was wrong, and the consequence is worse than a wording
slip: **`POST /api/quotes/pending/{id}/reject` with `{}` on a live id discards a proposed price and
records `Guid.Empty` as `ResolvedByUserId`**, with no expiry check either. §4 now lists five
endpoints, not four, and calls this one out as the one that looks safe and is not. My §2 probe
missed it because I used a nonexistent id and only ever saw the not-found branch — §9 now warns
about exactly that.

**2. The `POST /api/orders` `Split` — real, but I had the outcome wrong (finding 2).** The unguarded
`request.Asset.Split('/')[1]` at `OrderService.cs:82` is there, and `OrderSide.Buy = 0` is the
deserialized default, so a slashless asset takes the `[1]` branch. I wrote that it answers 500. It
does not — I measured it:

```
POST /api/orders   {"asset":"MAUA","amount":1,"price":1}
-> 400 {"success":false,"message":"خطا در ایجاد سفارش","data":null}
   run-logs/orders.out.log: System.IndexOutOfRangeException
```

The catch-all at `OrderService.cs:187` converts it. That is a more interesting fact than the one I
had assumed, because an opaque 400 is **indistinguishable from a business refusal** where a 500 at
least announces itself. §3 is rewritten around that split — two endpoints answer an opaque 400, one
answers 500 — and it now records the asymmetry that produces it: the catch-all excludes
`ArgumentException`, so the `Order` factories' throws still surface as 500 while everything else in
that method becomes a 400.

**3. The Admin bypass (finding 4).** Confirmed at `OrderService.cs:118-131`: `var isadmin =
user?.Role == UserRole.Admin;` then `if (!isadmin) if (!balanceCheckSuccess)` and `if (!isadmin) if
(!hasSufficientBalance)`. An Admin-role account places an order of any size with no balance and no
credit. `/api/quotes/accept` has no such bypass — checked, `QuoteFillService.cs:130-153` is
unconditional — so the difference between those two rows was real and my table erased it. Now stated
on the row.

### The rest

| # | Finding | Fix |
|---|---|---|
| 3 | `/api/quotes`'s NRE blamed on `QuoteFillService`, which it never calls | Re-cited to `QuoteRepository.cs:20` — confirmed `symbol.Trim()` with no null check |
| 5 | §2 said `{}` yields 500 on 2 endpoints; my own table showed otherwise | `{}` is now stated as the gentle case; the 500 needs a differently-shaped valid body, and §3 says which |
| 6 | "the envelope, message in Persian" false on both counts | New subsection: four endpoints return bare anonymous objects; `changeBalance` and the outbox endpoints return internal English. §8 now says not to pipe `message` to a customer unconditionally |
| 7 | `redrive` answers 404 not 400; `redrive-all-failed` enforces nothing | Both confirmed (`Program.cs:1168`, `:1189-1204`), split into their own rows; §6 is a new list of every 404 endpoint |
| 8 | `.ProducesValidationProblem(400)` missed | Confirmed — `Program.cs:687`, the only one in the repo. Now §1's own subsection, and §7 acknowledges that one endpoint already advertises `HttpValidationProblemDetails` while returning something else. You were right that this weakened the argument as written; it is stronger stated honestly, since a filter would spread that inconsistency to all 24 |
| 9 | `Order.Create` does not exist | `CreateMakerOrder` / `CreateLimitOrder`, both cited; the "unreachable today" claim is now qualified for `userId`, which nothing pre-checks |
| 10 | §4 not updated | `STANDARDS.md` §4 now lists what `docs/design/` holds. §2 placement kept — that is where the owner asked for it, next to the #235/#237 rules |
| 11 | STANDARDS duplicates the measurements | Trimmed to the rule plus the link, with an explicit note to keep counts in one file. This also removed the "two paths answer 500" figure that finding 5 had already made wrong |
| 12 | "catalogue" is British | Fixed in all three places; §1 requires American spelling |
| 13 | INDEX.md `Last Updated` stale | Moved to 2026-09-08 |

### One process note

Finding 2 is the useful one to learn from. Every other statement in that document was measured; that
one I reasoned out from the code and wrote in the same voice as the measured ones. It was wrong, in a
document whose entire value is that its claims are checkable. Every source citation has now been
re-resolved mechanically against the working tree, and anything not fired against a live service is
marked "from code" in §4 rather than blending in.

```
scripts/check-doc-paths.sh    409 tracked text file(s) checked, every reference resolves.
dotnet build TallaEgg.sln --no-incremental    0 Warning(s)  0 Error(s)
dotnet test TallaEgg.sln                      Failed: 0, Passed: 894, Skipped: 0
```

Still documentation only — no code, schema or configuration changed.

### Still not fixed, deliberately

The defects the document now records are described, not repaired — fixing them changes endpoint
behavior, which #242 explicitly is not. The list has grown with what this review turned up, and each
is worth its own issue:

1. **`reject` accepts an empty body and discards a live pending quote**, recording `Guid.Empty` as
   the resolver. The most serious of the set.
2. **Three unguarded `Split`/`Trim` calls** on `symbol`/`asset` — one 500, two opaque 400s.
3. **Four endpoints act on destructive C# defaults** (`isEnabled`, `isActive`, plus the two Users
   endpoints that write to the administrator's row).
4. **`.ProducesValidationProblem(400)` on `POST /api/orders`** advertises a body the endpoint never
   returns — the surviving half of what #237/#239 set out to clean up, and a one-line deletion.
5. **The Admin balance bypass** — behavior, and possibly intended; worth confirming rather than
   assuming.

Say the word and I will open them.

PR #188, opened 2026-09-01
[Feat] Simulator trades every symbol in the catalogue, at each one's own precision — issue #147

MohKardan:
**Branch Name**: `feat/simulator-trades-every-symbol`
**Linked Issue**: Closes #147

---

## Description

The simulator hard-coded one symbol, so a full run — 100 users, 120 quotes, 1000 trades — was
entirely gold. It now reads its symbols from `CurrenciesConstant.AllTradingPairs` and gives each
one its own quotes, its own wallet funding and its own trade sizes, at that symbol's own
precision.

That precision is the point. On 2026-08-27 a run reported `trades attempted 1000, errors 0` with a
verifiably clean database, and **a single manual BTC trade immediately afterwards found #146** —
`Orders.Amount` was `decimal(18, 2)` while wallets and trades were `(28, 8)`, so `2.111` was stored
as `2.11` and the collateral difference stayed locked with no order holding it. The simulator could
not have found it: MAUA's precision is 2, exactly what the column held, so every gold quantity
round-tripped unchanged. A run that touches one symbol proves that symbol works and reads as if it
proved the platform works.

---

## Type of Change
- [x] Feature (new capability) — dev tooling only; no product code is touched.

---

## Changes Made

- **`SymbolPlan.cs` (new)** — one trading pair as a run will trade it. Derives the trade-size band
  from that pair's own `MinQuantity`, `MinNotional` and `BaseDecimalPlaces` (largest trade =
  30× the smallest tradable size, clamped to `MaxQuantity`), and resolves the pair's quote-command
  keyword and `CREDIT_` ledger.
- **Phase 2** turns auto-quote off for every symbol the run trades, not just `MAUA/IRT`.
- **Phase 4** funds each quote currency, each traded symbol's base asset, and each base asset's
  `CREDIT_` ledger — sized from the same band Phase 7 draws from, so no symbol's reserve runs out
  earlier than another's. It previously deposited Toman and MAUA and nothing else, which is why
  registration's three default wallets were all anyone ever had.
- **Phase 6** publishes per symbol through the real `buyPrice-sellPrice` admin command, using each
  pair's own alias (`سکه`, `بیت`; none for `MAUA/IRT`, which the command already defaults to). The
  mid **anchors on the symbol's currently published quote** and walks by a fraction rather than a
  fixed toman step — an absolute step that varies gold plausibly is lost in the rounding of a
  Bitcoin price, and a quote more than ±5% from the current one is *held for approval*, not
[...]

Issue #39, opened 2026-07-25
Reconciliation & re-drive for permanently-failed outbox settlements (stuck trades / locked balances)


MohKardan:
## Summary
When an outbox `TradeSettlement` message exhausts its retries, it is marked permanently failed and abandoned — but the underlying trade is already recorded and the participants' balances remain locked and unsettled, with no path to recover. There is currently no reconciliation or operational tooling to find and resolve these stuck settlements.

## Evidence (observed live during testing)
```
[ORDERS] [ERR] Outbox message "b4b6489d-..." (aggregate "13b54c08-...") permanently failed after 5 attempts.
[ORDERS] [ERR] Outbox message "65f5d9fa-..." (aggregate "889058a1-...") permanently failed after 5 attempts.
```
After these, the matched trades exist in `Trades`, the buyers/sellers still have the collateral in `LockedBalance`, and no wallet transactions were written. The state is inconsistent (trade recorded, wallets not settled) and nothing surfaces or fixes it.

## Why it matters
The whole point of the outbox is that a matched trade is *never left unsettled*. Retry handles transient failures, but a **permanent** failure (a real bug, like the fee/credit bugs found during this test) currently ends in a silent dead-end: money is stuck locked and the two ledgers (Orders `Trades` vs Wallet `Transactions`) disagree. For real money this needs a human-visible, recoverable path.

## Proposed scope
1. **Reconciliation report/job**: periodically (or on demand) compare `Trades` against wallet `Transactions` by trade id and list trades with no corresponding settlement (including permanently-failed outbox messages).
2. **Re-drive path**: an operator action to reset a `Failed` outbox message back to `Pending` (safe because settlement is idempotent on the trade id) so it retries after the root cause is fixed.
3. **Alerting**: raise a clear, visible signal when a message goes permanently failed (not just an ERR log line).
4. Decide the policy for a genuinely unsettleable trade (e.g. compensating action / manual offline settlement, matching the gold-shop model).

## Acceptance criteria
- Given a trade whose outbox message permanently failed, an operator can list it and re-drive settlement after fixing the cause, and it settles exactly once (idempotent).
- A permanent failure produces a visible alert, not only a log line.

## Immediate cleanup (this test session)
[...]

PR #259, opened 2026-09-08
[Fix] Simulator reset deletes both sides of a simulated fill, not just the simulated one — issue #257

### LINKED ISSUE #257: Simulator reset deletes the trade and the customer's order, but leaves the dealer's — 4,451 of 4,503 local orders are orphans
The simulator's reset deletes a fill's trade and the customer's order, but not the dealer's order
on the other side. The dealer is a real account, so it is never in the id list the delete is scoped
to. **Every simulated fill therefore leaves one orphan order behind, permanently.**

Measured on a dev machine on 2026-09-08, on `main` at `496ca88`.

## What the reset does

[`DataReset.cs`](../blob/main/TelegramBot/TallaEgg.TelegramBot.Simulator/DataReset.cs) collects the
simulated user ids — `SELECT Id FROM Users WHERE TelegramId < 0` — and then, in the Orders
database:

```sql
DELETE FROM Trades
WHERE BuyerUserId IN (simulated) OR SellerUserId IN (simulated)   -- either side is enough

DELETE FROM Orders
WHERE UserId IN (simulated)                                       -- only the simulated side
```

The two predicates are not the same shape, and that is the whole defect. A quote fill creates two
orders — the customer's and the quote publisher's (`QuoteFillService.CreateSideAsync`, called
twice) — and the publisher is the market maker: a **real** account with a positive `TelegramId`,
deliberately outside the range the reset is scoped to.

So the trade is deleted because one of its sides qualifies, the customer's order is deleted, and
the dealer's order survives with nothing left pointing at it. It is `Completed`, its
`RemainingAmount` is `0`, and the trade that completed it no longer exists.

## What it has accumulated here

```sql
SELECT COUNT(*) FROM Orders;                     -- 4503
SELECT COUNT(*) FROM Trades;                     --   26

-- orders with no trade on either side
SELECT COUNT(*) FROM Orders o WHERE NOT EXISTS (
  SELECT 1 FROM Trades t WHERE t.BuyOrderId = o.Id OR t.SellOrderId = o.Id);
-- 4451
```

**4,451 of 4,503 orders have no trade at all.** Split by the note each side is created with:

```
dealer side   (طرف مقابل مظنه …)   4468
customer side (پذیرش مظنه …)         33
other                                 2
```

4,468 dealer-side orders against 33 customer-side ones, from a service that creates them strictly
in pairs. 4,460 of them sit on a single account — `TelegramId 6389449308`, `Role = Admin`, a real
developer account, exactly the one `DataReset`'s own class comment describes as having been deleted
once by the older `TelegramId >= 900000000` filter.

Growth is per simulator run, not per day of use:

```
2026-08-26      2
2026-08-27     10
2026-08-31   1067
2026-09-01   3381
```

The largest single block is 3,389 orders under one quote id, one user, over about nineteen hours,
with zero trades referencing any of them.

## Why it matters

It is **dev-only** — production never runs the simulator — and no customer is affected. What it
costs:

1. **It grows without bound.** Nothing ever removes these rows; each run adds one per fill.
2. **It corrupts the admin's order history.** `GET /api/orders/user/{id}` for that account returns
   thousands of orders that never happened, and the bot's own history screens read from it.
3. **It makes the Orders table useless as evidence, silently.** This is not hypothetical: issue
   #250 opens with "4,523 orders in the database, every one of them says `Market`", and reasoned
   from that figure about what the product does. Almost every one of those rows is this residue.
   The issue's central measurement was of the simulator's leftovers, not of the product.

(3) is the reason this is worth fixing rather than tolerating. A dev database that quietly answers
questions wrong costs more than one that is merely dirty.

## Suggested fix

Delete the counterparty side **through the trade**, before deleting the trades that identify it —
not by `UserId`, which cannot see the other half of a pair:

```sql
-- both sides of every trade a simulated user was part of, while the link still exists
DELETE FROM Orders WHERE Id IN (
  SELECT BuyOrderId FROM Trades WHERE BuyerUserId IN (…) OR SellerUserId IN (…)
  UNION
  SELECT SellOrderId FROM Trades WHERE BuyerUserId IN (…) OR SellerUserId IN (…));

DELETE FROM Trades WHERE …;          -- then the trades
DELETE FROM Orders WHERE UserId IN (…);  -- then anything left that never matched
```

Ordering matters and is the opposite of the current one: the `Trades` rows are what identify the
counterparty orders, so they have to be read before they are destroyed.

Two things to decide, neither of which should be guessed:

- **The 4,460 rows already there.** They can be identified the same way — a `Completed` order with
  `RemainingAmount = 0` and no trade on either side is not a state the product can produce. Whether
  to clean them up, and on which databases, is the owner's call.
- **Whether the market maker's own orders should be deletable at all.** The reset is scoped to
  `TelegramId < 0` precisely so a real account is never touched, and that guard exists because an
  earlier filter deleted a real developer's account, wallets and history. Deleting the dealer's
  order because a simulated user was on the other side is a deliberate narrowing of that guard, and
  it should be written down as one.

## What is not wrong here

- **The `TelegramId < 0` scoping itself is right** and should not be widened. Its class comment
  records why.
- **The outbox drain is right** and is not involved. `DataReset` waits for `OutboxMessages` to have
  no `Pending` row before deleting anything (#184, #175); that is a different concern and works.
- **The wallet residue is already known** and is a separate consequence of the same scoping — the
  market maker's `LockedBalance` accumulates across runs. That is not this.

## Related

- #250 / PR #256 — where this was found; its headline row count is this residue
- #185 (#184, #175) — the outbox drain in the same class
- #147 — the run that trades every symbol, which is what made the volume large

MohKardan:
## Description

`DataReset` deleted trades **by participant** but orders **by owner**, and those are different
shapes. A quote fill creates two orders — the customer's and the market maker's — and the market
maker is a real account by design, so `UserId IN (simulated)` could never see their half. The
trade went, the customer's order went, and the dealer's order stayed behind: `Completed`,
`RemainingAmount = 0`, with nothing pointing at it. That is a state the product cannot otherwise
produce.

## Issue/Task Reference

Closes #257

## Two corrections to the issue

Both were found by checking its claims, and both changed the work.

**1. "Every simulated fill leaves an orphan" does not hold on current `main`.** The simulator
promotes `users[0]` to market maker — a *simulated* account (`TelegramIdBase = -1_000_000_000`) —
so both sides of a fill today are simulated and both orders are deleted correctly. Measured on the
run already in the database when this started: 6 trades, both sides simulated in every one, and
what the unfixed reset would have orphaned from them was **0**.

```
B|S|count|first                |last
0|0|   16|2026-08-26 08:19:43. |2026-08-31 19:42:37.     <- real/real, untouched
1|1|    6|2026-09-08 17:52:20. |2026-09-08 17:52:21.     <- both sides simulated
```

The residue accumulated between 2026-08-27 and 2026-09-01 15:10, when `AutoQuotePublisherService`
kept replacing a run's quote with one published by a *real* account — the condition the comment in
`Simulation.cs` describes. PR #188 turned auto-quote off per traded symbol on 2026-09-01 for an
unrelated reason and incidentally closed that path; no orphan has been created since. So this is a
**latent** defect rather than an actively growing one. It still fires whenever a non-simulated
account is the counterparty, and the reset is what every other measurement of this database rests
on, so it is worth fixing.

**2. The issue's suggested SQL cannot run.** It proposes deleting the orders before the trades.
`Trades` holds four foreign keys into `Orders` — `BuyOrderId`, `SellOrderId`, `MakerOrderId`,
`TakerOrderId` — all `NO_ACTION`. Proved in a rolled-back transaction:

```
DELETE FAILED: The DELETE statement conflicted with the REFERENCE constraint
"FK_Trades_Orders_BuyOrderId". The conflict occurred in database "TallaEggOrders",
table "dbo.Trades", column 'BuyOrderId'.
```

The principle is right — read the link before destroying it — but the ids have to be **captured**
first, then the trades deleted, then the orders.

## Changes Made

- `DataReset.DeleteOrderDataAsync` now captures both sides of every trade a simulated user was part
  of, deletes those trades, then deletes the captured orders, then anything still owned by a
  simulated user.
- The captured delete is guarded by `NOT EXISTS` against surviving trades, so an order filled
  against both a simulated and a real user is left alone.
- Written as **one batch with table variables**, not four commands and a `#temp` table: a
  parameterised `SqlCommand` goes through `sp_executesql`, so a `#temp` table created by one
  command is gone before the next runs. This was found the hard way — see below.
- The SQL is an `internal const` so the suite can assert the statement order.
- Class comment records the narrowing of the `TelegramId < 0` guard and why it does not reopen the
  incident that guard exists to prevent.
- `scripts/delete-simulator-orphan-orders.sql` for the residue already present.

### The failure, then the fix

Reproduced against real SQL Server on throwaway databases carrying the same four foreign keys, so
the FK behaviour is genuine. One simulated fill (simulated customer, real dealer) plus a control
fill between two real accounts.

**Before — unmodified `main`:**

```
=== BEFORE RESET ===
orders=4 trades=2 orphans=0

Reset: deleted 1 Trades rows.
Reset: deleted 1 Orders rows.

=== AFTER RESET ===
orders=3 trades=1 orphans=1
  0B000001 owner=3333...  real customer side
  0A000002 owner=2222...  sim dealer side      <- ORPHANED
  0B000002 owner=2222...  real dealer side
```

**After — this branch:**

```
=== BEFORE RESET ===
orders=4 trades=2 orphans=0

Reset: deleted 1 Trades rows.
Reset: deleted 2 counterparty Orders rows.
Reset: deleted 0 Orders rows.

=== AFTER RESET ===
orders=2 trades=1 orphans=0
  0B000001 owner=3333...  real customer side
  0B000002 owner=2222...  real dealer side
```

Both sides of the simulated fill are gone; the real/real fill is untouched.

### The guard is load-bearing

An order on a real account filled twice — once against a simulated user, once against a real one:

```
=== BEFORE ===
orders=3 trades=2 orphans=0

Reset: deleted 1 Trades rows.
Reset: deleted 1 counterparty Orders rows.

=== AFTER ===
orders=2 trades=1 orphans=0
  0D000001 owner=2222...  real resting order   <- survives, real trade still points at it
  0D000003 owner=3333...  real taker
```

With the `NOT EXISTS` removed, the same case does not over-delete — it throws:

```
Unhandled exception. Microsoft.Data.SqlClient.SqlException: The DELETE statement conflicted
with the REFERENCE constraint "FK_Trades_Orders_SellOrderId".
```

### The `#temp` table trap

The first implementation used four commands and a `#temp` table. It built, and failed at runtime:

```
Error Number:208 — Invalid object name '#CounterpartyOrders'.
```

A parameterised `SqlCommand` is sent through `sp_executesql`, which scopes the temp table to that
call. Hence the single batch.

### Unit tests

Four new tests (7 cases) in `SimulatorResetOrphanTests`. Each was mutation-checked — every mutation
fails exactly one test and nothing else:

| Mutation | Result |
|---|---|
| capture moved after the trades delete | `AreCapturedBeforeTheTradesAreDeleted` red, 6 pass |
| `NOT EXISTS` removed | `SkipsOrdersASurvivingTradeStillReferences` red, 6 pass |
| maker/taker dropped from the capture | `MatchesEveryOrderForeignKey(MakerOrderId)` red, 6 pass |

```
    0 Warning(s)  0 Error(s)

dotnet test TallaEgg.sln
Passed!  - Failed: 0, Passed: 918, Skipped: 0, Total: 918
```

### End-to-end

`driver.ps1 start` then `driver.ps1 smoke`, 2026-09-08:

```
23:03:39 Reset: found 5 previously simulated users.
23:03:39 Reset: deleted 6 Trades rows.
23:03:39 Reset: deleted 12 counterparty Orders rows.
23:03:39 Reset: deleted 0 Orders rows.
23:03:50 === Done in 00:00:11.6. Registered 5 (4 approved), trades attempted 10, errors 0 ===
Simulation completed with errors 0; 10 settlement(s) completed, no new failures.
```

The by-owner delete now finds nothing left, which is correct for a run whose every fill was
simulated on both sides — it remains the backstop for orders that never reached a trade.

Database afterwards:

```
orphanCompleted=0   tradesMissingAnOrder=0   outboxPending=0   duplicateSettlements=0
orphanTransactions=0
```

Two pre-existing items, neither touched by this change and both already known: 101 `Failed` outbox
messages, all dated 2026-09-01, and the market maker's accumulated `LockedBalance` — which #257
explicitly puts out of scope.

### Cleanup script

Verified on throwaway data first: of six orders — one orphan, one `Cancelled`, one `Failed`, two
traded, one referenced only as `MakerOrderId` — it deleted exactly the orphan and left the other
five, including the maker-only reference. Idempotent on a second run (`Deleted: 0`).

Then run against the local dev database:

```
=== Orders database: TallaEggOrders ===
  Orders before:            4496
  Completed, no trade:      4449
  Deleted:                  4449
  Orders after:             47
  Trades (unchanged):       22
```

47 orders and 22 trades: 44 sides pairing to 22 trades, plus the 3 trade-less `Cancelled`/`Failed`
orders the script deliberately spares. The 4,449 deleted rows were exported to a TSV outside the
repo before the run.

(The issue's headline figures were 4,451 of 4,503; re-measuring at the time of the fix gave 4,449
of 4,496, the difference being ordinary activity in between.)

## Found and not fixed

Both worth their own issue — reported, not fixed, per scope discipline:

1. **`driver.ps1` restores auto-quote with `UpdatedByUserId = Guid.Empty`.** `AutoQuoteSettings` for
   `BTC/IRT` and `MAUA/IRT` currently sit `IsEnabled = 1` with an empty updater, and
   `AutoQuotePublisherService` publishes as `settings.UpdatedByUserId` — so an auto-published quote
   would be attributed to a user that does not exist. That is precisely the shape of counterparty
   that created this issue's residue in the first place.
2. **Seven customer-side orphans on real dev accounts** (2026-08-27 and 2026-08-31) are residue of
   the older `TelegramId >= 900000000` filter incident that `DataReset`'s class comment describes,
   not of this defect. They were removed by the cleanup script along with the rest.

## Security Checklist

- [x] `.gitignore` blocks sensitive files — confirmed `config/appsettings.global.json` stays ignored in the worktree used here
- [x] SQL injection protections — the id list stays a parameter; the new statements add no concatenation

## Deployment Notes

**None.** The simulator is a developer tool and is not deployed; production never runs it and has
none of this residue. The cleanup script is run by hand against a developer database and is not
wired into anything.

**Rollback**: revert the commit. The reset returns to its previous behaviour, which leaves orphans
but breaks nothing.

## Breaking Changes?

- [x] No breaking changes

### DIFF
--- a/TelegramBot/TallaEgg.TelegramBot.Simulator/DataReset.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Simulator/DataReset.cs
@@ -152,6 +152,15 @@ public sealed class DataReset(string usersDbConnectionString, string walletDbCon
     /// </para>
     ///
     /// <para>
+    /// <b>One transaction</b>, because the capture is only valid until the trades are deleted.
+    /// If the counterparty delete failed on its own — a deadlock against the matching sweep,
+    /// say — the trades would already be gone and the orders they identified would be
+    /// unrecoverable, creating exactly the permanent orphans this fix exists to remove. The
+    /// delete-by-owner it replaced had no such window, because <c>UserId</c> stays true however
+    /// far the batch got.
+    /// </para>
+    ///
+    /// <para>
     /// Maker and taker are matched alongside buy and sell because each of the four is its own
     /// foreign key, so any of them can hold an order in place.
     /// </para>
@@ -164,6 +173,8 @@ public sealed class DataReset(string usersDbConnectionString, string walletDbCon
     /// </summary>
     internal const string OrderDataDeleteSql =
         """
+        SET XACT_ABORT ON;
+
         DECLARE @Sim TABLE (Id uniqueidentifier PRIMARY KEY);
         INSERT INTO @Sim (Id)
         SELECT DISTINCT CAST(value AS uniqueidentifier) FROM STRING_SPLIT(@Ids, ',');
@@ -178,13 +189,20 @@ public sealed class DataReset(string usersDbConnectionString, string walletDbCon
         WHERE t.BuyerUserId IN (SELECT Id FROM @Sim)
            OR t.SellerUserId IN (SELECT Id FROM @Sim);
 
+        DECLARE @DeletedOwners TABLE (UserId uniqueidentifier);
+
+        BEGIN TRANSACTION;
+
         DELETE FROM Trades
         WHERE BuyerUserId IN (SELECT Id FROM @Sim)
            OR SellerUserId IN (SELECT Id FROM @Sim);
         DECLARE @TradesDeleted int = @@ROWCOUNT;
 
         DELETE FROM Orders
+        OUTPUT deleted.UserId INTO @DeletedOwners
         WHERE Id IN (SELECT Id FROM @Counterparty)
+          AND Status = 'Completed'
+          AND RemainingAmount = 0
           AND NOT EXISTS (
               SELECT 1 FROM Trades t
               WHERE t.BuyOrderId = Orders.Id OR t.SellOrderId = Orders.Id
@@ -194,7 +212,12 @@ public sealed class DataReset(string usersDbConnectionString, string walletDbCon
         DELETE FROM Orders WHERE UserId IN (SELECT Id FROM @Sim);
         DECLARE @OrdersDeleted int = @@ROWCOUNT;
 
-        SELECT @TradesDeleted, @CounterpartyDeleted, @OrdersDeleted;
+        COMMIT TRANSACTION;
+
+        DECLARE @RealAccountsDeleted int =
+            (SELECT COUNT(*) FROM @DeletedOwners WHERE UserId NOT IN (SELECT Id FROM @Sim));
+
+        SELECT @TradesDeleted, @CounterpartyDeleted, @OrdersDeleted, @RealAccountsDeleted;
         """;
 
     /// <summary>
@@ -221,14 +244,17 @@ public sealed class DataReset(string usersDbConnectionString, string walletDbCon
     ///
     /// <para>
     /// This is a deliberate narrowing of the <c>TelegramId &lt; 0</c> guard described on the
-    /// class, and the only one: an order belonging to a real account is deleted solely because a
-    /// simulated user was the counterparty of a trade it was part of. It does not reopen the
-    /// incident that guard exists to prevent — that was a filter which selected real
-    /// <em>users</em> wholesale, deleting an account, its wallets and its history. Nothing here
-    /// widens which users are selected, and the <c>NOT EXISTS</c> keeps any order still
-    /// referenced by a surviving trade, so a real order filled against both a simulated and a
-    /// real user keeps the history that is genuinely its own. Without that clause the delete
-    /// does not merely over-reach, it fails outright on FK_Trades_Orders_SellOrderId.
+    /// class, and it is kept as narrow as the defect: an order on a real account is deleted only
+    /// when a simulated user was the counterparty of a trade it was part of, <b>and</b> that
+    /// trade consumed it entirely. Two clauses hold that line. <c>NOT EXISTS</c> keeps any order
+    /// a surviving trade still references, so one filled against both a simulated and a real
+    /// user stays — without it the delete does not merely over-reach, it fails outright on
+    /// FK_Trades_Orders_SellOrderId. <c>Completed</c> with <c>RemainingAmount = 0</c> keeps a
+    /// real user's resting order that a simulated user only partially filled: that order existed
+    /// before the run and is not the run's to delete, and removing it would strand its
+    /// <c>LockedBalance</c> in a real wallet with no order left to reconcile against. What is
+    /// deleted is only an order that exists solely because of a trade being deleted — which is
+    /// what the dealer's side of a quote fill is.
     /// </para>
     /// </summary>
     private async Task DeleteOrderDataAsync(List<Guid> userIds, CancellationToken ct)
@@ -248,8 +274,16 @@ public sealed class DataReset(string usersDbConnectionString, string walletDbCon
         await reader.ReadAsync(ct);
 
         logger.LogInformation("Reset: deleted {Count} Trades rows.", reader.GetInt32(0));
-        logger.LogInformation("Reset: deleted {Count} counterparty Orders rows.", reader.GetInt32(1));
-        logger.LogInformation("Reset: deleted {Count} Orders rows.", reader.GetInt32(2));
+
+        // Both sides of a deleted trade come out through the capture, so this count mixes the
+        // simulated user's own order with the counterparty's. The second number is the one that
+        // matters when reading a run: it is how far outside TelegramId < 0 this reset reached.
+        logger.LogInformation(
+            "Reset: deleted {Count} Orders rows through those trades, {RealCount} of them on a real account.",
+            reader.GetInt32(1), reader.GetInt32(3));
+
+        logger.LogInformation("Reset: deleted {Count} further Orders rows owned by a simulated user.",
+            reader.GetInt32(2));
     }
 
     private Task WaitForOutboxToDrainAsync(CancellationToken ct) =>
--- a/TelegramBot/TallaEgg.TelegramBot.Simulator/DataReset.cs
+++ b/TelegramBot/TallaEgg.TelegramBot.Simulator/DataReset.cs
@@ -4,8 +4,12 @@ using Microsoft.Extensions.Logging;
 namespace TallaEgg.TelegramBot.Simulator;
 
 /// <summary>
-/// Wipes whatever a previous simulation run left behind, scoped strictly to
-/// <c>TelegramId &lt; 0</c> so a real user's data can never be touched. Runs first, every
+/// Wipes whatever a previous simulation run left behind, scoped to users with
+/// <c>TelegramId &lt; 0</c> so a real user's account, wallets and history can never be
+/// touched. The one thing outside that scope which is deleted is the order sitting on the
+/// other side of a simulated fill, which belongs to the market maker — a real account by
+/// design; see <see cref="DeleteOrderDataAsync"/> for why that is required and why it does
+/// not widen the guard. Runs first, every
 /// time: the whole point of a clean slate is that every run exercises
 /// registration-through-settlement from scratch rather than building on state a previous run
 /// happened to leave around.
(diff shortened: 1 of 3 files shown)


MohKardan:
## Self-review (`/code-review high 259`) — five findings, four fixed in bc0151c

**1. Counterparty delete had no `Status`/`RemainingAmount` clause — fixed.** The real one. A real
user's resting order that a simulated user only *partially* filled was deleted along with the
trade, stranding its `LockedBalance` in a real wallet with no order row left to reconcile against.
`OrderStatus.Partially` is a real state and `ExecuteAtomicMatchAsync` clamps with `Math.Min`, so the
case is reachable. The delete is now restricted to orders a deleted trade consumed entirely, which
is exactly what a quote fill's dealer side is — an order that exists *solely* because of the trade
being deleted.

Proved both ways, on a real user's order left `Partially` filled with 4 of 10 units open:

```
### without the clause
    Reset: deleted 2 Orders rows through those trades, 1 of them on a real account.
orders=0 trades=0                          <- the real user's resting order is gone

### with it
    Reset: deleted 1 Orders rows through those trades, 0 of them on a real account.
orders=1 trades=0
  survives: real resting, part-filled [Partially]
```

**2. No transaction around the batch — fixed.** A failure after the trades delete but before the
counterparty delete would have left the captured ids unrecoverable, creating permanently the very
orphans this PR removes. The delete-by-owner it replaced had no such window, because `UserId` stays
true however far the batch got. Now `SET XACT_ABORT ON` with `BEGIN`/`COMMIT TRANSACTION` around
all three deletes, and a test pins that the trades delete and the counterparty delete are both
inside it.

**3. `NOT EXISTS` test sliced to the end of the whole string — fixed.** Correct: moving the guard
onto a later `DELETE` would have kept the test green. Both statement-scoped assertions now go
through one `StatementAt` helper that slices to the statement's own terminating semicolon.

**4. Log conflated the two sides — fixed.** The capture takes both sides of every deleted trade, so
"counterparty Orders rows" read 12/0 where the real split was 6 and 6. The delete now reports the
number that actually matters when reading a run — how far outside `TelegramId < 0` it reached:

```
Reset: deleted 2 Orders rows through those trades, 1 of them on a real account.
Reset: deleted 0 further Orders rows owned by a simulated user.
```

**5. Cleanup script destroys rows `OrderCollateralReconciler` needs — does not stand.** The
reconciler is reached only from `OutboxProcessorService`, per settled trade
(`ReleaseResidualLockIfCompletedAsync(trade.BuyOrderId/SellOrderId)`), never as a sweep over
historical orders — the retrospective reconciliation its own comment mentions is #39, which does not
exist yet. So nothing would ever have revisited those rows. And they could not have sized a lock in
any case: `ComputeResidualLockAsync` is `locked − sum(this order's trades)`, and their trades were
deleted long ago, so it would have returned the entire lock rather than the residue. The 4,449 rows
were exported to a TSV before deletion, so the forensic trail for the market maker's accumulated
`LockedBalance` (#124) is intact.

## Re-verified after the changes

```
dotnet build TallaEgg.sln --no-incremental     0 Warning(s)  0 Error(s)
dotnet test  TallaEgg.sln                      Passed! - Failed: 0, Passed: 920, Total: 920
```

All three reproductions still behave, and `driver.ps1 smoke` again:

```
Reset: deleted 10 Trades rows.
Reset: deleted 20 Orders rows through those trades, 0 of them on a real account.
Reset: deleted 0 further Orders rows owned by a simulated user.
=== Done in 00:00:13.3. Registered 5 (4 approved), trades attempted 10, errors 0 ===
Simulation completed with errors 0; 10 settlement(s) completed, no new failures.

orphanCompleted=0  tradesMissingAnOrder=0  outboxPending=0  failedNotFrom0901=0
```

